Issue #88, opened 2021-08-13
The game asciisector crashes during gameplay with a missing syscall error


nniro:
The game is asciisector version 7.2 from the site : asciisector.net

The game runs correctly until we either :

   * Open the Exchange terminal and close it
   * Try to land on a planet or a station

At that point, the game crashes with this message :

Error: Unsupported Syscall 0x52h (82)

PR #89, opened 2021-08-13
Added syscall 82 (for #88)

nniro:


ptitSeb (commented):
You should also create the direct wrap line 119, for the build where __NR_rename does exist

ptitSeb:
Thanks, nice!

nniro:
Yeah, I added the part on line 119 at first but then I wasn't sure because __NR_rename would give out an error at compile time (so it really needs the macro conditional).

nniro:
Thank you for writing such a nice project! 

Merci beaucoup, c'est super de pouvoir faire rouler des jeux sur ARM! :D

ptitSeb:
De rien :)

(asciisec fait partie de ma "liste de jeux à tester regulièrement", mais je ne joue pas. je fait juste un nouveau jeux et je quite :( )

PR #136, opened 2021-10-17
Add Snapdragon 845 compilation

Heasterian:


mogery:
This is broken. Line 78 of CMakeLists should start with elseif, not if.

PR #166, opened 2021-11-19
Minor changes to allow box64 to build and run on 64-bit PowerPC.

eerykitty:
ppc64le/ELFv2 appears to use the same varargs type as LoongArch64. I changed the loongarch varargs macros to apply to powerpc64 as well, and it builds and appears to work correctly.

The only other change is a small difference in structs between 32 and 64 bit powerpc uc_mcontext.


ptitSeb:
Nice! I guess a profile in the CMakeList.txt would also be a good thing.

PR #176, opened 2021-11-22
Improve README

ryanfortner:


docs/README.md
```diff
@@ -1,20 +1,33 @@
...
+<p align="center">Linux Userspace x86_64 Emulator with a twist
+<p align="center">
+  <a href="https://github.com/ptitSeb/box64/blob/main/docs/CHANGELOG.md">
+    View changelog</a>
+  |
+  <a href="https://github.com/ptitSeb/box64/blob/main/docs/README_CN.md">
+    中文</a>
+  |
+  <a href="https://github.com/ptitSeb/box64/issues/new">
+    Report an error</a>
```
ptitSeb, line 12: I don't like having html code inside the makdown file. I prefer to keep markdown syntax only.

ptitSeb (changes requested):

docs/README.md
ryanfortner: Hmm... is there any way to center text in markdown without using html?

Itai-Nelken: markdown doesn't support aligning. github markdown supports with HTML.

ptitSeb: Yes, but I don't like poluting markdown file with html. Please don't try to align.

ryanfortner:
@ptitSeb is this better?

ptitSeb (approved):
Thanks :)

PR #199, opened 2021-12-30
Add initial RISCV support

Alexmitter:
To begin this, I have to admit that I have not successfully ran any binary, but that matches my experience on the pinebook pro, the device I use to cross check if what if what my RISCV machine does is matching aarch64. That's why I WIP'd it.

I checked something as simple as GNU Hello and it failed on both aarch64 and riscv in the same way.

Please advise on which binaries are good to test box64 on those machines :)


ptitSeb:
Start with the included tests. The test01 is the easiest.
First thing to do is check how the vaarg are handled.

ptitSeb:
I didn't understood what you said about the aarch64 test? You used a RiscV build on a Aarch64?

Alexmitter:
I have a RISCV machine and a Pinebook Pro with a freshly compiled version following the instructions. What I did to test is to compare what both do. Not very smart, I know. I completely missed the tests folder, now executing on my RISCV machine.

I ran through all of them and think we do have a PASS here :)
Here is a log from test01 to test18: https://gist.github.com/Alexmitter/1999fc8ecc438476f54a3c5119452180

Alexmitter:
More info about my machine, its a Hifive Unmatched. There is not much riscv hardware around yet, so its really the only thing I have and can test this on, but it seems to work just fine :)

Alexmitter:
I just noticed that half of the log was missing in the previous gist, here a full log of all 18 tests: https://gist.github.com/Alexmitter/0f5c230e4f3582f79c68ed172d50bf65
Sorry for that

ptitSeb:
I should receive a risc-v board soon, and will work on adding dynarec support there. Can I merge this PR in it's current state as a starting point?

Alexmitter:
@ptitSeb I waited for feedback from you to remove the WIP tag. I believe it is ready to merge as a starting point, please feel free to.

PR #223, opened 2022-02-02
Correct package name in Debian changelog

CalcProgrammer1:
The package name is "box86" in changelog and it prevents building .deb packages.


ryanfortner:
Thanks @CalcProgrammer1 !

Issue #266, opened 2022-04-04
Function parameters of type XID in libX11


mhavu:
I noticed that most functions in libX11 with a parameter of type `XID` list it as `p` in their signature, but `XSaveContext` lists it as `u`. `XID` is defined as `unsigned long` in *wrapped/wrappedlibx11.c*. Should the `XID` type be `L`, or is `p` correct? I assume `u`is wrong in any case.

ptitSeb:
`l` and `p` are correct, they will be both 8-bytes on 64bits (and 4-bytes on 32bits). `u` is incorrect on 64bits, but all `u` (and `i`) are upgrade to 8-bytes anyway on box64, so it's worong but it's automatically corrected.

PR #267, opened 2022-04-05
Added some libfontconfig and libX11 wrappings

mhavu:
Added wrapping for libfontconfig function `FcFontSetCreate` and libX11 functions `_XAllocID`, `_XAllocIDs`, and `XKillClient`. Harmonized the signature of XID type parameters to `L` in libX11 (see issue #266).


mhavu:
Hang on. Should `_XAllocIDs` be `GOM` to ensure alignment?

ptitSeb:
> Hang on. Should `_XAllocIDs` be `GOM` to ensure alignment?

No need, alignment will be the same on all plateforms

PR #382, opened 2022-08-11
Add SW64 base support

wannacu:
About SW64 support, there still are some problems about va function and signal handler.


src/emu/x64syscall.c
```diff
@@ -514,6 +517,12 @@ void EXPORT x64Syscall(x64emu_t *emu)
             R_EAX = (int)syscall(__NR_inotify_init1, 0);
             break;
         #endif
+        case 262:
+            R_EAX = (uint64_t)(int64_t)my_fstatat(emu, (int)R_RDI, (char*)R_RSI, (void*)R_RDX, (int)R_R10d);
+            break;
+        case 334:
```
ptitSeb, line 523: Can you comment on why this sys call is disabled.
Also, error should be "-1" and `errno = ENOSYS` set to indicate the error.

ptitSeb: yeah, me neither for now, but it's always good to document what is disabled, just in case.

PR #494, opened 2023-01-09
Add support for Rpi3 on 64-bit ARM (aarch64)

neildavis:
The Raspberry Pi 3 (and Zero 2) models use ARM Cortex-A53 CPUs with a 64-bit armv8 architecture (aarch64). As of Raspberry Pi OS 'Bullseye' there are 64-bit OS images available. Seems a shame that only tuned build options are available for RPi4.


ptitSeb:
Please add some how to use in `docs/COMPILE.md` also

neildavis:
> Please add some how to use in `docs/COMPILE.md` also

Done. Force pushed the branch

PR #507, opened 2023-02-03
Update x64emu.c

RainbowC0:
Added some adapations for Android as there is no __builtin_aarch64_get_fpcr and __builtin_aarch64_set_fpcr in Bionic.


ptitSeb:
Thanks!

PR #511, opened 2023-02-08
Fixed parameter order for BOX64_DYNAREC_HOTPAGE env parsing.

tristanmorgan:
The `sscanf()` needed the first two params swapped to correctly parse the env var.


ptitSeb:
Nice catch, thanks.

PR #523, opened 2023-02-21
LibreOffice enablement on Ubuntu Touch

fredldotme:
WIP branch around making LibreOffice x86_64 work on Ubuntu Touch aarch64.

This is part of an effort for my Winebox project which also ships Wine and registers the binfmt_misc handler using systemd.

I feel like I'll stick around with a few contributions every now and then, so stay tuned. :)


src/wrapped/wrappedgdk3.c
```diff
@@ -120,6 +120,11 @@ EXPORT void my3_gdk_window_remove_filter(x64emu_t* emu, void* window, void* f, v
     my->gdk_window_remove_filter(window, findFilterFct(f), data);
 }
 
+EXPORT unsigned int my3_gdk_threads_add_idle(x64emu_t* emu, void* function, void* data)
+{
+    my->gdk_threads_add_idle(function, data);
```
ptitSeb, line 125: You probably want to use a callback tranformation for `function`. Something like `findGSourceFuncFct` but you'll have to copy it from `wrappedglib2.c` first.

src/wrapped/wrappedxml2_private.h
```diff
@@ -250,7 +250,7 @@ GO(xmlCtxtGetLastError, pFp)
 GO(xmlCtxtReadDoc, pFppppi)
 GO(xmlCtxtReadFd, pFpippi)
 GO(xmlCtxtReadFile, pFpppi)
-//GOM(xmlCtxtReadIO, pFEpBBpppi)
+GO(xmlCtxtReadIO, pFppppppi)
```
ptitSeb, line 253: the `B` in the signature means callBacj, so they need special care with a `find_XXXX` and all.

src/wrapped/wrappedgdk3.c
mcagabe19: lol

fredldotme:
I haven't tested it in a while, but I'd guess things have changed and this PR is not necessary anymore, hence closing.

ptitSeb:
Yeah, lots have changed in the Wrapping of libraries department!

PR #526, opened 2023-02-24
[DYNAREC] Fix a couple of typos

julliard:


ptitSeb:
Nice catch, thank you!

PR #552, opened 2023-03-12
[RV64_DYNAREC] Added POP reg opcode

ksco:
Just make sure this is the right way to add new instruction support, next PR will be bigger.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_00.c
+++ b/src/dynarec/rv64/dynarec_rv64_00.c
@@ -63,12 +63,8 @@ uintptr_t dynarec64_00(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
         case 0x57:
             INST_NAME("PUSH reg");
             gd = xRAX+(opcode&0x07)+(rex.b<<3);
-            if(gd==xRSP) {
-                MV(x1, gd);
-                gd = x1;
-            }
+            SD(gd, xRSP, -8);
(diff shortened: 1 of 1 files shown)


ptitSeb:
Looks good to me.
But don't add too many opcode at a time. There are still some bugs in the Dynarec for now, and they need to be found before adding mass of opcode.

ksco:
Ok, then I keep the patch size small, so what family of instructions can be added now?

ptitSeb:
I have pushed some fixes. It seems fine now.
Best way to proceed is to see `BOX64_DYNAREC_LOG box64 ../test/test01` output and look for the 1st non-implemented opcde and to add it. Then check `ctest` is still ok. And from time to time, check that some bigger program still works. I use `box64 ./7z b` to see if it's still works (but I don't let it finish, it's too slow for now)

I'll probably merge this PR soon.

If you miss some macro, look at arm64 dynarec, or ask me.

ksco:
Thanks for the advice, seems approachable.

> I'll probably merge this PR soon.

Do you mean PR #550? Current PR #552 is already merged.

ptitSeb:
Yes, I meant the #550 
I wanted the infrastructure to be ready and something that compile and run, but I don't want this branch to live too long...

ksco:
Ok, reasonable, if #550 gets merged, I'll PR directly to `main`.

ptitSeb:
yep, no problem.

PR #553, opened 2023-03-12
[RV64_DYNAREC] Added 29 SUB opcode

ksco:
It's indeed a big chunk of code just to simulate flags and it's hard to test. I wonder if is there any user mode test suite for x86_64.

Also, [Csmith](https://github.com/csmith-project/csmith) might be useful, but I haven't tried it yet.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_emit_math.c
+++ b/src/dynarec/rv64/dynarec_rv64_emit_math.c
@@ -1,0 +1,89 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <stddef.h>
+#include <pthread.h>
+#include <errno.h>
+
+#include "debug.h"
+#include "box64context.h"
+#include "dynarec.h"
+#include "emu/x64emu_private.h"
+#include "emu/x64run_private.h"
+#include "x64run.h"
+#include "x64emu.h"
+#include "box64stack.h"
+#include "callback.h"
+#include "emu/x64run_private.h"
+#include "x64trace.h"
+#include "dynarec_native.h"
+#include "../tools/bridge_private.h"
+
+#include "rv64_printer.h"
+#include "dynarec_rv64_private.h"
+#include "dynarec_rv64_functions.h"
+#include "dynarec_rv64_helper.h"
+
+// emit SUB32 instruction, from s1, s2, store result in s1 using s3 and s4 as scratch
+void emit_sub32(dynarec_rv64_t* dyn, int ninst, rex_t rex, int s1, int s2, int s3, int s4, int s5)
+{
+    IFX(X_PEND) {
+        SDxw(s1, xEmu, offsetof(x64emu_t, op1));
+        SDxw(s2, xEmu, offsetof(x64emu_t, op2));
+        SET_DF(s3, rex.w?d_sub64:d_sub32);
+    } else IFX(X_ALL) {
+        SET_DFNONE(s3);
+    }
+
+    IFX(X_AF) {
+        // for later borrow chain calculation
+        NOT(s5, s1);
+    }
+
+    SUBxw(s1, s1, s2);
+
+    IFX(X_PEND) {
+        SDxw(s1, xEmu, offsetof(x64emu_t, res));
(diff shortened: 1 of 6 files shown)


src/dynarec/rv64/dynarec_rv64_emit_math.c
```diff
@@ -0,0 +1,89 @@
...
+        // bc = (res & (~op1 | op2)) | (~op1 & op2)
+        OR(s3, s5, s2);
+        AND(s4, s1, s3);
+        AND(s5, s5, s2);
+        OR(s4, s4, s5);
+        IFX(X_AF) {
+            // af = bc & 0x8
+            ANDI(s3, s4, 8);
+            BEQZ(s3, 4);
+            ORI(xFlags, xFlags, 1 << F_AF);
```
ptitSeb, line 59: you probably need to clear that bit before

src/dynarec/rv64/dynarec_rv64_emit_math.c
```diff
@@ -0,0 +1,89 @@
...
+        }
+        IFX(X_CF) {
+            // cf = bc & (rex.w?(1<<63):(1<<31))
+            SRLI(s3, s4, rex.w?63:31);
+            BEQZ(s3, 4);
+            ORI(xFlags, xFlags, 1 << F_CF);
```
ptitSeb, line 65: and all the other too. Some ANDI with `~((1<<F_AF) | (1<<F_CF)....)` might do the trick. Need to check if it fits the 12bits (and the sign bit)

src/dynarec/rv64/dynarec_rv64_helper.c
```diff
@@ -320,4 +321,21 @@ void rv64_move64(dynarec_rv64_t* dyn, int ninst, int reg, int64_t val)
...
+    SRLI(s3, s1, 3);
+    ANDI(s3, s3, 28);
+    ADD(s4, s4, s3);
+    LW(s4, s4, 0);
+    NOT(s4, s4);
+    SRLW(s4, s4, s1);
+    ANDI(s4, s4, 1);
+
+    BEQZ(s4, 4);
+    ORI(xFlags, xFlags, 1 << F_PF);
```
ptitSeb, line 340: need to clear F_PF bit before.

ptitSeb (changes requested):
The xFlags bits need to be cleared.

ksco:
This seems not right, all tests are failing now.

ksco:
Ok, back to normal, please review again. @ptitSeb

src/dynarec/rv64/dynarec_rv64_emit_math.c
```diff
@@ -0,0 +1,92 @@
...
+#include "rv64_printer.h"
+#include "dynarec_rv64_private.h"
+#include "dynarec_rv64_functions.h"
+#include "dynarec_rv64_helper.h"
+
+// emit SUB32 instruction, from s1, s2, store result in s1 using s3 and s4 as scratch
+void emit_sub32(dynarec_rv64_t* dyn, int ninst, rex_t rex, int s1, int s2, int s3, int s4, int s5)
+{
+    // clear flags
+    ANDI(xFlags, xFlags, ~((1UL<<F_AF) | (1UL<<F_CF) | (1UL<<F_OF) | (1UL<<F_ZF) | (1UL<<F_SF) | (1UL<<F_PF)));
```
ptitSeb, line 30: do that only if some flags are requested. No need to clear them if it's just a SUB. `IFX(X_ALL)` should do the trick

ptitSeb (changes requested):
One last change (an optimization) and it's good to go.
Nice job!

src/dynarec/rv64/dynarec_rv64_emit_math.c
ptitSeb: Maybe a macro, like 'CLEAR_FLAGS()` could be writen, to ease the writting of all the othe stuff?

ptitSeb (approved):

PR #555, opened 2023-03-13
[RV64_DYNAREC] Added 8B MOV opcode

xctan:


src/dynarec/rv64/rv64_emitter.h
```diff
@@ -221,6 +221,8 @@ f28–31  ft8–11  FP temporaries                  Caller
 
 // rd = [rs1 + imm12]
 #define LD(rd, rs1, imm12)          EMIT(I_type(imm12, rs1, 0b011, rd, 0b0000011))
+// rd = [rs1 + imm12]
+#define LDxw(rd, rs1, imm12)        EMIT(I_type(imm12, rs1, 0b011<<(1-rex.w), rd, 0b0000011))
```
ptitSeb, line 225: I don't think this is correct. 
it should be `EMIT(I_type(imm12, rs1, 0b010+rex.w, rd, 0b0000011))`

ptitSeb (changes requested):
I don't think `LDwx` emitter is correct.

src/dynarec/rv64/rv64_emitter.h
xctan: If funct3 is 0b010, the instruction would be LW and the result is sign-extended to 64-bits.

ptitSeb: Yes, indeed, I double checked and you are right, it's better to use LWU.

ptitSeb (approved):

PR #559, opened 2023-03-14
[RV64_DYNAREC] Added C1 /7 SAR opcode

xctan:


ksco:
I'm not sure, but is there any chance we can simply clear out some flags? e.g. `c >= rex.w ? 64 : 32`?

ksco:
Oh it's a shift right arithmetic, then we need to peek at MSB also, but it might result in fewer RISC-V instructions.

xctan:
Still work in progress, currently 32-bit version is not implemented, and some flags can be optimized.

src/dynarec/rv64/dynarec_rv64_emit_shift.c
```diff
@@ -0,0 +1,72 @@
...
+    }
+    if(!c) {
+        IFX(X_PEND) {
+            SDxw(s1, xEmu, offsetof(x64emu_t, res));
+        }
+        return;
+    }
+    IFX(X_CF) {
+        SRAI(s3, s1, c-1);
+        ANDI(s3, s3, 1); // LSB
```
ptitSeb, line 47: maybe add a special case when c==1, no neeed for srai, just an andi is enough

ptitSeb:
There is a branch conflict now, sorry. Can you fix it @xctan ?

xctan:
Sure.

src/dynarec/rv64/dynarec_rv64_emit_shift.c
xctan: Fixed.

ksco (approved):

PR #567, opened 2023-03-15
[RV64_DYNAREC] Added 80 /7 CMP opcode

ksco:


src/dynarec/rv64/dynarec_rv64_helper.h
```diff
@@ -76,6 +76,29 @@
...
+                        wback = xRAX+(nextop&7)+(rex.b<<3);     \
+                        wb2 = 0;                \
+                    } else {                    \
+                        wback = (nextop&7);     \
+                        wb2 = (wback>>2)*8;     \
+                        wback = xRAX+(wback&3); \
+                    }                           \
+                    MV(i, wback);               \
+                    ANDI(i, i, 0xff);           \
+                    if (wb2) SLLI(i, i, wb2);   \
```
ptitSeb, line 91: I think the logic should be the oposite:
SRLI first, than AND.

src/dynarec/rv64/dynarec_rv64_helper.h
```diff
@@ -104,14 +127,19 @@
...
         }                                                                 \
         IFX(X_CF) {                                                       \
-            /* cf = bc & (rex.w?(1<<63):(1<<31)) */                       \
-            SRLI(scratch1, scratch2, rex.w?63:31);                        \
+            /* cf = bc & (1<<(width-1)) */                                \
+            if (width == 8) {                                             \
+                ANDI(scratch1, scratch2, 0x80);                           \
+            } else {                                                      \
+                SRLI(scratch1, scratch2, width-1);                        \
+                if (width == 16) ANDI(scratch1, scratch1, 1);             \
```
ptitSeb, line 135: What is this special case with width==16?

xctan (approved):

src/dynarec/rv64/dynarec_rv64_helper.h
```diff
@@ -104,14 +127,19 @@
...
+            }                                                             \
             BEQZ(scratch1, 4);                                            \
             ORI(xFlags, xFlags, 1 << F_CF);                               \
         }                                                                 \
         IFX(X_OF) {                                                       \
-            /* of = ((bc >> rex.w?62:30) ^ (bc >> rex.w?63:31)) & 0x1; */ \
-            SRLI(scratch1, scratch2, rex.w?62:30);                        \
+            /* of = ((bc >> (width-2)) ^ (bc >> (width-1))) & 0x1; */     \
+            SRLI(scratch1, scratch2, width-2);                            \
             SRLI(scratch2, scratch1, 1);                                  \
```
ptitSeb, line 143: This should be
```C
SRLI(scratch1, scratch2, width-1);
SRLI(scratch2, scratch1, 1);
```
or that will make too much right shifting

src/dynarec/rv64/dynarec_rv64_helper.h
ksco: Yes, you're right.

src/dynarec/rv64/dynarec_rv64_helper.h
ptitSeb: Oh, it's because for witdh==32 or 64, the ANDI is useless? Is that really true for width==32?

ptitSeb: I suppose the ZEROUP has already be called?

src/dynarec/rv64/dynarec_rv64_emit_tests.c
```diff
@@ -23,6 +23,69 @@
...
+        SET_DFNONE(s4);
+    }
+
+    IFX(X_AF | X_CF | X_OF) {
+        // for later flag calculation
+        NOT(s5, s1);
+    }
+
+    // It's a cmp, we can't store the result back to s1.
+    SUB(s6, s1, s2);
```
ptitSeb, line 43: Don't we need an `ANDI(s6, s6, 0xff);` also?

src/dynarec/rv64/dynarec_rv64_helper.h
ksco: > I suppose the ZEROUP has already be called?

Yes, when width==32, ZEROUP is called outside.

src/dynarec/rv64/dynarec_rv64_helper.h
ksco: Fixed and force-pushed.

src/dynarec/rv64/dynarec_rv64_emit_tests.c
ptitSeb: X_ZF also needs it.
And also might need to put back the value in a register,

It safer to be sure the value is 8bits only, so I think a systematic ANDI is better.

src/dynarec/rv64/dynarec_rv64_emit_tests.c
```diff
@@ -23,6 +23,69 @@
...
+        emit_pf(dyn, ninst, s6, s3, s4);
+    }
+}
+
+// emit CMP8 instruction, from cmp s1 , 0, using s3 and s4 as scratch
+void emit_cmp8_0(dynarec_rv64_t* dyn, int ninst, int s1, int s3, int s4) {
+    CLEAR_FLAGS();
+    IFX_PENDOR0 {
+        SB(s1, xEmu, offsetof(x64emu_t, op1));
+        ADDI(s4, xZR, 0);
```
ptitSeb, line 67: we probably don't need `s4` here and can use `xZR` all along

src/dynarec/rv64/dynarec_rv64_helper.h
ptitSeb: yeah, 16bits operation are not very common on 64bits programs...

src/dynarec/rv64/dynarec_rv64_emit_tests.c
ksco: > X_ZF also needs it.

Indeed, force-pushed.

src/dynarec/rv64/dynarec_rv64_emit_tests.c
ksco: Yes, we don't need `s4` here, I guess we can just `SB(xZR, xEmu, offsetof(x64emu_t, op2));`? `s4` is indeed written in `emit_pf`, but that shouldn't matter?

ptitSeb: Yes. It was just to also remove it completly from the function call, but we have to keep it for emit_pf...

ksco: Force-pushed.

PR #571, opened 2023-03-16
[RV64_DYNAREC] Added (81/83) /7 CMP opcode

ksco:


ksco:
`MOV Reg, Id` also requires `MOV32w`, so I pushed it to current PR.

src/dynarec/rv64/rv64_emitter.h
```diff
@@ -108,8 +108,10 @@ f28–31  ft8–11  FP temporaries                  Caller
 #define SPLIT20(A)  (((A)+0x800)>>12)
 #define SPLIT12(A)  ((A)&0xfff)
 
-// MOVE64x is quite complex, so use a function for this
+// MOV64x/MOV32w is quite complex, so use a function for this
 #define MOV64x(A, B)    rv64_move64(dyn, ninst, A, B)
+#define MOV32w(A, B)    rv64_move32(dyn, ninst, A, B)
```
ptitSeb, line 113: `rv64_move32'...)` does not handle zeroup, but `rev64_move64(...)` does, that why I didn't created `MOV32w(...)` before, there is no gain in having different macro.

ptitSeb (changes requested):
MOV32w will not ZEROUP in it's current state.

ptitSeb:
Also, conflict on the branch here @ksco

ksco:
I rebased and pushed a new commit handling MOV32w.

src/dynarec/rv64/rv64_emitter.h
```diff
@@ -108,8 +108,10 @@ f28–31  ft8–11  FP temporaries                  Caller
 #define SPLIT20(A)  (((A)+0x800)>>12)
 #define SPLIT12(A)  ((A)&0xfff)
 
-// MOVE64x is quite complex, so use a function for this
+// MOV64x/MOV32w is quite complex, so use a function for this
 #define MOV64x(A, B)    rv64_move64(dyn, ninst, A, B)
+#define MOV32w(A, B)    do{ rv64_move32(dyn, ninst, A, B); if((A&0xffffffffLL)==A && (A&0x80000000)) ZEROUP(A); }while(0);
```
ptitSeb, line 113: I think just the `A&0x80000000` test is enough, as we might use `MOV32w(reg, -1)` and the the first mask test might not work if -1 is considered int64 (wich might happens on some cases)

PR #572, opened 2023-03-16
[RV64_DYNAREC] Optimize MOV64 emitter

ksco:


ptitSeb:
It's a nice optim.

But rv64_move32 stiill does signextension.

ksco:
> But rv64_move32 still does sign extension.

Yes, this is related to #571, I'll fix this there.

PR #576, opened 2023-03-16
[RV64] Enable test for riscv64 in CI

xctan:
Box64 works inside qemu-user, so we can actually run the test in github action. I tried to use ctest in the container, but it failed for unknown reason (and I can't reproduce on my machine), so I wrote a script (translated from CMakeLists.txt) to do the tests.

### DIFF
--- a/.github/runtests.sh
+++ b/.github/runtests.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+run_test() {
+  id=$1
+  append=$2
+
+  echo "Test $id"
+  ./box64 ../tests/test$id $append > /tmp/test$id.log
+  if [ $? -ne 0 ]; then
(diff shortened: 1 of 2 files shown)


xctan:
I'm not sure whether the problem of ctest can be fixed... but placing tests in two separate places is a not a promising idea for maintenance.

xctan:
> I prefer simple maintenance for sure.

I probably figured out the issue. We are building box64 outside the qemu-user container, and path of tools like cmake are placed at different locations. Without reconfigure and rebuild, ctest will refuse to work.

ptitSeb:
It makes sense. ctest is part of cmake tools, so without a preper build done, ctest has nothing to do.

xctan:
Another better attempt is available. https://github.com/ptitSeb/box64/pull/583

PR #577, opened 2023-03-17
[RV64_DYNAREC] Added C1 /(4/6) SHL opcode

ksco:
test03 and test04 now run without `Dynarec stopped because of Opcode`.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_emit_shift.c
+++ b/src/dynarec/rv64/dynarec_rv64_emit_shift.c
@@ -23,15 +23,76 @@
 #include "dynarec_rv64_functions.h"
 #include "dynarec_rv64_helper.h"
 
+// emit SHL32 instruction, from s1 , constant c, store result in s1 using s3, s4 and s5 as scratch
+void emit_shl32c(dynarec_rv64_t* dyn, int ninst, rex_t rex, int s1, uint32_t c, int s3, int s4, int s5)
+{
+    CLEAR_FLAGS();
+    IFX(X_PEND) {
+        if (c) {
+            MOV64x(s3, c);
+            SDxw(s3, xEmu, offsetof(x64emu_t, op2));
+        } else SDxw(xZR, xEmu, offsetof(x64emu_t, op2));
+        SDxw(s1, xEmu, offsetof(x64emu_t, op1));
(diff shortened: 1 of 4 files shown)


src/dynarec/rv64/dynarec_rv64_emit_shift.c
```diff
@@ -23,19 +23,106 @@
...
+            MOV64x(s3, c);
+            SDxw(s3, xEmu, offsetof(x64emu_t, op2));
+        } else SDxw(xZR, xEmu, offsetof(x64emu_t, op2));
+        SDxw(s1, xEmu, offsetof(x64emu_t, op1));
+        SET_DF(s4, rex.w?d_shl64:d_shl32);
+    } else IFX(X_ALL) {
+        SET_DFNONE(s4);
+    }
+
+    if (c >= (rex.w?64:32)) {
```
xctan, line 41: Not necessary. The constant is ensured to be less than the width of operand register.

```c
u8 = (F8)&(rex.w?0x3f:0x1f);
```

And x86 will [mask](https://www.felixcloutier.com/x86/sal:sar:shl:shr) the shift count before doing shift.

> The count is masked to 5 bits (or 6 bits if in 64-bit mode and REX.W is used).

src/dynarec/rv64/dynarec_rv64_emit_shift.c
```diff
@@ -23,19 +23,106 @@
...
+            SDxw(s3, xEmu, offsetof(x64emu_t, op2));
+        } else SDxw(xZR, xEmu, offsetof(x64emu_t, op2));
         SDxw(s1, xEmu, offsetof(x64emu_t, op1));
-        SDxw(s3, xEmu, offsetof(x64emu_t, op2));
         SET_DF(s4, rex.w?d_shr64:d_shr32);
     } else IFX(X_ALL) {
         SET_DFNONE(s4);
     }
+
+    if (c >= (rex.w?64:32)) {
```
xctan, line 117: Same above.

xctan (changes requested):

src/dynarec/rv64/dynarec_rv64_emit_shift.c
ksco: Indeed.

PR #583, opened 2023-03-18
Enable test for RV64 and ARM64 in CI

ksco:
If we can pass the environment variable `QEMU_LD_PREFIX=/usr/{riscv64,aarch64}-linux-gnu/lib/`
to ctest, it should also work and be cleaner, but I don't know how to do it as
I am not familiar with cmake.

Note that TEGRAX1 is not currently working because of a mismatch in GLIBC version.
Therefore, we skipped this platform.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -92,10 +92,20 @@ jobs:
           make -j$(nproc) VERBOSE=1
 
       - name: "Test Box64"
-        if: ${{ matrix.platform == 'X64' }}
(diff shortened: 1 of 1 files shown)


ksco:
> pass the environment variable QEMU_LD_PREFIX=/usr/riscv64-linux-gnu/lib/ to ctest

@ptitSeb Do you have any idea we can do this? In this way, we can also enable tests for other architectures.

ptitSeb:
NIce! Thanks :)

PR #591, opened 2023-03-19
[RV64_DYNAREC] Added more opcode

ksco:


ksco:
I noticed that the `fork` test occasionally fails, and it also occurs when specifying more CPUs than the actual number on the local machine with the `-j` option. @ptitSeb, do you know what the reason for this is?

ptitSeb:
the test i supposed to be reproducible, but sometime, the text output can be different. It's a test issue.

src/dynarec/rv64/dynarec_rv64_00.c
```diff
@@ -499,6 +519,21 @@ uintptr_t dynarec64_00(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+            INST_NAME("MOV xH, Ib");
+            u8 = F8;
+            MOV32w(x1, u8);
+            if(rex.rex) {
+                gb1 = xRAX+(opcode&7)+(rex.b<<3);
+            } else {
+                gb1 = xRAX+(opcode&3);
+                SRLI(x1, x1, 8);
+            }
+            ANDI(gb1, x1, 0xff);
```
ptitSeb, line 535: I don't think this is correct here.

ptitSeb (changes requested):
B4..B7 opcode seems wrong. I would expect an AND to mask and an OR to inject the value.

src/dynarec/rv64/dynarec_rv64_00.c
ksco: Yes, I made this mistake again.

src/dynarec/rv64/dynarec_rv64_00.c
```diff
@@ -528,11 +528,16 @@ uintptr_t dynarec64_00(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             MOV32w(x1, u8);
             if(rex.rex) {
                 gb1 = xRAX+(opcode&7)+(rex.b<<3);
+                LUI(x2, 0xffff0);
```
ptitSeb, line 531: it's 8bits value that is injected, not a 16bits

src/dynarec/rv64/dynarec_rv64_00.c
```diff
@@ -528,11 +528,16 @@ uintptr_t dynarec64_00(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+                AND(gb1, gb1, x2);
+                OR(gb1, gb1, x1);
             } else {
                 gb1 = xRAX+(opcode&3);
-                SRLI(x1, x1, 8);
+                MOV64x(x2, 0xffffffff0000ffffLL);
```
ptitSeb, line 536: it's 8bits value that injected too, still not a 16bits

ptitSeb (changes requested):
8bits mask, not 16bits

ksco:
Sorry, fixed.

src/dynarec/rv64/dynarec_rv64_00.c
```diff
@@ -499,6 +519,25 @@ uintptr_t dynarec64_00(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+        case 0xB4:
+        case 0xB5:
+        case 0xB6:
+        case 0xB7:
+            INST_NAME("MOV xH, Ib");
+            u8 = F8;
+            MOV32w(x1, u8);
+            if(rex.rex) {
+                gb1 = xRAX+(opcode&7)+(rex.b<<3);
+                ANDI(gb1, gb1, (int32_t)-256);
```
ptitSeb, line 531: `~0xff` works fine too

ksco: Yeah, it's more readable, force-pushed.

src/dynarec/rv64/dynarec_rv64_helper.h
```diff
@@ -233,6 +247,15 @@
     j64 = (dyn->insts)?(dyn->insts[ninst].epilog-(dyn->native_size)):0;\
     B(j64)
 
+// Branch to MARKSEG if reg is 0 (use j64)
+#define CBZ_MARKSEG(reg)    \
+    j64 = GETMARKSEG-(dyn->native_size);   \
+    BEQZ_safe(reg, j64);
```
ptitSeb, line 253: no need to use the safe version here, the jump will be withing reach all the time for MARK* conditionnal jump

ptitSeb (approved):

src/dynarec/rv64/dynarec_rv64_helper.h
ksco: Fixed, but I don't really understand how `maskseg` works, just copied from ARM port, can you explain a little?

ptitSeb:
all the "MARK" suff are just simple Label and Jmp_to_label.
You define the label with `MARKSEG;` macro, and jump there with the `CBZ_MARKSEG(reg);` one. The multi-pass Dynarec infrastructure will hanlde the computation of the offset. The limit is that all `MARK*` label must reside inside the same x64 instruction, because all offset are reset when starting a new one, but that not a problem as all the logic in inside 1 x64 opcode anyway.
The only execption is the `XXX_NEXT(...)` conditionnal jump, that alow to jump to the next x64 instruction (so outside current instruction).

Is it clear now?

ksco:
Yes, and it's clever!

ptitSeb:
I posted a small youtube video on reddit with box64 in action: https://www.reddit.com/r/RISCV/comments/11vtqkf/video_of_superhexagon_running_on_visonfive2_riscv/

ksco:
This is so cool! My VisionFive2 is on the way, can't wait.

PR #623, opened 2023-03-23
[ARM64_DYNAREC] Fixed UFLAG_OP12 helper macro

ksco:
I saw @ptitSeb fixed this on the RISC-V port at https://github.com/ptitSeb/box64/commit/b2484afe96ecb2e68150432fd7f8fe05b26dabee, but it's actually copied from ARM side...


ptitSeb:
It was working fine on ARM64 because xEmu is 0, but it's better to fix it for sure.

PR #626, opened 2023-03-23
Update README

ksco:
It is recommended to insert spaces between Chinese characters and English, numbers, and symbols because it can improve readability. In addition, README_CN has been synchronized to the latest.


ptitSeb:
Thanks!

PR #649, opened 2023-03-29
[RV64_DYNAREC] Added DF F0+i FCOMIP and DF /0 FILD opcodes

xctan:


xctan:
If correct behavior on NaN is not important, FCOMIP can be implemented with fewer instructions.

src/dynarec/rv64/dynarec_rv64_df.c
```diff
@@ -0,0 +1,119 @@
...
+            INST_NAME("FCOMIP ST0, STx");
+            SETFLAGS(X_ALL, SF_SET);
+            v1 = x87_get_st(dyn, ninst, x1, x2, 0, X87_COMBINE(0, nextop&7));
+            v2 = x87_get_st(dyn, ninst, x1, x2, nextop&7, X87_COMBINE(0, nextop&7));
+            CLEAR_FLAGS();
+            IFX(F_ZF | F_PF | F_CF) {
+                if(ST_IS_F(0)) {
+                    FEQS(x5, v1, v1);
+                    FEQS(x4, v2, v2);
+                    AND(x5, x5, x4);
```
ptitSeb, line 61: Shouldn't this be a OR, as we want to detect any input being NaN.

ptitSeb:
I think NaN behaviour is important. maybe we could try some optinnal flags to disable nan behaviour. Maybe use the BOX64_DYNAREC_FASTNAN env. var. that already exist? but this one is off by default while I think this should be on by default, so, not sure.

src/dynarec/rv64/dynarec_rv64_df.c
xctan: If v1 is NaN, v1 != v1, so x5 is set to zero. AND here ensures that both v1 and v2 are *not* NaN.

ptitSeb: Ah yeah, right, sorry.

xctan:
Is there any other problem with this PR? NaN improvement can be made in future PRs.

PR #650, opened 2023-03-29
[RV64_DYNAREC] Added some 66 0F Pxxx opcodes

ksco:
I'd like to keep the patch size relatively small, as there are still missing facilities in the SSE infra.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_660f.c
+++ b/src/dynarec/rv64/dynarec_rv64_660f.c
@@ -149,7 +149,13 @@ uintptr_t dynarec64_660F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int 
             AND(x1, x1, x5);
             OR(gd, gd, x1);
             break;
-
+        case 0xDB:
+            INST_NAME("PAND Gx,Ex");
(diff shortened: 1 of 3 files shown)


src/dynarec/rv64/dynarec_rv64_660f.c
```diff
@@ -107,7 +107,34 @@ uintptr_t dynarec64_660F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+            nextop = F8;
+            switch((nextop>>3)&7) {
+                case 2:
+                    INST_NAME("PSRLD Ex, Ib");
+                    GETEX(x1, 1);
+                    u8 = F8;
+                    if(u8) {
+                        if (u8>31) {
+                            // just zero dest
+                            SD(xZR, x1, 0);
```
ptitSeb, line 120: You need to use `fixedaddress+0` and `fixedaddress+8` for EX

ksco: Yes, sorry.

ksco: Fixed.

PR #653, opened 2023-03-30
[RV64_DYNAREC] Added 0F 2E,2F opcode & some fixes in printer

ksco:
Please review this one carefully, don't know if I understand the corresponding SSE infra correctly. I believe it's right, but still not so sure.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_0f.c
+++ b/src/dynarec/rv64/dynarec_rv64_0f.c
@@ -200,7 +200,43 @@ uintptr_t dynarec64_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             SD(x3, wback, fixedaddress+0);
             SD(x4, wback, fixedaddress+8);
             break;
-
+        case 0x2E:
+            // no special check...
+        case 0x2F:
+            if(opcode==0x2F) {INST_NAME("COMISS Gx, Ex");} else {INST_NAME("UCOMISS Gx, Ex");}
+            SETFLAGS(X_ALL, SF_SET);
+            nextop = F8;
+            GETGX(x3);
+            GETEXSS(v0, 0);
+            CLEAR_FLAGS();
+            // if isnan(gd) || isnan(v0)
+            IFX(X_ZF | X_PF | X_CF) {
+                FLW(gd, x3, 0);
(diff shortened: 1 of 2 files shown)


ksco:
test15 failed, investigating.

ksco:
It's actually test16 is failing, but why CI says test15?

ptitSeb:
> test15 failed, investigating.

Because 1 test is skipped and it messed with the count. It's anoying, I know.

ksco:
Side note: cosim doesn't seem to compare flags?

ptitSeb:
> Side note: cosim doesn't seem to compare flags?

No, that's not possible because flags are optimized by the Dynarec. You can only observe consequences of flags (like a conditionnal jump is not correct, or a setxx, things like that...)

ksco:
Can we disable flag optimizations and just calculate all flags when running in `BOX64_DYNAREC_TEST` mode?

ptitSeb:
> Can we disable flag optimizations and just calculate all flags when running in `BOX64_DYNAREC_TEST` mode?

That's not what we want to do. Flags optimisation is an important part of the Dynarec, so we want to compare the Dynarec as-is, not some "carefull" version of it.

ksco:
Ok, makes sense again, thanks for answering. I have no more questions, this PR is now ready for review, CI passed.

src/dynarec/rv64/dynarec_rv64_0f.c
```diff
@@ -200,7 +200,43 @@ uintptr_t dynarec64_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+                FLTS(x2, gd, v0);
+                BEQ_MARK2(x2, xZR);
+                ORI(xFlags, xFlags, 1<<F_CF);
+                B_NEXT_nocond;
+            }
+            MARK2;
+            // else if gd == v0
+            IFX(X_ZF) {
+                FEQS(x2, gd, v0);
+                BEQ_MARK3(x2, xZR);
```
ptitSeb, line 235: a branch to NEXT could be used here also...

ksco: Yeah, will do it in next PR.

PR #655, opened 2023-03-31

ksco:
test16 is almost running fully in Dynarec now, except for `cvtsd2si`. At first glance, float -> int conversion seems tricky.


ptitSeb:
Yeah int coversion is tricky because all invalid opration should endup as `MININT` value, so `0x80000000` for `cvtsd2si`. I'll give it a try later.

PR #661, opened 2023-04-02
[RV64_DYNAREC] Small fixes caught by cosim

ksco:


src/dynarec/rv64/dynarec_rv64_helper.h
```diff
@@ -66,6 +66,7 @@
 #define GETED(D)  if(MODREG) {                          \
                     ed = xRAX+(nextop&7)+(rex.b<<3);    \
                     wback = 0;                          \
+                    if (!rex.w) ZEROUP(ed);              \
```
ksco, line 69: @ptitSeb  This is wrong, sorry. Fixing..

ptitSeb: oh, yeah, that's a bit erleay to zeroup the value!

PR #662, opened 2023-04-02
[RV64_DYANREC] Fixed GETED, again

ksco:
Previous fix(https://github.com/ptitSeb/box64/commit/2a346ef0b951ce9a2dd7296fbad15f24b7566971) is problematic when ed is the source reg. Don't know if there are better ways, but this is what I came up with.


ptitSeb:
I don't like having a new parameter to GETED (and a C++ one also). I would rather ZEROUP when needed, and let GetED as it was before.
A macro like `if(MODREG) ZEROUP(ed)` can be imagined maybe, and used in the few cases it needs to be?

ptitSeb:
Or maybe a different macro, like `GETEDxw(a, D)`?

ksco:
Yeah, you're right. I should do this inside the opcode itself.

PR #671, opened 2023-04-06
[RV64_DYNAREC] Added more opcodes for Stardew Valley

ksco:
Note that Stardew Valley as well as the SSE2 test from simde stopped working after https://github.com/ptitSeb/box64/commit/49a3255a476f0904db2f054d9fba3a6e57b3fbe8. I haven't found the cause yet, but I think it's not a problem with `MOVDQU` itself.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_00.c
+++ b/src/dynarec/rv64/dynarec_rv64_00.c
@@ -870,7 +870,7 @@ uintptr_t dynarec64_00(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 if(gd!=ed) {    // it's sometimes used as a 3 bytes NOP
                     MV(gd, ed);
                 }
-                else if(!rex.w) {
+                if(!rex.w) {
                     ZEROUP(gd);   //truncate the higher 32bits as asked
                 }
             }
--- a/src/dynarec/rv64/dynarec_rv64_660f.c
+++ b/src/dynarec/rv64/dynarec_rv64_660f.c
@@ -198,10 +198,7 @@ uintptr_t dynarec64_660F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int 
             nextop = F8;
             GETGX(x1);
             GETEX(x2, 0);
-            LD(x3, wback, fixedaddress+0);
-            LD(x4, wback, fixedaddress+8);
-            SD(x3, gback, 0);
-            SD(x4, gback, 8);
+            SSE_LOOP_MV_Q(x3);
             break;
         case 0x70: // TODO: Optimize this!
             INST_NAME("PSHUFD Gx,Ex,Ib");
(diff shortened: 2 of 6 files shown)


ptitSeb:
But MOVDQU is a simple unalinged 128bits move from Ex to Gx. It's a simple one compared to other. I would say that MOVDQU alors more complex opcode that followed to be run using Dynarec and it those other opcode that fails for different reasons

ksco:
You may have misread what I mean, but yeah I don't think it's a problem with MOVDQU either. It’s a bit hard to find the real cause tho.

ptitSeb:
Yeah, I did mysread and edited my comment on 2nd read.

On a side note, I think I will roll back the recent changes in the interpreter about the removal of all isnan/isinf tests. At least I add to add back the isnan test on arm.
Note that converting a float/double that is NaN, INF or overflowing to an integer is Undefined Behavior in C. That means the result end up dependaning on the arch it's running on, and we probably don't want that for the interpreter.

ksco:
Oh sure, I didn't know that it's a UB.

ksco:
> Note that Stardew Valley as well as the SSE2 test from simde stopped working...

I noticed something in the SSE2 test log that says:

```
0x4000a4fcba: Dynarec stopped because of Opcode FE FF FF 48 8D 8D 70 FE FF FF 48 8B 95 B0 FD ( => /home/ksco/Develop/box64/build/iBZd:test_simde_mm_add_epi16 + 367)
```

This is nonsense because it's not a valid opcode at all. Also, I found some suspicious instruction sequences in `objdump`.

```
   1c9ad:       0f 29 85 00 fe ff ff    movaps %xmm0,-0x200(%rbp)
   1c9b4:       48 8d 8d 70 fe ff ff    lea    -0x190(%rbp),%rcx
```

Is this something wrong with `geted`?

ptitSeb:
I checked geted
The opcode here is `0f 29 85 00 fe ff ff`
image

So "nextop" should be `85`, and it's handled line 128 of `src/dynarec/rv64/dynarec_rv64_helper.c`

That nextop value is loaded line 190 of `src/dynarec/rv64/dynarec_rv64_0f.c`
All seems in order here. 

I do agree that the dynarec stop message is suspicious. Can you use `BOX64_DYNAREC_DUMP`  to check the assembly dump?

ksco:
https://ksco.cool/kZEY
I uploaded the full log with `BOX64_DYNAREC_DUMP=1 BOX64_DYNAREC_LOG=1`.

ptitSeb:
The issue is there a 2 `nextop = F8` in PADDW 66 0F FD opcode. You add the fix to your PR?

ksco:
Ok, SSE test works now, I'll have another try with Stardew Valley.

ksco:
Finally got Stardew Valley to work again! The issue is with `PSHUFD Gx,Ex,Ib` when doing an in-place shuffle. But the solution I came up with this is a bit cursed IMO.

ptitSeb:
I think it's good enough for now. There are some special case that could be optimized, as they are often used (when a elemnt is duplicated), but it will be for later, when the dynarec is mostly complete.

PR #676, opened 2023-04-08
[RV64_DYNAREC] Added more opcodes for SV and some fixes

ksco:
Tested with SV.


ksco:
Oh, something I should mention is that I noticed there are some `subpd` logs in ref17.txt, but there is no SUBPD opcode at all in the test17 binary file, is this expected?

ptitSeb:
Compiler do what they want to optimize things...

ksco:
Ok, but since we're testing the instruction itself, maybe we should consider turning off optimizations when compiling these tests?

ptitSeb:
Sure. Fill free to build a new version if you can have a better one, with more opcodes tested :)

PR #677, opened 2023-04-09
[TESTS] Disable compiler optimizations for test17 and some related fixes

ksco:


ksco:
Oh ARM counterparts are failing probably for the same reason too, but I know nothing about ARM..

ptitSeb:
How did you regereate the ref17.txt?

ksco:
I run it in an x86-64 native machine with `./test17 > ref17.txt`.

ptitSeb:
ok, then there is a issue with dynarec or interpreter for subpd. I probably didn't added some of the nan test to this one I suppose.

ksco:
It's probably something off with `maxsd/minsd`, and I've fixed the issue in RISC-V and interpreter counterparts.

ksco:
"If the values being compared are both 0.0s (of either sign), the value in the second source operand is returned." 
https://www.felixcloutier.com/x86/maxsd

We didn't deal with this case before, it probably happened to be optimized by gcc.

ptitSeb:
I'll merge and fix it afterward...

ptitSeb:
So, yeah, the issue in the MINSD/MAXSD test with 0 / -0 that should copy the arg.
TBH, It will add a lot of code and I'm not sure it's usefull in real life stuff

ksco:
Yeah it is probably not so useful, but we need to pass that test.

ptitSeb:
bah, I found an easy way to do it, tests are green for me locally.

ksco:
A clever trick!

ptitSeb:
same could probably be used on rv64 too, to avoid specific test of 0 value

ksco:
Let me try it out.

PR #680, opened 2023-04-09
[RV64_DYNAREC] Added more 66 0F opcode for SV

ksco:
Btw, SV can now boot into the first screen in about 15 minutes, which is some progress for sure. But I doubt if it can be playable even fully runs in Dynarec.


ptitSeb:
bah, we'll see. I do some perf analysis later, so if I can find something (not easy because most perf tool are not working on riscv)

PR #684, opened 2023-04-10
[RV64_DYNAREC] Added more opcodes for SV

ksco:


ksco:
How do we implement the `BSR` opcode? It is essentially a `Count Leading Zeros` operation, which will be super slow if we do a while loop like what is done in the interpreter.

ksco:
Something like https://github.com/ptitSeb/box64/blob/main/src/tools/bitutils.c#L11 will be faster, but it's hard to implement in assembly.

ptitSeb:
That deBruijn64 algorithm doesn't look too difficult to implement with emiter. I can do it later.

ptitSeb:
The algorithm is for trailing 0, so for BSF. It will not work for BSR, unless bit are reversed, but there is no opcode for that, and reversing bits is actually super complex.
BSF is trivial, BSR will need a bit of work. I'll write something with table and a short (unrolled) loop I think.

ksco:
Yeah, `Count Leading Zeros` and `Count Trailing Zeros` implementations are different, but they all need some kind of lookup table for sure.

[LeadingZeros64](https://github.com/golang/go/blob/96428e160de755d6b3b3b7dc1b9c81196957d679/src/math/bits/bits.go#L37)
[TrailingZeros64](https://github.com/golang/go/blob/96428e160de755d6b3b3b7dc1b9c81196957d679/src/math/bits/bits.go#L90)

ptitSeb:
Yeah, I started something similar, but with a shorter table (4bits instead of 8). But I need to debug it...

ksco:
Ack.

ptitSeb:
with latest PR, SV is much less sluggish :) Things are moving nicely!

ksco:
Yeah, first screen gets smoother, but it still needs ~15min to load.

ksco:
I seem to have read somewhere that you tried to run World of Goo? Is this game playable already?

ptitSeb:
I need to retry, but it was a bit long to load.
I'll think I'll do a video this weekend anyway, WorldOfGoo or StardewValley or SuperMeatBoy... What run better.
I tried also a full GOG installation, but it was yet too slow. But with more and more opecode, it will be interesting soon.

ptitSeb:
I pushed a new screenshot on reddit: https://www.reddit.com/r/RISCV/comments/12jv181/ut2004_running_on_a_visionfive2_with_box64_gl4es/

It's a bit slow, missing some opcodes in the dynarec... But it works, and doesn't even take 15min to load ;)

ksco:
Wow, just tried out, World of Goo runs so well on VF2!

ptitSeb:
Yes it does :)

ksco:
Also, where can I buy a copy of UT2004? I saw you say that it can now run at 20fps sometimes, so I think it has more potential than SV to be playable if we implement more opcodes for it.

ptitSeb:
This is an old copy, with an unofficial linux patch to 64bits (as it's initialy a 32bits app). But I think it's on Steam also, and probably on Epic Store too.

ksco:
Thanks, I've successfully run this game.

ptitSeb:
There are still some missing opcode, especially visible when you change camera angle... but it's already very promising

ksco:
Yeah, I can see a lot of missing opcodes with `BOX64_DYNAREC_LOG=1`. I grabbed a free demo of this game, so I don't have the Single Player option available. When I go into Instant Action (practice mode), the screen goes black after loading and pressing [Fire] to start (sound and HUD are fine).

631681651618_ pic

ksco:
Got GTA 3 running! It's only 10fps or so, but still, feels great to see it running. There's something wrong with libmpg123 wrapper, I'll try to fix it tomorrow.
641681672526_ pic

ptitSeb:
nice :)

PR #685, opened 2023-04-11
[RV64_DYNAREC] Fixed emit_sub8/16

ksco:
RISC-V does not have `SUBB` or `SUBH` instructions. When using `SUB` or `SUBW`, there are some scenarios where the result will not be sign-extended. It's a bit tricky and hard to spot.


ptitSeb:
launch with gdb and ctl-C it. Then use the signal function (findDBfromnativeadress and getpc, something like that) to go back to x64 address. You also need `BOX64_LOAD_ADDR=0x10000000` probably to have fixed address, and you need a DUMP for reference.

ksco:
Thanks, I'll try that.

ksco:
Oh, this PR can be merged if there is no other issue. I'll fix the stuck issue in another PR, but after adding more opcodes for today ;)

PR #688, opened 2023-04-11
[RV64_DYNAREC] Add more opcodes for SV

ksco:
Haven't tested with SV yet but I will do the test tomorrow.


ptitSeb:
Ok, I'll wait until tomorrow to merge then.

I also noticed that WorldOfGoo doesn't work anymore, I need to debug that.

ksco:
Ok, I've tested it, still works.

ptitSeb:
> Ok, I've tested it, still works.

Yeah, I just forgot to not use system SDL2 when I launch it. It launch, and is playable also.

ksco:
Oh, I mean I've done the regression with Stardew Valley, but still good news on World of Goo, I'll try it later for sure.

ptitSeb:
Yeah, I was't sure... Eiher way, I merged the PR ;)

PR #693, opened 2023-04-13
[RV64_DYNAREC] Added more opcodes for SV

xctan:


xctan:
I tried to implement FXSAVE and FXRSTOR by copying from ARM side, but SV fails at an early stage. I haven't figured out why this happens, so they are removed from this patch. Other opcodes have been tested by SV.

src/dynarec/rv64/dynarec_rv64_00.c
```diff
@@ -958,6 +970,15 @@ uintptr_t dynarec64_00(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             READFLAGS(X_ALL);
             PUSH1(xFlags);
             break;
+        case 0x9D:
+            INST_NAME("POPF");
+            SETFLAGS(X_ALL, SF_SET);
+            POP1(xFlags);
```
ptitSeb, line 976: Flags needs to be converted to "RV64 format", with F_OF->F_OF2.

src/dynarec/rv64/dynarec_rv64_0f.c
```diff
@@ -344,6 +344,20 @@ uintptr_t dynarec64_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             }
             break;
 
+        case 0x5A:
+            INST_NAME("CVTPS2PD Gx, Ex");
+            nextop = F8;
+            GETGX(x1);
+            GETEX(x2, 0);
+            s0 = fpu_get_scratch(dyn);
+            FLW(s0, wback, fixedaddress);
```
ptitSeb, line 353: both Read should be done at the same time, in case Gx==Ex, to avoid overwrite PS[1] with PD[0]

ptitSeb (changes requested):
Added 2 comments to resolve

PR #706, opened 2023-04-17

ksco:
GTA 3 runs almost entirely in Dynarec now, but the framerate is still less than 10, unfortunately.


ptitSeb:
The GPU can be a limiting factor also. re can be quite graphic intensive, espcialy with post-process on

ksco:
Yeah, the GPU may not be powerful enough for this game, it's a great test target though.

PR #710, opened 2023-04-18
[RV64_DYNAREC] Added more opcode and some fixes

ksco:
test01 ~ test16 runs fully in Dynarec now.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_df.c
+++ b/src/dynarec/rv64/dynarec_rv64_df.c
@@ -111,6 +111,22 @@ uintptr_t dynarec64_DF(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                         FCVTDL(v1, x1, RD_RNE);
                     }
                     break;
+                case 1:
+                    INST_NAME("FISTTP Ew, ST0");
+                    v1 = x87_get_st(dyn, ninst, x1, x2, 0, EXT_CACHE_ST_F);
+                    addr = geted(dyn, addr, ninst, nextop, &wback, x3, x4, &fixedaddress, rex, NULL, 0, 0);
+                    FCVTWD(x4, v1, RD_RTZ);
+                    SH(x4, wback, fixedaddress);
+                    x87_do_pop(dyn, ninst, x3);
+                    break;
+                case 3: // TODO: fix rounding
+                    INST_NAME("FISTP Ew, ST0");
+                    v1 = x87_get_st(dyn, ninst, x1, x2, 0, EXT_CACHE_ST_F);
+                    addr = geted(dyn, addr, ninst, nextop, &wback, x3, x4, &fixedaddress, rex, NULL, 0, 0);
+                    FCVTWD(x4, v1, RD_RTZ);
+                    SH(x4, wback, fixedaddress);
+                    x87_do_pop(dyn, ninst, x3);
+                    break;
                 default:
                     DEFAULT;
                     break;
--- a/src/dynarec/rv64/dynarec_rv64_df.c
+++ b/src/dynarec/rv64/dynarec_rv64_df.c
@@ -115,15 +115,35 @@ uintptr_t dynarec64_DF(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                     INST_NAME("FISTTP Ew, ST0");
                     v1 = x87_get_st(dyn, ninst, x1, x2, 0, EXT_CACHE_ST_F);
                     addr = geted(dyn, addr, ninst, nextop, &wback, x3, x4, &fixedaddress, rex, NULL, 0, 0);
+                    if(!box64_dynarec_fastround) {
+                        FSFLAGSI(xZR); // reset all bits
+                    }
                     FCVTWD(x4, v1, RD_RTZ);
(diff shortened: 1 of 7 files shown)


src/dynarec/rv64/dynarec_rv64_helper.c
```diff
@@ -994,18 +994,19 @@ int x87_setround(dynarec_rv64_t* dyn, int ninst, int s1, int s2, int s3)
     // MMX/x87 Round mode: 0..3: Nearest, Down, Up, Chop
     // RV64: 0..7: Nearest, Toward Zero (Chop), Down, Up, Nearest tie to Max, invalid, invalid, dynamic (invalid here)
     // 0->0, 1->2, 2->3, 3->1
-    SLLI(s1, s1, 1);
```
ksco, line 997: I can't figure out what the original version is trying to do, and it's problematic, so I replaced it with a more trivial version (which has three direct jumps).

ptitSeb: Well, it's trying to change from 0, 1, 2, 3 to 0, 2, 3, 1 as the last comment says

src/dynarec/rv64/dynarec_rv64_df.c
```diff
@@ -111,10 +111,26 @@ uintptr_t dynarec64_DF(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                         FCVTDL(v1, x1, RD_RNE);
                     }
                     break;
+                case 1:
+                    INST_NAME("FISTTP Ew, ST0");
+                    v1 = x87_get_st(dyn, ninst, x1, x2, 0, EXT_CACHE_ST_F);
+                    addr = geted(dyn, addr, ninst, nextop, &wback, x3, x4, &fixedaddress, rex, NULL, 0, 0);
+                    FCVTWD(x4, v1, RD_RTZ);
```
ptitSeb, line 118: don't we need a check for Invalid Operation to change value to `0x8000` here?

src/dynarec/rv64/dynarec_rv64_df.c
```diff
@@ -111,10 +111,26 @@ uintptr_t dynarec64_DF(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+                    SH(x4, wback, fixedaddress);
+                    x87_do_pop(dyn, ninst, x3);
+                    break;
+                case 3: // TODO: fix rounding
+                    INST_NAME("FISTP Ew, ST0");
+                    v1 = x87_get_st(dyn, ninst, x1, x2, 0, EXT_CACHE_ST_F);
+                    addr = geted(dyn, addr, ninst, nextop, &wback, x3, x4, &fixedaddress, rex, NULL, 0, 0);
+                    FCVTWD(x4, v1, RD_RTZ);
```
ptitSeb, line 126: Same

src/dynarec/rv64/dynarec_rv64_df.c
ksco: Oh, the interpreter seems not doing it on this one, I'll add that on both sides.

ptitSeb: Ah, it's probably missing there too them. I would expect all conversion to int to react the same way on invalid operations

ksco: Emmm, can I just use `FR_NV` on this? It will need a lot of opcodes if not.

ptitSeb: Start with that, and encased the code in a `if(!box64_dynarec_fastround)` yes.

ksco: I've pushed a new commit, but since this is a `double -> int16` conversion, but we only have `FCVTWD` at hand, it's not accurate, right?

ptitSeb: Yep, there should be another test when doing 32bits -> 16bits reduction, to check for overflow there too

src/dynarec/rv64/dynarec_rv64_df.c
```diff
@@ -123,6 +123,8 @@ uintptr_t dynarec64_DF(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                         FRFLAGS(x5);   // get back FPSR to check the IOC bit
                         ANDI(x5, x5, 1<<FR_NV);
                         CBZ_NEXT(x5);
```
ptitSeb, line 125: CBZ_NEXT? So the next test will never be done

src/dynarec/rv64/dynarec_rv64_df.c
```diff
@@ -123,6 +123,8 @@ uintptr_t dynarec64_DF(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+                        SRLIW(x5, x4, 16);
```
ptitSeb, line 126: You are sure this will work for negative values. I think it will never be 0 because the high part will be filled with 0xffffffff after the `FCVTWD`?

ksco: Sorry, I've force-pushed, I believe it's right this time.

PR #712, opened 2023-04-19

xctan:
64 33 `XOR Gd, Seg:Ed` is causing black screen in ut2004 and Crypt of the NecroDancer (but sound works), so I commented that section. @ksco and I have no idea about this problem, but this instruction is heavily used in stack protection code and is a bottleneck. @ptitSeb can you look at this?


ptitSeb:
Ok, I'll merge then debug the commented code on my side.

ksco:
Stardew Valley also fails to start the latest main, I'm bisecting..

ptitSeb:
Shall I wait before merge?

PR #715, opened 2023-04-19
[RV64_DYNAREC] Fixed 66 0F 3A 0B ROUNDSD opcode

ksco:
Still not enough for Stardew Valley to run again, unfortunately.


ptitSeb:
I didn't noticed before, but round is not as simple as converting to integer and back to double.
There are test to be done (for nan/infinity) so the value is no lost, and case for very large value.

It's really unfortunate that there is no `FRINT` opcode in RV64.
A reference method opf conversion can be fond there: https://github.com/riscvarchive/riscv-glibc/blob/riscv-next/sysdeps/riscv/rv64/rvd/s_rint.c
More than 2 opcodes and a few jmp it seems...

ksco:
@xctan Can you fix this? Or we just comment it out for now? I need to get Stardew Valley work first..

xctan:
Ok, let's get Stardew Valley work first, just comment it.

PR #717, opened 2023-04-20

ksco:
Added for Human Resource Machine.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_0f.c
+++ b/src/dynarec/rv64/dynarec_rv64_0f.c
@@ -151,6 +151,16 @@ uintptr_t dynarec64_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 FMVD(v1, v0);
             }
             break;
+        case 0x13:
+            INST_NAME("MOVLPS Ex,Gx");
+            nextop = F8;
+            GETGX(x1);
+            GETEX(x2, 0);
+            LD(x3, gback, 0);
+            SD(x3, wback, fixedaddress+0);
+            if(!MODREG)
+                SMWRITE2();
+            break;
         case 0x14:
             INST_NAME("UNPCKLPS Gx,Ex");
             nextop = F8;
@@ -191,8 +201,8 @@ uintptr_t dynarec64_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             SD(x4, gback, 8);
             break;
         case 0x17:
-            nextop = F8;
             INST_NAME("MOVHPS Ex,Gx");
+            nextop = F8;
             GETGX(x1);
             GETEX(x2, 0);
             LD(x4, gback, 8);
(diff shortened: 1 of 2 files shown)


src/dynarec/rv64/dynarec_rv64_0f.c
```diff
@@ -151,6 +151,16 @@ uintptr_t dynarec64_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 FMVD(v1, v0);
             }
             break;
+        case 0x13:
```
ksco, line 154: @ptitSeb Just a question to ask, I originally implemented this opcode as:
```
case 0x13:
    INST_NAME("MOVLPS Ex,Gx");
    nextop = F8;
    GETGXSD(v0);
    GETEXSD(v1, 0);
    if (v0!=v1) FMVD(v1, v0);
    if(!MODREG) SMWRITE2();
    break;
```

But the game cannot run correctly, do you know what's wrong with this implementation?

ptitSeb:
Well, when a missing opcode is part of a main loop, it hurts the performances badly yes.
HumanRssourceMachine is a good game :) It's follow up "7 Billion Humans" is also good.

src/dynarec/rv64/dynarec_rv64_0f.c
ptitSeb: I don't don't see anything wrong here. not sure the use of "SD" variant is a good thing for a "PS" opcode tho.

ksco: Hmm, it's strange indeed, if we use this implementation, the little people on the landing page will turn blue..

ksco: Ok, I'll try that.

ksco: Oh, I think it is because `GETEXSD` cannot be used as a destination when ed is a memory address.

https://github.com/ptitSeb/box64/blob/84b0ee2284598a65ef7f727bb27b8c1527e8b5a3/src/dynarec/rv64/dynarec_rv64_helper.h#L301-L309

ptitSeb: Ah, well, yeah, there must be some "EXBACK" or something like that indeed!

ksco:
@ptitSeb `geted` function has an `i12` param, what does this mean?

ptitSeb:
when `i12` is `1`, it will fill `fixedaddress` with an `imm12` offset (if needed), so you need to use fixedaddress with Lxx/Sxx afterward. You can also put 2, 4, 8 or 16 value, wich mean you will do fixedvalue+2/4/8/16, so it's imm12 minus that value...

ksco:
So when should I use a non-zero value?

ptitSeb:
Most of the time.
I mean, each time you can use a `fixedaddress`  that is an `imm12` to read or write the data.

ksco:
Ok, that's clear, thanks.

PR #730, opened 2023-04-23

ksco:


ksco:
Search for `INST_NAME("` in rv64 folder gives 510 results, while in arm64 gives 1040 results, so we're roughly halfway there.

ksco:
Afk for a while, this can be reviewed.

PR #743, opened 2023-04-25

ksco:


ksco:
I did fixed the DIVPS opcode, but it seems just me messed up locally, sorry.

ptitSeb:
no, problem, it's just in case a commit was missing...

PR #745, opened 2023-04-26

ksco:


src/dynarec/rv64/dynarec_rv64_df.c
```diff
@@ -136,12 +136,12 @@ uintptr_t dynarec64_DF(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
                     v1 = x87_get_st(dyn, ninst, x1, x2, 0, EXT_CACHE_ST_F);
-                    addr = geted(dyn, addr, ninst, nextop, &wback, x3, x4, &fixedaddress, rex, NULL, 0, 0);
-                    u8 = sse_setround(dyn, ninst, x2, x3);
+                    u8 = sse_setround(dyn, ninst, x1, x2);
+                    addr = geted(dyn, addr, ninst, nextop, &wback, x2, x3, &fixedaddress, rex, NULL, 1, 0);
                     if(!box64_dynarec_fastround) {
                         FSFLAGSI(xZR); // reset all bits
                     }
-                    FCVTWD(x4, v1, RD_RM);
+                    FCVTWD(x4, v1, RD_RTZ);
```
ksco, line 144: This does fixed test22, but I don't why really. @ptitSeb can you take a look at this? Took me quite some time to find this..

ksco: Note: The newly added `FLD double` opcode allows this opcode to be executed in test22.

ksco: IIUC, double -> int16 conversion should follow the rounding mode, right?

ptitSeb: This is strange. I need to double check the behaviour on x86_64 for the extreme cases, but I would expect converting to 16bits to have the same rules as 32bits or 64bits.

ksco: Yeah, ok, thanks!

ksco: According to this [table](https://www.felixcloutier.com/x86/fist:fistp#tbl-3-27), the rounding mode is indeed not affected by rounding mode, it's fixed to RTZ.
And ARM side is using `VFCVTZSs/d` opcode for this one, this seems also an RTZ mode conversion.

ptitSeb: This page still mentio the "rounding mode". Also, if FIST(P) does RTZ, what is the use of FISTTP opcode?
I still need to check the 16bits variant of FIST(P) on the tests. Hopefully I'll have time to do that today (I expect ARM64 to not be correct also).

ksco: Ah right, seems partially depends on the rounding mode (only when `−1 < F < −0` or `+0<F<+1`)

ksco: Yes, and test22 is failing now, I'm tring to make it pass.

ksco:
Ok, I fixed it in above commit, it's a silly typo..

ksco:
So, `ctest` are fully* run on Dynarec, do you have any recommendations on where to find more tests? Still have ~500 instructions to be implemented.

*Some opcodes that are not implemented by ARM64 are skipped in RV64 also.

ptitSeb:
Well, now just pick your favorite games.

Note: all SSE4+ / AES / PMULL stuff are implemented? AES/PMULL are part of ctests.

ksco:
Ah, right, A small number of opcodes in test17 (sse_intrinsics) and test18 (aes) are skipped also, these're complicated..

ptitSeb:
Yes, but the AES and PMULL will not be emulated for now, there is no opcode for that on RV64 yet. So the C helper function will just be used (like on ARM64 when the extension are not present).

ptitSeb:
But anyway, if you prefer to focus on a specific game, that's also fine :)

PR #750, opened 2023-05-02
[GDBSTUB] Added preliminary GDBSTUB support

ksco:
It's far from usable, but I want to know if this patch is generally in good shape.

It is already possible to connect to the GDB client without attaching the binary, `info registers` and `stepi` directives are working.

Usage:
```
BOX64_DYNAREC=0 BOX64_GDBSTUB=1 BOX64_LOG=2 ./box64 ../tests/test01
```

```
gdb
(gdb) set architecture i386:x86-64
(gdb) target remote :1234
(gdb) i r
(gdb) si
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,6 +271,7 @@ set(ELFLOADER_SRC
     "${BOX64_ROOT}/src/emu/x64syscall.c"
     "${BOX64_ROOT}/src/emu/x64tls.c"
     "${BOX64_ROOT}/src/emu/x64trace.c"
+    "${BOX64_ROOT}/src/gdbstub/gdbstub.c"
     "${BOX64_ROOT}/src/librarian/librarian.c"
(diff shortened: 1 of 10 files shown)


ptitSeb:
This looks ok for now, but you need to change the build script, because for now, gdbstub depend on Dynarec for build (while it cannot be used with Dynarec).

ksco:
OK, I don't know much about CMake, I need to learn it.

ptitSeb:
I'll do it later if you struggle too much. Just tell me.

ksco:
Is this a correct fix? I'm just following the existing pattern, but it doesn't look like it depends on Dynarec anymore.

ptitSeb:
Yeah, looks good to me.

ptitSeb:
why closing? did it drift too much from master?

ksco:
> why closing? did it drift too much from master?

Yeah, I might not have the time to do it anytime soon, and it is not that urgent.

PR #753, opened 2023-05-03
[RV64_DYNAREC] Added more opcodes for Bastion and some fixes

ksco:
Still one more opcode (namely `67 42 88 34 00 mov byte ptr [eax + r8d], sil`) is needed by this game, and `geted32` is necessary for this one. 

@ptitSeb Could you implement it when you have the time?


ptitSeb:
I'll check geted32 and a 67 opcode ok.

PR #759, opened 2023-05-04
[RV64_DYNAREC] Added more opcode for Unciv and some fixes

ksco:
The game seems running fully in Dynarec now.


ksco:
Oh, I forgot to mention that the INSERTPS opcode is missing in ARM Dynarec also, we might want to implement it too :)

ptitSeb:
Oh, it's missing there? I should have an other equivalent INSERTxx already coded iirc.

PR #774, opened 2023-05-10
[RV64_DYNAREC] Added more opcodes and a small fix

ksco:
for Terraria.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_66.c
+++ b/src/dynarec/rv64/dynarec_rv64_66.c
@@ -83,8 +83,7 @@ uintptr_t dynarec64_66(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             INST_NAME("ADD AX, Iw");
             SETFLAGS(X_ALL, SF_SET_PENDING);
             i32 = F16;
-            SLLI(x1 , xRAX, 48);
-            SRLI(x1, x1, 48);
+            ZEXTH(x1 , xRAX);
             MOV32w(x2, i32);
             emit_add16(dyn, ninst, x1, x2, x3, x4, x6);
             LUI(x3, 0xffff0);
@@ -113,8 +112,7 @@ uintptr_t dynarec64_66(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             INST_NAME("OR AX, Iw");
             SETFLAGS(X_ALL, SF_SET_PENDING);
             i32 = F16;
-            SLLI(x1, xRAX, 48);
-            SRLI(x1, x1, 48);
+            ZEXTH(x1, xRAX);
             MOV32w(x2, i32);
             emit_or16(dyn, ninst, x1, x2, x3, x4);
             LUI(x3, 0xffff0);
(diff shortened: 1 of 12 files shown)


src/dynarec/rv64/dynarec_rv64_66.c
```diff
@@ -702,6 +702,26 @@ uintptr_t dynarec64_66(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+                        MOV32w(x4, 1);
+                    } else {
+                        INST_NAME("SHL Ew, CL");
+                        ANDI(x4, xRCX, 0x1f);
+                    }
+                    UFLAG_IF {MESSAGE(LOG_DUMP, "Need Optimization for flags\n");}
+                    SETFLAGS(X_ALL, SF_PENDING);
+                    GETEW(x1, 0);
+                    UFLAG_OP12(ed, x4)
+                    SLL(ed, ed, x4);
```
ptitSeb, line 718: you can use option `rv64_zbb` path here, with `zexth(...)`

ksco: indeed

ksco: Done, I searched and replaced all `SLLI + SRLI` pairs.

ptitSeb: Oh, wait, you have to make it optionnal. Zbb is not always supported. Use the `if(rv64_zbb)` Or change the `ZEXTH(...)` opcode itself to check for zbb and use SLLi/SRLI as a fallback.

ptitSeb:
Ah yes, nice! thanks.

src/dynarec/rv64/dynarec_rv64_660f.c
```diff
@@ -419,6 +418,41 @@ uintptr_t dynarec64_660F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+                    d1 = fpu_get_scratch(dyn);
+                    v1 = fpu_get_scratch(dyn);
+                    MOV64x(x3, 1ULL << __DBL_MANT_DIG__);
+                    FCVTDL(d1, x3, RD_RTZ);
+                    for (int i=0; i<2; ++i) {
+                        FLD(d0, wback, fixedaddress+8*i);
+                        FEQD(x4, d0, d0);
+                        BNEZ_MARK(x4);
+                        B_MARK3_nocond;
+                        MARK; // d0 is not nan
```
ksco, line 437: @ptitSeb  Just to be sure, is `MARK` macros going to work inside a for loop?

ptitSeb: Ah no, it will not. It will just get the MARK for the last iteration

ksco: It's computed and applied in different passes, right?

ptitSeb: Yes. It's computed on pass2 and applied on pass3.

ksco: Ok, I'll fix this in the upcoming PR.

PR #778, opened 2023-05-11
[RV64_DYNAREC] Added more opcodes and some fixes

ksco:
for Cassette Beasts.


ksco:
The newly added `0F` opcodes are not working correctly. I'm using `GETGX` and `GETEX`, but that's SSE infra. We might need some equivalents for MMX? Also, `mmx_get_reg[_empty]` doesn't seem to handle single precision float.

ptitSeb:
Yeap. There should be `GETEM` and `GETGM`. Should be similar to EX/GX, because it's also SIMD operation mainly

ksco:
The existing `mmx_*` functions do look similar to `sse_*` ones,  except it uses `mmxcache` instead of `ssecache`.

ksco:
I'll try to add them see if it works.

ptitSeb:
And note that there is only 8 mmx registers, not 16 like on sse

PR #780, opened 2023-05-11
[RV64_DYNAREC] Added more 0F MMX opcodes and some fixes

ksco:
for Cassette Beasts.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_0f.c
+++ b/src/dynarec/rv64/dynarec_rv64_0f.c
@@ -685,6 +685,22 @@ uintptr_t dynarec64_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
         case 0x71:
             nextop = F8;
             switch((nextop>>3)&7) {
+                case 2:
+                    INST_NAME("PSRLW Em, Ib");
+                    GETEM(x1, 1);
+                    u8 = F8;
+                    if (u8>15) {
+                        // just zero dest
+                        SD(xZR, x1, fixedaddress);
+                    } else if(u8) {
+                        for (int i=0; i<4; ++i) {
+                            // EX->uw[i] >>= u8;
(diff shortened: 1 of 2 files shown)


ksco:
I just noticed the mmx test is failing.

ksco:
The issue is caused by `MOVQ Gm, Em`, but I don't see anything wrong with it.

ptitSeb:
How do you know it's this one? You used cosim or you disabled the opcode?

ksco:
I disabled the opcode and it works fine. Also, the error message is [strange](https://github.com/ptitSeb/box64/actions/runs/4950119761/jobs/8853293227#step:6:124).

ptitSeb:
when you disable an opcode, you stop a block. So it works fine maybe because of another opcodes later in the block.

ptitSeb:
It's safer to do a "dump", using BOX64_LOAD_ADDR maybe to fix loaded address, than use BOX64_NODYNAREC=0xAAAA-0xBBBB to do a bisect on the faulty opcode

ksco:
Yeah, I'm aware of that. I looked at block dumps and not seeing anything suspicious yet. Thanks for the advice, I'll try to debug more tomorrow.

ksco:
I've found the issue. It's something wrong with `mmx_forget_reg`, the value of `dyn->e.mmxcache[a]` is `-1`.

ptitSeb:
Isn't some macro to transform index to/from reg needs to be used there also?

ksco:
Why do we need that? The 8 max regs are already 1to1 mapped, right?

ptitSeb:
I'm not sure I'll keep the reg mapping fixed as it is. It might change in the future and maybe all regs will be allocated in a more dynamix way (except, porbably, for xmm0-xmm8 that will stay at the same place, to keep it ABI compatible).
Box64 use 24 float regs for it's cache, and the mapping of those 24regs is not linear on rv64.

PR #785, opened 2023-05-15
[RV64_DYNAREC] Added some missing fixes on GETEX

ksco:


ksco:
Another fix above, I pushed to the current PR for convenience.

PR #802, opened 2023-05-24
[RV64_DYNAREC] Fixed 66 0F 38 2B PACKUSDW opcode

ksco:
Cached by cosim when running dav1d 1.2.0. I believe that dav1d did not work yesterday because of this reason, because I also ran cosim yesterday, and it also reported inconsistent results for this opcode. But I can't reproduce the problem that dav1d prints a bunch of error logs and then exits abnormally.


ptitSeb:
Ah, so maybe the optimisation for dav1d is not good anymore for latest version. The SAFEFLAGS is probably at fault and should be removed (the CALLRET also, but it's not active on RV64 anyway).

ksco:
Yeah, confirmed, it's caused by `SAFEFLAGS=0`, do you need me to remove these two options for dav1d?

ptitSeb:
Yeah, removing the option is fine.

PR #807, opened 2023-05-28
[WRAPPER] Added some wayland function wrappers

ksco:
Added for [Telegram](https://desktop.telegram.org/), but it still not working (segfaulted).

### DIFF
--- a/src/wrapped/generated/functions_list.txt
+++ b/src/wrapped/generated/functions_list.txt
@@ -1616,6 +1616,7 @@
 #() pFpuuuu
 #() pFpuuup
 #() pFpupii
+#() pFpuppu
 #() pFpuppp
 #() pFpUdii
 #() pFpfffi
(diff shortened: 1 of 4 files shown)


ksco:
Hmm, is there any way we can wrap the `wl_proxy_marshal_constructor` function from Wayland? It has the signature: 
```
struct wl_proxy * wl_proxy_marshal_constructor(struct wl_proxy *proxy, uint32_t opcode, const struct wl_interface *interface,...)
```

The vaargs parameter is for "extra arguments for the given request". Do we have to record the "given request" to know the vaargs' layout? It seems complicated.

ptitSeb:
IIRC, this function also serve as a brdige for many callback function right?

But if there isn't any `_va` equivalent function, it will be complicated yes, because `wrapping ...` cannot be generic.

ksco:
I don't know Wayland/OpenGL stuff. But yes, there is no `_va` equivalent, unfortunately.

ptitSeb:
try to run your stuff with an emulated version of this lib? Just t see if it worth focussing on this one.

ksco:
Oh, right, almost forgot we have that option!

ksco:
By the way, I see most of the use cases of Box64 are running games, do you know any GUI apps that work with Box64 ARM64 already? I can test these for the RV64 too.

ptitSeb:
I know Heroic works on ARM64. It's a 64bits app that can handle GOG and Epic store. But it's an electron app, so it might not run fine yet on rv64.
Lutris should also works. It's a Python app so that should just work.

Most of the launcher will use Wine to run games, so that may be difficult again to run on rv64.

ksco:
Ok, thanks!

PR #808, opened 2023-05-28
[RV64_DYNAREC] Added F3 0F B8 POPCNT opcode and fixed LOCK CMPXCHG16B for ML2

xctan:


src/dynarec/rv64/dynarec_rv64_f0.c
```diff
@@ -283,9 +283,16 @@ uintptr_t dynarec64_F0(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
-                                // there is no atomic move on 16bytes, so faking it
+                                // there is no atomic move on 16bytes, so implement it with mutex
+                                LD(x9, xEmu, offsetof(x64emu_t, context));
+                                ADDI(x9, x9, offsetof(box64context_t, mutex_16b));
+                                ADDI(x4, xZR, 1);
+                                MARKLOCK;
+                                AMOSWAP_W(x4, x4, x9, 1, 1);
+                                // x4 == 1 if locked
+                                BNEZ_MARKLOCK(x4);
+
```
ptitSeb, line 294: I had something more drastic in mind: directly use the address of `context->mutex_16b` instead of fetching it, to avoid a memory read (because it's a global immutable constant), and also change ALL the LOCK prefixed opcode to use that mutex (to really simulate the LOCK prefix), else, you can miss a CMPXHG16B and a 8bytes write at the same address (for example).
Also, the Interpreter needs to also be change to use the mutex approch, but with the custom mutex.

That's a lot of changes, so it can be done is a different commit anyway (and I can do it later if needed).

ksco: Maybe we can further use an array of locks and a hash function to reduce the penalty of this approach..

ptitSeb: No no no, an array of lock would defait the pupose of the mutex. a LOCK prefix is supposed to be costly in term of perfomances. LOCK actually lock the whole memory for this command, all other core will have to wait...

The advantage of using the mutex approch is that it will remove the alignment constraint. The cost penality should be reasonable, as there was also some waiting before.

The issue with the mutex approach is that it is incomplete. All other regular write will not chack for the lock (for speed purpose), so a regular write on the 16bytes being swapped will not invalidate the current swap.

There isn't any really good solution here, without a 16bytes atomic operation.

ksco: Ok, looks like overall it's a better solution than what we have now.

xctan: Maybe it's better to add another flag to switch to full mutex mode in riscv64 dynarec? 16 bytes atomic operation is rarely used outside wine.

PR #815, opened 2023-06-01
[RV64_DYNAREC] Added F0 80 /1 LOCK OR opcode for ICEY

xctan:


xctan:
Looks like ICEY (in wine) now gets stuck in a dead loop just before the game logo appears. It uses some unimplemented opcodes though, so maybe we can better figure out what happens with these opcodes implemented.

cc @ksco

xctan:
FYI wine detection in POPF breaks ICEY in the early stage. I commented that section to get the game to start up.

ptitSeb:
How is that? It should not break anything. If TF is set, it is expexted to do a TRACE signal. Maybe signal handling is not 100% correct on RV64?

PR #822, opened 2023-06-06
[DYNAREC] Fixed native_fprem

ksco:


ptitSeb:
Oh, I inverted C0 and C1?! Nice find (I need to fix box86 probably too)

ptitSeb:
box86 was good already..

PR #823, opened 2023-06-06
[DYNAREC] Fixed call_c issues

ksco:
1. We cannot use xRAX as a return value register when using `call_c`, it will be "restored" after the c function returns.
2. `div64` and `div32` sets `xRDX`, so even if `xRDX` is a callee-saved register, we still need to save & restore it.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_00_3.c
+++ b/src/dynarec/rv64/dynarec_rv64_00_3.c
@@ -899,15 +899,14 @@ uintptr_t dynarec64_00_3(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int 
                         AND(xRAX, x2, xMASK);
                         ZEROUP(xRDX);
                     } else {
-                        if(ninst 
-                           && dyn->insts[ninst-1].x64.addr 
-                           && *(uint8_t*)(dyn->insts[ninst-1].x64.addr)==0x31 
+                        if(ninst
+                           && dyn->insts[ninst-1].x64.addr
+                           && *(uint8_t*)(dyn->insts[ninst-1].x64.addr)==0x31
                            && *(uint8_t*)(dyn->insts[ninst-1].x64.addr+1)==0xD2) {
                             SET_DFNONE();
                             GETED(0);
-                            DIVU(x2, xRAX, ed);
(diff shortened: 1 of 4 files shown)


src/dynarec/rv64/dynarec_rv64_00_3.c
```diff
@@ -899,25 +899,23 @@ uintptr_t dynarec64_00_3(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
-                           && dyn->insts[ninst-1].x64.addr 
-                           && *(uint8_t*)(dyn->insts[ninst-1].x64.addr)==0x31 
+                        if(ninst
+                           && dyn->insts[ninst-1].x64.addr
+                           && *(uint8_t*)(dyn->insts[ninst-1].x64.addr)==0x31
                            && *(uint8_t*)(dyn->insts[ninst-1].x64.addr+1)==0xD2) {
                             SET_DFNONE();
                             GETED(0);
-                            DIVU(x2, xRAX, ed);
                             REMU(xRDX, xRAX, ed);
```
ptitSeb, line 908: but what if ed is RDX?

ksco: Hmmm, indeed, I never thought about this, sorry.

ksco: > but what if ed is RDX (this is why the code was like that)

We can use `if (ed == xRDX)` to save 1 instruction when ed is not xRDX, but I don't think it's worth it, I'll just revert this change.

ksco:
> I think the restore of RAX and RDX in the case of function call is a bad idea, as they are the return registers. I'm fine with the current workaround, but it should be note that we have an small optimisation oportunity here.

But we need to sync RAX and RDX from xEmu to Dynarec after the call, is there a better way to do it?

ptitSeb:
> > I think the restore of RAX and RDX in the case of function call is a bad idea, as they are the return registers. I'm fine with the current workaround, but it should be note that we have an small optimisation oportunity here.
> 
> But we need to sync RAX and RDX from xEmu to Dynarec after the call, is there a better way to do it?

I'm still unsure. Will need to study some dump to see how (in)efficient calls are.

PR #831, opened 2023-06-09
[WRAPPER] Added more libicu wrapped functions (for #829)

ksco:
for #829


ksco:
Huh, I forgot to rebase first, sad story..

PR #858, opened 2023-06-24
A few cosmetic fixes

julliard:
A few cosmetic fixes for things spotted while trying to do a PE build under Wine. These should all be no-ops.

### DIFF
--- a/src/build_info.h
+++ b/src/build_info.h
@@ -1,6 +1,6 @@
 #ifndef __BUILD_INFO_H__
 #define __BUILD_INFO_H__
 
-void PrintBox64Version();
+void PrintBox64Version(void);
 
 #endif //__BUILD_INFO_H__
--- a/src/dynarec/arm64/dynarec_arm64_00.c
+++ b/src/dynarec/arm64/dynarec_arm64_00.c
@@ -1,7 +1,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <stddef.h>
-#include <pthread.h>
 #include <errno.h>
 #include <signal.h>
 
--- a/src/dynarec/arm64/dynarec_arm64_0f.c
+++ b/src/dynarec/arm64/dynarec_arm64_0f.c
@@ -1,7 +1,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <stddef.h>
-#include <pthread.h>
 #include <errno.h>
 
 #include "debug.h"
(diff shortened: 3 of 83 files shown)


ptitSeb:
a PE build ... of box64? are you plugin box64 in wine peloader or something like that? removing elfloader and using something PE oriented instead?

julliard:
What I'm trying to do is basically use only the emulator parts, and plug it into the existing wow64 support, so syscalls go to Wow64SystemServiceEx which takes care of 32->64 transitions, and Wine handles everything else on the 64-bit side.

ptitSeb:
Nice.

Signals are also probably a bit tricky.

theofficialgman:
> What I'm trying to do is basically use only the emulator parts, and plug it into the existing wow64 support, so syscalls go to Wow64SystemServiceEx which takes care of 32->64 transitions, and Wine handles everything else on the 64-bit side.

by that do you mean you trying to build wine on ARM64 with `--enable-archs=i386,x86_64.aarch64` so that the wine PE userspace is linux aarch64 (eg: wine lib folder looks like `aarch64-unix  aarch64-windows  i386-windows  x86_64-windows`) and you are using box64 for any/all x86/x64 binary recompilation only? I couldn't get that to work with wine/box64 as it is. wine would never call box64... it juser errored whenever it met x86/x64 code

julliard:
> Signals are also probably a bit tricky too.

All we really need from the emulator is to fetch the current context, all signal handling is done on the Wine side. When the exception needs to be handled in 32-bit we reenter emulation through KiUserExceptionDispatcher. It seems to work OK in emulation mode, but I haven't tried dynarec yet.

julliard:
> by that do you mean you trying to build wine on ARM64 with `--enable-archs=i386,x86_64.aarch64` so that the wine PE userspace is linux aarch64 (eg: wine lib folder looks like `aarch64-unix aarch64-windows i386-windows x86_64-windows`) and you are using box64 for any/all x86/x64 binary recompilation only? I couldn't get that to work with wine/box64 as it is. wine would never call box64... it juser errored whenever it met x86/x64 code

Yes, the Unix interfaces and ntdll are aarch64, the rest is x86 PE code running through emulation.

For x64 the plan is to implement ARM64EC and have basically all of Wine run as native ARM code, but the ARM64EC toolchain is not ready yet.

Heasterian:
Small question from me. Is it going to work with wineandroid.drv?

julliard:
> Small question from me. Is it going to work with wineandroid.drv?

It should work fine, but wineandroid.drv is pretty much abandoned at the moment, someone will have to pick it up again. Maybe emulation support will create some new motivation to maintain it...

PR #866, opened 2023-06-26
[DYNAREC_RV64] Fixed A0 MOV AL,Ob

ksco:


ksco:
I fixed rv64_move32 in https://github.com/ptitSeb/box64/pull/866/commits/9861c4f2b899edf5c1cbb1ef71be1f910af99d93 above! On a very rare condition, the original implementation could be wrong: like `MOV64x(x1, 0x7fffffff0000)`.

Original:
```
        lui     a0, 524288 // 0x80000
        addi    a0, a0, -1
        slli    a0, a0, 16
        ret
```

Fixed:
```
        lui     a0, 524288 // 0x80000
        addiw   a0, a0, -1
        slli    a0, a0, 16
        ret
```

PR #878, opened 2023-07-03
A few Push/Pop fixes and cleanups

julliard:


ptitSeb:
Oh, I meesed up PUSHA/POPA it seems!
Thanks for the PUSH -> PUSH64, it's cleaner indeed.

PR #922, opened 2023-08-04
[ARM64_DYNAREC] Fixed BT/BTC/BTR/BTS opcodes

wannacu:


src/dynarec/arm64/dynarec_arm64_660f.c
```diff
@@ -1809,7 +1810,7 @@ uintptr_t dynarec64_660F(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int n
             B_NEXT(cNE);
             MOV32w(x1, 1);
             LSLxw_REG(x1, x1, x2);
-            EORx_REG(ed, ed, x1);
+            ORRx_REG(ed, ed, x1);
```
ptitSeb, line 1813: because there is a ` B_NEXT(cNE)`  before, in here the we know the bit is 0, so EOR is the same as ORR (but ORR is more clear)

src/dynarec/arm64/dynarec_arm64_660f.c
```diff
@@ -1872,7 +1873,7 @@ uintptr_t dynarec64_660F(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int n
             B_NEXT(cEQ);
             MOV32w(x1, 1);
             LSLxw_REG(x1, x1, x2);
-            EORx_REG(ed, ed, x1);
+            BICx_REG(ed, ed, x1);
```
ptitSeb, line 1876: Same as bifore, we know th bit is 1 here, so EOR is same as BIC

PR #936, opened 2023-08-17
Added some opcodes

wannacu:


src/dynarec/rv64/dynarec_rv64_66f0.c
```diff
@@ -48,6 +48,29 @@ uintptr_t dynarec64_66F0(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+                ed = xRAX+(nextop&7)+(rex.b<<3);
+                MOV32w(x6, 0xFFFF);
+                AND(x2, ed, x6);
+                emit_adc16(dyn, ninst, x1, x2, x3, x4, x5);
+                NOT(x6, x6);
+                AND(ed, ed, x6);
+                OR(ed, ed, x2);
+            } else {
+                addr = geted(dyn, addr, ninst, nextop, &wback, x2, x3, &fixedaddress, rex, LOCK_LOCK, 0, 0);
+                LHU(x3, wback, 0);
```
ptitSeb, line 68: Where is th elock mecanism? no lock for 16bits for now?

src/dynarec/rv64/dynarec_rv64_f0.c
```diff
@@ -339,6 +339,59 @@ uintptr_t dynarec64_F0(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+                MOV_U12(x3, 0xFF);
+                SLLI(x3, x3, wb2*8);
+                NOT(x3, x3);
+                AND(wback, wback, x3);
+                OR(wback, wback, x1);
+            } else {
+                addr = geted(dyn, addr, ninst, nextop, &wback, x1, x3, &fixedaddress, rex, LOCK_LOCK, 0, 0);
+                LBU(x3, wback, 0);
+                emit_adc8(dyn, ninst, x3, x2, x4, x5, x6);
+                SB(x3, wback, 0);
```
ptitSeb, line 371: No lock mecanism for 8bits either?

src/dynarec/rv64/dynarec_rv64_f0.c
```diff
@@ -339,6 +339,59 @@ uintptr_t dynarec64_F0(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+            GETGD;
+            SMDMB();
+            if(MODREG) {
+                ed = xRAX+(nextop&7)+(rex.b<<3);
+                emit_adc32(dyn, ninst, rex, ed, gd, x3, x4, x5, x6);
+            } else {
+                addr = geted(dyn, addr, ninst, nextop, &wback, x2, x1, &fixedaddress, rex, LOCK_LOCK, 0, 0);
+                MARKLOCK;
+                LRxw(x1, wback, 1, 1);
+                emit_adc32(dyn, ninst, rex, x1, gd, x3, x4, x5, x6);
```
ptitSeb, line 389: I don't think this will work if there are some flags computations. There will be too many instruction between the read and the write.
You should do the ADC inline between the read/write (with precomputing the Carry before the read) and preserving the initial regs if needed, and than call adc32 if there are some flags needed outside the read/write sequence.

ptitSeb: A better way might be to use the atomic add, with pre-adding the Carry to GD before.

src/dynarec/rv64/dynarec_rv64_66.c
```diff
@@ -117,6 +117,37 @@ uintptr_t dynarec64_66(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+            break;
+        case 0x15:
+            INST_NAME("ADC AX, Iw");
+            READFLAGS(X_CF);
+            SETFLAGS(X_ALL, SF_SET_PENDING);
+            i32 = F16;
+            SRLI(x6, xMASK, 16);
+            AND(x1, xRAX, x6);
+            MOV32w(x2, i32);
+            emit_adc16(dyn, ninst, x1, x2, x3, x4, x5);
```
ptitSeb, line 148: Something is wrong here.

x1 = upper part of RAX
but you adc x1 with i32 (that should be the lower part of RAX instead).

Also, you will loose the upper 32bits of RAX anyway with this, wich is also wrong.
You need to fix your masking (and you'll have to use 1 more opcde)
like x1 = xRAX AND mask
adc x1, x2
xRAX = xRAX NAND mask
xRAX = xRAX or x1

src/dynarec/rv64/dynarec_rv64_f0.c
```diff
@@ -339,6 +339,101 @@ uintptr_t dynarec64_F0(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+                addr = geted(dyn, addr, ninst, nextop, &wback, x1, x3, &fixedaddress, rex, LOCK_LOCK, 0, 0);
+                ANDI(x3, wback, 0b11);
+                MOV_U12(x4, 0xff);
+                BNEZ_MARK(x3);
+                MARKLOCK;
+                LR_W(x5, wback, 1, 1);
+                AND(x4, x5, x4);
+                ANDI(x5, x5, -256);
+                ADDW(x6, x4, x2);
+                ANDI(x9, xFlags, 1 << F_CF);
```
ptitSeb, line 377: Is would be more optimal to to pre-add the Carry flags to GD a,d so move
```
ANDI(x9, xFlags, 1 << F_CF);
ADDW(x6, x2, x9);
```
Out of the loops.
Thoses needs some comment to be sure what regs old what value, to ease the debugging.
Like, line 374 x4 that was holding 0xff is destroy and now have [wback]&x4 inside the loop, that seems suspicious.

ptitSeb (changes requested):
Sorry, more changes requested.

src/dynarec/rv64/dynarec_rv64_00_0.c
```diff
@@ -178,7 +187,42 @@ uintptr_t dynarec64_00_0(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+            emit_adc32(dyn, ninst, rex, gd, ed, x3, x4, x5, x6);
+            break;
+        case 0x14:
+            INST_NAME("ADC AL, Ib");
+            READFLAGS(X_CF);
+            SETFLAGS(X_ALL, SF_SET_PENDING);
+            u8 = F8;
+            ANDI(x1, xRAX, 0xff);
+            emit_adc8c(dyn, ninst, x1, u8, x3, x4, x5, x6);
+            ANDI(xRAX, x1, 0xff);
```
ptitSeb, line 216: This is wrong. only AL (so lower 8bits) should changed, the rest of RAX should be left alone.
so it should be
BICI(xRAX, xRAX, 0xff) or something equivalent (I don't rememeber what's availble)
ORI(xRAX, xRAX, x1)

src/dynarec/rv64/dynarec_rv64_66f0.c
```diff
@@ -48,6 +49,67 @@ uintptr_t dynarec64_66F0(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+            nextop = F8;
+            GETGW(x1);
+            SMDMB();
+            if(MODREG) {
+                ed = xRAX+(nextop&7)+(rex.b<<3);
+                MOV32w(x6, 0xFFFF);
+                AND(x2, ed, x6);
+                emit_adc16(dyn, ninst, x1, x2, x3, x4, x5);
+                NOT(x6, x6);
+                AND(ed, ed, x6);
```
ptitSeb, line 65: an option ANDN use path could be nice

ptitSeb:
Thanks, that's a big one.

PR #941, opened 2023-08-21
[ARM64_DYNAREC] Added 0F F1/F3/F7 opcodes

wannacu:


src/dynarec/arm64/dynarec_arm64_0f.c
```diff
@@ -1975,7 +1975,16 @@ uintptr_t dynarec64_0F(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
...
             break;
-
+        case 0xF1:
+            INST_NAME("PSLLW Gm,Em");
+            nextop = F8;
+            GETGM(d0);
+            GETEM(d1, 0);
+            v0 = fpu_get_scratch(dyn);
+            VMOVHto(x1, d1, 0);
+            VDUPH(v0, x1);
```
ptitSeb, line 1985: Not sure tranfering to/from NEON regs is faster then duplicating the regs, but that's simpler.

PR #948, opened 2023-08-23
Allow running at NULL, and trigger a segfault (backport from box86 5cdc1b9fa2238938e49176fcd937a1294c55a03d )

rafael2k:
This is just a start, as I "almost" managed to run VARA.exe on wine/hangover with box64 in WoW64, but not there yet. But with this change (same I need to run VARA.exe on box86) I got a bit ahead in the tests.

The debug info:
http://abradig.org.br/hermes/debug.txt

### DIFF
--- a/src/emu/x64run.c
+++ b/src/emu/x64run.c
@@ -57,10 +57,8 @@ int Run(x64emu_t *emu, int step)
     if(emu->quit)
         return 0;
     if(addr==0) {
-        emu->quit = 1;
-        printf_log(LOG_INFO, "%04d|Ask to run at NULL, quit silently\n", GetTID());
-        print_cycle_log(LOG_INFO);
-        return 0;
+        // Some programs, like VB6 VARA.exe, need to trigger that segfault to actually run... (ticket #830 in box86)
+        printf_log(LOG_INFO, "%04d|Ask to run at NULL, will segfault\n", GetTID());
     }
     //ref opcode: http://ref.x64asm.net/geek32.html#xA1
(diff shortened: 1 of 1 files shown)


rafael2k:
Thanks. I'm pretty sure the issue I'm facing is related to WoW64 on wine.

ptitSeb:
That can be yeah. Wow64 support is pretty young in box64, there are some bugs left...

rafael2k:
Can you point me which env variables I should set to get as much verbosity as possible for debugging box64?
Anyway, I think I need to focus on debugging wine...

ptitSeb:
For debugging, you need at least `BOX64_LOG=1 BOX64_SHOWSEGV=1 BOX64_SHOWBT=1` that gives a good baseline of trace.
For testing, you should try with `BOX64_DYNAREC=0` to test without dynarec, just in case. You might want to use `~/.box64rc` and create a profile for your executable to avoid running the whole wine without dynarec.

Also. when looking at a segfault, check `CS`. If it's `0x0033` it's 64bits code, if it's `0x0023` it's 32bits (and if something else, that probably very wrong)

ptitSeb:
More tips:

What I usualy do is use a "TRACE" enabled build and do a dump of the program (using a profile if it's a windows stuff), with `BOX64_TRACE=0-1 BOX64_TRACE_FILE=dump.txt BOX64_DYNAREC_DUMP=2` and then look at the segfault and to the corresponding emitted code. Note that if you don't use a trace enabled box64, the dump x86_64 code will be "generic" and not the exact code, making it a bit more difficult to follow.
When debug linux code I use somethinbg like `BOX64_LOAD_ADDR=0x10000000` to have reproductible run (load at the same address), but on wine this as little effect because wine is doing the loading. If the issue is in the main executable, the load address is often fixed, so you can use ghidra or IDA or anything to look at the original code and have a broader view of it is going on...

PR #951, opened 2023-08-25
Added (66) 0F 2A/2C/2D opcodes

wannacu:


ptitSeb (changes requested):
The fastround is not handled properly for ARM64: those are Packed Single conversion: so it's 4 (or 2) float value beeing converted. If the IOC is set, you don't know which of the 4 value is overflowing and the only solution is to unrool the SIMD ocode to do it element by element to find the one(s) that overflow. Here you just consider it's the 1st element.

ptitSeb:
Thanks, looks better now.

PR #964, opened 2023-09-04
[ARM64_DYNAREC] Added some opcodes

wannacu:


src/dynarec/arm64/dynarec_arm64_0f.c
```diff
@@ -421,9 +421,12 @@ uintptr_t dynarec64_0F(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
                     GETGM(q0);
                     GETEM(q1, 0);
                     d0 = fpu_get_scratch(dyn);
-                    MOVI_8(d0, 0b10001111);
+                    d1 = fpu_get_scratch(dyn);
+                    MOVI_8(d0, 0b111);
                     VAND(d0, d0, q1);  // mask the index
+                    VSSHR_8(d1, q1, 7); // zero mask
                     VTBL1_8(q0, q0, d0);
+                    VBIC(q0, q0, d1);
```
ptitSeb, line 429: Why are you changing this one? Was it wrong before?

src/dynarec/arm64/dynarec_arm64_660f.c
```diff
@@ -308,9 +308,12 @@ uintptr_t dynarec64_660F(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int n
                     GETGX(q0, 1);
                     GETEX(q1, 0, 0);
                     d0 = fpu_get_scratch(dyn);
-                    MOVIQ_8(d0, 0b10001111);
+                    d1 = fpu_get_scratch(dyn);
+                    MOVIQ_8(d0, 0b1111);
                     VANDQ(d0, d0, q1);  // mask the index
                     VTBLQ1_8(q0, q0, d0);
+                    VSSHRQ_8(d1, q1, 7); // zero mask
+                    VBICQ(q0, q0, d1);
```
ptitSeb, line 316: And why changing this one too?

VTBL1 is supposed to set element to 0 if index is over the source size, and this is done by setting bit 7 of index (so idx>=128, so out of range).

src/dynarec/arm64/dynarec_arm64_0f.c
ptitSeb: I see.
So, I fixed the issue in the original code, and added your test in the test suite.
Place rebase you code with lastest code. You can remove your changes to both PSHUFB, the code is working correctly now (issue was only on the EMM PSHUFB, mask was too large).

ptitSeb:
Nice one with the use of VDUP instead on the MOVeS & D.

PR #977, opened 2023-09-13
[RV64_DYNAREC] Added some support for XTheadBb extension

ksco:


ksco:
Tested with Stardew Valley, which uses these changes, and works great.

PR #1002, opened 2023-10-02
[ANDROID] Fix Most Libraries Names and set PThread Symbols on libc an…

KreitinnSoftware:
…d add __assert2 and __sF


ptitSeb:
Why did you closed it?

Issue #939, opened 2023-08-20
Dwarf Fortress 0.50.x Linux beta SIGSEVs immediately upon launch


cocona20xx:
OS: Asahi Fedora Remix (pre-release)
Hardware: Apple Silicon M1 Mac Mini (8 cores, 16gb ram)
Log output upon crash: 
```text
[rin@2603-7000-da42-f100-32a5-f063-50d6-8706 dflinux]$ box64 dwarfort
Dynarec for ARM64, with extension: ASIMD AES CRC32 PMULL ATOMICS PageSize:16384 Running on Icestorm Firestorm with 8 Cores
Params database has 25 entries
Box64 with Dynarec v0.2.4 45c4117a built on Aug 20 2023 00:36:40
Using default BOX64_LD_LIBRARY_PATH: ./:lib/:lib64/:x86_64/:bin64/:libs64/
Using default BOX64_PATH: ./:bin/
Counted 77 Env var
Looking for dwarfort
Rename process to "dwarfort"
Using emulated libg_src_lib.so
Using native(wrapped) libSDL2-2.0.so.0
Using native(wrapped) libdl.so.2
Using native(wrapped) libc.so.6
Using native(wrapped) ld-linux-x86-64.so.2
Using native(wrapped) libpthread.so.0
Using native(wrapped) libutil.so.1
Using native(wrapped) librt.so.1
Using native(wrapped) libm.so.6
Using native(wrapped) libSDL2_image-2.0.so.0
Using emulated libsteam_api.so
Using emulated /lib/x86_64-linux-gnu/libstdc++.so.6
Using emulated /lib/x86_64-linux-gnu/libgcc_s.so.1
Using emulated ./libfmod_plugin.so
Using emulated libfmod.so.13
Loading audio...Audio loaded successfully!refreshProtection(0x37f88010): 0x37f88000/0x7 (ret=-1/Permission denied)
Strange SIGSEGV with Access error on 0x37f88010 for 0x37f88010, db=0x37f3efe0, prot=0x7 (old_addr=(nil))
refreshProtection(0x37f88010): 0x37f88000/0x7 (ret=-1/Permission denied)
refreshProtection(0x37f88010): 0x37f88000/0x7 (ret=-1/Permission denied)
Strange SIGSEGV with Access error on 0x37f88010 for 0x37f88010, db=0x37f3efe0, prot=0x7 (old_addr=0x37f88010)
[...]

PR #1008, opened 2023-10-03
Add c32rtomb wrapped function for dfhack (fixes console, part of issue #939)

seeschloss:
Looks like I had missed something else, that's what broke dfhack's console.

I checked the pointer argument of c32rtomb (`std::mbstate_t`) and it's already used in the wrapped functions `mbrlen` or `mbrtowc` as a "p" so it seems fine to me (and it does fix the console issue, so it's at last better than nothing).


ptitSeb:
yeah, looks good to me. Thanks!

PR #1009, opened 2023-10-04
Added Snapdragon 8 Gen 2 profile

strongtz:
Compilation works with GCC 12.1.0

Tested on AYN Odin 2 with mainline Linux 6.6.0-rc3-next-20230929


ptitSeb:
Maybe disable the profile for SD8Gen2 on the CI for now if it needs gcc 12, because only gcc-11 seems to be available on the ci for now.

ptitSeb:
Also also, ODIN 2 works with mainline linux??? :O

strongtz:
> Also also, ODIN 2 works with mainline linux already??? :O

Yes. I already have touchscreen, usb, battery, bluetooth, wifi, display, gpu working.

However, freedreno (gl driver) doesn't support a7xx at the moment, only vulkan (turnip) is available. So the only way to get accelerated desktop is through zink. But plasma desktop crashes with zink. As for GNOME, desktop works, but many applications including system settings are crashing.

ptitSeb:
Very good news about the odin. Maybe I'll get one if linux is supported

Saikatsaha1996:
> > Also also, ODIN 2 works with mainline linux already??? :O
> 
> Yes. I already have touchscreen, usb, battery, bluetooth, wifi, display, gpu working.
> 
> However, freedreno (gl driver) doesn't support a7xx at the moment, only vulkan (turnip) is available. So the only way to get accelerated desktop is through zink. But plasma desktop crashes with zink. As for GNOME, desktop works, but many applications including system settings are crashing.

Do you think to merge your sm8350 ?  Sm8350 update is ready for race?

PR #1025, opened 2023-10-13
[DYNAREC] Fix x87cache issues

ksco:
Reported by @fan-wenjie


ksco:
@ptitSeb Seems working, but I can't test real programs right now.

ptitSeb:
Looks ok. I would have created new macro for PUSH/PUSHEMPTY/POP for easier changes, but this one works too.

I can wait if you can test a bit later.

ksco:
> I can wait if you can test a bit later.

Are you able to test it for arm64? Like running some games (UT2004 maybe?) or WOW64 apps that use x87, I want to make sure it doesn't break things that are already working. I'll do some tests for rv64 later today.

> I would have created new macro for PUSH/PUSHEMPTY/POP

Yeah, this looks better, but we'll only need to create new macros for PUSH/PUSHEMPTY, POP empty is harder to detect as we assume the x87cache is empty at the entry of a block IIUC.

ptitSeb:
> 
> > I would have created new macro for PUSH/PUSHEMPTY/POP
> 
> Yeah, this looks better, but we'll only need to create new macros for PUSH/PUSHEMPTY, POP empty is harder to detect as we assume the x87cache is empty at the entry of a block IIUC.

Pop empty is the same: you need to test if the stack is -8 (because it can be negative, if you start a block and just pop po pop...)

src/dynarec/arm64/dynarec_arm64_helper.c
```diff
@@ -974,6 +974,16 @@ void x87_do_pop(dynarec_arm_t* dyn, int ninst, int s1)
         }
 }
 
+int x87_cachecount(dynarec_arm_t* dyn, int ninst)
```
ksco, line 977: Ah, I see, we already have `dyn->n.stack` for this, so this function is unnecessary then...

ksco:
Ok, I've tested some games on rv64, seems all good ;)

ptitSeb:
Ok, I'll push then.

PR #1030, opened 2023-10-15
[INTERP] Fix fpu_round

ksco:
The `nearbyint` functions round their argument to an integer value in floating-point format, **using the current rounding direction**, therefore it's not "ROUND_Nearest". We should use `round`.


ptitSeb:
I'm pretty sure `nearbyint` is used in more than 1 place!

ksco:
Ah, it does! Let me change all of 'em then.

ksco:
Ok, done.

ptitSeb:
Tests are not happy with this chnages

ksco:
Seems 0.5 in `cvtps2pd` should be rounded to 0. 🤔️

ptitSeb:
> Seems 0.5 in `cvtps2pd` should be rounded to 0 in "ROUND_Nearest" mode, that's really strange, am I missing something? 🤔️

Mmm no, cvtps2pd 0.5 should still be 0.5. Value should not be round to integer at all on cvtps2pd, as it's just packed float to packet double.

ksco:
Sorry I mean the cvtps2dq, these names confuses me :'(

ksco:
Ah, it actually "Round to Nearest, ties to Even", so 0.5 is expected to round to 0.

ksco:
@ptitSeb seems working; also tested test22 with `BOX64_DYNAREC_TEST=1` locally, works too (this is how I found the issue).

ptitSeb:
Ok, we don't want to set the rounding mode at each function call and not letting it reset to it's previous value

The rounding mode might needed to be tracked (might need to actually wrap `fesetround` to be able to restore old value after round is done (or do nothing if the current rounding mode is already ok).

ksco:
@ptitSeb I pushed some commits to take care of library wrapping, please review again.

PR #1044, opened 2023-10-27
[WRAPPER] Fixed my_execl

wannacu:


src/dynarec/arm64/dynarec_arm64_660f.c
```diff
@@ -1084,7 +1084,22 @@ uintptr_t dynarec64_660F(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int n
...
+                    INST_NAME("EXTRACTPS Ew, Gx, Ib");
+                    nextop = F8;
+                    GETGX(q0, 0);
+                    if (MODREG) {
+                        ed = xRAX+(nextop&7)+(rex.b<<3);
+                        u8 = F8&0b11;
+                        MOVx_REG(ed, xZR);
+                        VMOVSto(ed, q0, u8);
+                    } else {
+                        addr = geted(dyn, addr, ninst, nextop, &wback, x2, &fixedaddress, NULL, 0xfff, 0, rex, NULL, 0, 0);
```
ptitSeb, line 1097: geted as a delta of `1` here, because the opcode have an extra byte attached

src/dynarec/arm64/dynarec_arm64_660f.c
```diff
@@ -1097,21 +1112,70 @@ uintptr_t dynarec64_660F(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int n
...
+                            case 3:
+                                GO(3)
+                                break;
+                        }
+                        #undef GO
                     } else {
                         SMREAD();
-                        addr = geted(dyn, addr, ninst, nextop, &wback, x1, &fixedaddress, &unscaled, 0xfff<<2, 3, rex, NULL, 0, 1);
-                        LDW(x1, wback, fixedaddress);
+                        addr = geted(dyn, addr, ninst, nextop, &ed, x1, &fixedaddress, &unscaled, 0xfff<<4, 3, rex, NULL, 0, 0);
```
ptitSeb, line 1154: Same here, you need that delta of `1`

ptitSeb: Also, the `0xfff<<4` is not coherent with a mask of `3`. You are using a `LDW` opcode, wich is a 32bits read, so it should be `0xfff<<2` and mask `3`

src/dynarec/arm64/dynarec_arm64_660f.c
ptitSeb: it should be `0xfff<<2` and a mask of `3` here, because it's a `STW` opcoration, wich is 32bits write.

ptitSeb (changes requested):
You need to fix both `geted` invocations.

ptitSeb:
Thanks :)

Issue #1057, opened 2023-11-11
[RV64] Cannot MMap/load elf /usr/lib/x86_64-linux-gnu/libgcc_s.so.1 inspite of it existing


Vixea:
[logs.txt](https://github.com/ptitSeb/box64/files/13326539/logs.txt)

HW:
[hwinfo.txt](https://github.com/ptitSeb/box64/files/13326564/hwinfo.txt)
Distro:
Debian(RevyOS)

This happens with any program, and box64 was self compiled.
Thank you for your time. You do amazing work!

Vixea:
Oh and here's the output of ls
```
XXXXXXXX@lpi4a:~/Documents$ ls /usr/lib/x86_64-linux-gnu/
libcrypto.so.1.1  libpng12.so.0  libstdc++.so.5
libgcc_s.so.1     libssl.so.1.1  libstdc++.so.6
```

ksco:
Please paste the whole log so we know which version (commit) of box64 are you using.

Vixea:
[full_log.txt](https://github.com/ptitSeb/box64/files/13327591/full_log.txt)

ksco:
So you're not using the latest tree, please `git pull` and rebuild.

Vixea:
```
sipeed@lpi4a:/media/sipeed/167c8b5c-c517-4ec3-835b-5ef977851fc9/projects/box64$ git pull
Updating 70e4bd41..62634c6e
Fast-forward
 src/dynarec/arm64/dynarec_arm64_00.c | 52 ++++++++++++++++++++----------------
 1 file changed, 29 insertions(+), 23 deletions(-)
```
this is the only change so I doubt it would matter

ksco:
Looking at the log you pasted:

```
Box64 with Dynarec v0.2.5 936fa209 built on Oct 31 2023 20:56:46
```

https://github.com/ptitSeb/box64/commit/936fa209 was committed 2 weeks ago. Maybe you forgot to install it after the build?

Vixea:
possibly

Vixea:
[new_log.txt](https://github.com/ptitSeb/box64/files/13327620/new_log.txt)
progress I thought I had a updated version.... shrug

ksco:
I submitted a fix above.

ksco:
[...]

PR #1058, opened 2023-11-12
[ELFLOADER] Fixed align issue (may help #1057)

ksco:
for #1057

### DIFF
--- a/src/elfs/elfloader.c
+++ b/src/elfs/elfloader.c
@@ -205,8 +205,8 @@ int AllocLoadElfMemory(box64context_t* context, elfheader_t* head, int mainbin)
     void* image = mmap64((void*)(head->vaddr?head->vaddr:offs), head->memsz, 0, MAP_ANONYMOUS|MAP_PRIVATE|MAP_NORESERVE, -1, 0);
     #endif
     if(image!=MAP_FAILED && !head->vaddr && image!=(void*)offs) {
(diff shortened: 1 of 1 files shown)


ptitSeb:
There is no mention in linux docs that `lenght` should be aligned to pagesize. Only `addr` and `offset` need to be. Did you test your change extensively? because elfloader is really sensitive! Does this change fix the ticket?

ksco:
IIUC, at line 258, we're calculating an aligned size for current fragment which will be used to map memory for the fragment. So we should use the `head->multiblocks[n].paddr` instead of `e->p_paddr`.

Before this fix, wine will crash at line 312 inside fread as we're not mapping enough memory for it. 

Some logs after the fix:
```
Mmap64 for (@0x7f00000000 0x2bc500) for elf "/home/debian/wine-8.20/bin/../lib/wine/x86_64-unix/ntdll.so" returned 0x3fd6c54000 instead
Using emulated /home/debian/wine-8.20/bin/../lib/wine/x86_64-unix/ntdll.so
Mmap64 for (@0x7f01000000 0x1bbc8) for elf "/lib/x86_64-linux-gnu/libunwind.so.8" returned 0x3fd6c38000 instead
Using emulated /lib/x86_64-linux-gnu/libunwind.so.8
```

ksco:
> Does this change fix the ticket?

According to the reporter, yes, and it fixed wine for me as well.

PR #1074, opened 2023-11-20
[RV64_DYNAREC] Fixed zbb andn optimization

ksco:
`andn` is `rs1 & ~rs2`. @ptitSeb can you test if this fixes the wine issue on VF2?


ptitSeb:
It probably help, but it's not enough and wine still crashes on the VF2.

PR #1110, opened 2023-12-04
[TERMUX] Fix loading box64rc and lib x86_64-linux-gnu path on termux

LynYugiri:
Now on the Termux.
The default path list:
 box64.box64rc at /data/data/com.termux/files/usr/etc/box64.box64rc
lib x86_64-linux-gnu dirctory at /data/data/com.termux/files/usr/lib/x86_64-linux-gnu


mcagabe19:
@Creeper-xie we needed in box86 too

PR #1126, opened 2023-12-08
[DYNAREC_RV64] Added more opcodes for flatout.exe

ksco:


src/dynarec/rv64/dynarec_rv64_dc.c
```diff
@@ -77,7 +120,7 @@ uintptr_t dynarec64_DC(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                     v2 = fpu_get_scratch(dyn);
                     addr = geted(dyn, addr, ninst, nextop, &wback, x2, x1, &fixedaddress, rex, NULL, 1, 0);
                     FLD(v2, wback, fixedaddress);
-                    FCOMD(v1, v2, x1, x2, x3, x4, x5);
+                    FCOMD(v1, v2, x1, x6, x3, x4, x5);
```
ptitSeb, line 123: It was ok to use `x2` here, because the read value will never be written back, so wback is not used anymore and x2 is indeed free.
It doesn't hurt to use x6 tho.

PR #1132, opened 2023-12-09
[DYNAREC_RV64] Added more opcodes for flatout.exe

ksco:
Note to myself: after adding `PSLLQ`, the opening animation becomes abnormal.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_0f.c
+++ b/src/dynarec/rv64/dynarec_rv64_0f.c
@@ -1748,10 +1748,10 @@ uintptr_t dynarec64_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             GETGM();
             GETEM(x5, 0);
             for (int i = 0; i < 2; ++i) {
-                LHU(x1, gback, gdoffset + i * 4);
-                LHU(x2, gback, gdoffset + i * 4 + 2);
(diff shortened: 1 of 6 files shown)


src/dynarec/rv64/dynarec_rv64_0f.c
```diff
@@ -901,6 +894,94 @@ uintptr_t dynarec64_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+                    break;
+                case 4:
+                    INST_NAME("PSRAD Em, Ib");
+                    GETEM(x4, 1);
+                    u8 = F8;
+                    if (u8 > 31) u8 = 31;
+                    if (u8) {
+                        LD(x1, wback, fixedaddress);
+                        SRAI(x2, x1, 32 + u8);
+                        SRAIW(x1, x1, u8);
```
ptitSeb, line 924: Nice trick :)

ksco:
> Note to myself: after adding `PSLLQ`, the opening animation becomes abnormal.

Fixed, it's `MOVD Gm, Ed`.

src/dynarec/rv64/dynarec_rv64_0f.c
ksco: Glad that you like it! Don't know if it's better than two `LWU` though.

PR #1135, opened 2023-12-13
fix emit_shr16

liuli0217:
For the SHR instruction, the OF flag is set to the most-significant bit of the original operand.


ptitSeb:
AH yes, too many copy/pasta on my side. Nice catch!

PR #1142, opened 2023-12-15
Fix raspi5 building

mcagabe19no2fa:


sundhaug92:
Good spot on CMakeLists, but why remove the comment in COMPILE.md?

mcagabe19no2fa:
> Good spot on CMakeLists, but why remove the comment in COMPILE.md?

u may needed check my changes carefully
image

PR #1164, opened 2023-12-25
[DYNAREC_RV64] Added more opcodes for VMP-protected GI

xctan:


src/dynarec/rv64/dynarec_rv64_66.c
```diff
@@ -619,6 +618,15 @@ uintptr_t dynarec64_66(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
             OR(xRAX, xRAX, x1);
             break;
+        case 0x99:
+            INST_NAME("CWD");
+            SLLI(x1, xRAX, 48);
+            SRAI(x1, x1, 48);
+            SRLI(x1, x1, 48);
+            SRLI(xRDX, xRDX, 16);
+            SLLI(xRDX, xRDX, 16);
+            OR(xRDX, xRDX, x1);
```
ptitSeb, line 628: Are you sure about thios one? it looks a bit strange for the RDX stuffs

xctan: After line 625 x1[15:0] should be the result of sign extension of AX (and x1[63:16] are zero), and then the OR assigns it to RDX[15:0] (i.e. DX) without clobbering other bits. Am I understanding CWD correctly?

ptitSeb: yes, that' correct. Sorry, I read a bit too quickly.

PR #1168, opened 2023-12-26
[DYNAREC_RV64] Added more opcodes for VMP-protected GI

xctan:
66 0F BD BSR seems to trigger some unknown bugs in other opcodes.


ptitSeb:
> 66 0F BD BSR seems to trigger some unknown bugs in other opcodes.

So it ready to merge or shall I wait?

xctan:
Ready to merge. VMP should entirely run inside dynarec after this PR, but I'm unsure whether the hidden bug can be found in a few days.

ptitSeb:
Ok.

xctan:
The bug is that VMP cannot generate a meaningful anti-debugger message but garbages (so still having problem with debugger check).

PR #1180, opened 2023-12-31
[MEMORY] Switched from a sparse array to a red-black tree

rajdakin:
This PR switches the way memory permission is tracked, from a sparse array to a red-black tree (self-balancing binary search tree). The notion of hot pages has also been removed.
Performance does not seem to be very impacted, and memory usage has decreased a bit for processes which used a lot of RAM (for example, Steam uses ~100M less memory and every process of wine uses ~15M less memory).


hisham2630:
this change prevent gta v from loading properly it shows social club failed to intialize error 4
Just cause 2 won't launch at all.

PR #1185, opened 2024-01-04
[WRAPPERS] Add a `isSimpleWrapper` function for RV64

rajdakin:
This PR adds support for the `isSimpleWrapper` function on RISC-V 64. It also adds a default function returning 0 if the architecture is unsupported (i.e. not ARM or RISC-V).
There is also minor fixes in the script (the "free letters" comment does not contain `x` and `X`, an error message was incorrect, and a type hint was invalid for `mypy`).
The first commit contains the entire change, the second contains only a version bump.


ksco (approved):
We’re filtering all the sign-extended ones for RV64? This makes the simple functions set much smaller than AArch64, but it’s a start to enable this feature I think.

ksco:
A more complicated proposal: make `isSimpleWrapper` return a uint64_t and every 2-bits indicates the width of the current sign-extended operand:

```
00 -> not sign-extended
01 -> signed char
10 -> short
11 -> int
```

ptitSeb:
> A more complicated proposal: make `isSimpleWrapper` return a uint64_t and every 2-bits indicates the width of the current sign-extended parameter/return value:
> 
> ```
> 00 -> not sign-extended
> 01 -> signed char
> 10 -> short
> 11 -> int
> ```

Return value is already used to check float / double stuff iirc.

PR #1192, opened 2024-01-07
expose minimal libxcb-xinput wrapper

fbenkstein:
This exposes a minimal `libxcb-xinput` interface as well as the `XkbGetIndicatorState` function from libx11.

With this change I was able to run the [Enpass password manager](https://www.enpass.io/) downloaded from  https://apt.enpass.io/pool/main/e/enpass/enpass_6.9.2.1563_amd64.deb. My goal is to run this on a Raspberry Pi 5 but so far I only tested it on a Debian 12 VM on a MacBook Pro M1. It runs slow but does work and even supports the browser extension.

For anyone who also needs access to their Enpass vault on arm64, note that it also requires setting the following environment variable: `QT_XCB_NO_XI2=1`.


ptitSeb:
Nice,, thank you.

PR #1196, opened 2024-01-11
[RV64_DYNAREC] Fixed 66 0F B6 MOVZX opcode

xctan:
After tedious bisecting with BOX64_NODYNAREC, I found that this opcode was not correctly implemented. However, a strange SIGBUS error pops up when I'm trying to run YuanShen.exe (Genshin Impact) with this patch. If the fix is only applied to VMP areas (by checking `addr` in dynarec), the program can produce a meaningful "A debugger has been found running in your system.\nPlease, unload it from memory and restart your program." message instead of garbage as before.


ptitSeb:
Does that means it's not ready to merge yet?

xctan:
Ready to merge. I've tested wine explorer.exe and Flatout, and both are working.

PR #1198, opened 2024-01-12
[RV64_DYNAREC] Fixed 66 87 XCHG opcode

xctan:
Fixed SIGBUS error.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_66.c
+++ b/src/dynarec/rv64/dynarec_rv64_66.c
@@ -521,11 +521,13 @@ uintptr_t dynarec64_66(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 NOT(x5, x5);
                 SLLI(x5, x5, 16);
                 SLLI(x6, x6, 16);
+                // 0xffffffff0000ffff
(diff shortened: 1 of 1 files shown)


src/dynarec/rv64/dynarec_rv64_66.c
```diff
@@ -518,14 +518,13 @@ uintptr_t dynarec64_66(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 B_MARK2_nocond;
 
                 MARK3;
-                NOT(x5, x5);
-                SLLI(x5, x5, 16);
                 SLLI(x6, x6, 16);
-                NOT(x5, x5);
-                LR_W(x1, ed, 1, 0);
+                MOV64x(x5, 0xffffffff0000ffffULL);
```
ptitSeb, line 522: Why the mov? the `not(x5, x5)` after the `lui(x5, 0xffff0)` was not working fine?

xctan: Sorry for confusion. NOT+SLLI+NOT is working fine, but I was confused at that moment so I wrote this alternative version. I'll switch back to the old good implementation tomorrow.

Issue #1200, opened 2024-01-12
SteamApi_Init() fails on latest version


Sirox228:
i had a box64 version from +- December 12, just updated to latest one and it seems that steamapi in unreal engine games is no more working (it was working fine just before the update) ![Screenshot_2024-01-12-20-54-22-823_com termux x11](https://github.com/ptitSeb/box64/assets/71220271/f3392acf-145f-4124-b314-696cc7831d97)
the log after dxvk initialization is only those 2 lines:
```
shared memfd open() failed: Invalid argument
shared memfd open() failed: Invalid argument
```

ptitSeb:
you launch steam game on mobox? or is it some other solution? Is taht a windows version of steam or linux version? And what game what is?

Sirox228:
i'm running guilty gear strive on wine ge 8.13, it's pirated version of the game, i assume it uses goldberg steam emu. it was always working without steam, like any such game should work, but just in case i tried to start it while steam was running - same issue

ptitSeb:
Ok. It would be nice to do a bisect to find the offending commit, but I know it's difficult to do in mobox.

Sirox228:
i can do that in box64droid, it have same issue, but uses proot/chroot so i can compile it myself

ptitSeb:
> i can do that in termux-box, it have same issue, but uses proot/chroot so i can compile it myself

That would be very helpfull!

Sirox228:
i made the bisection. it seems that #5e9e1fa made the issue

ptitSeb:
Oh, the memory tracking change? that's strange. Well, ok, I'll have a deeper look at that tomorrow. Thanks for the bisect.

Sirox228:
issue is still there on latest version (with merged pr)

hisham2630:
> issue is still there on latest version (with merged pr)

same here

Sirox228:
fixed now

PR #1201, opened 2024-01-12
[RBTREE] Reverted memory tracking to 32 bits

rajdakin:
Some protections are in fact 32 bits wide (`PROT_GROWSDOWN`, `PROT_GROWSUP`). This PR reverts the narrowing to an 8-bit integer for the memory tracking (fixes *maybe* #1200 and similar issues).


ptitSeb:
I don't think it will fix #1200 because previous tracking was also 8bits. But haveing 32bits now is cheap in term of memory consuption, so why not.

PR #1215, opened 2024-01-23
 [DYNAREC_RV64] Enable basic simple wrapper optimization

ksco:


ptitSeb:
Is it ready to merge? float/double return are still disabled it seems for now.

ksco:
Yes it can be merged. For float/double, I think we need to know it's float or double than just number of args for sse_get_reg to work, this is the same situation as signed integers, which will be implemented later.

PR #1233, opened 2024-01-30
system/box64.conf.cmake: change rule name to be box64 specific

josch:
Having an entry named x86_64 in /proc/sys/fs/binfmt_misc/ is not ideal because:

 - it does not indicate who the binfmt provider is (others start with `llvm-*`, `python*` or `qemu-*` to indicate this)
 - it is a namespace grab -- there is no justification for box64 to grab the identifier "x86_64" over all other potential binfmt providers


mcagabe19 (approved):
indeed

ptitSeb:
Oh sure. I named it x86_64 because it was what it did, not to grab anything.

PR #1257, opened 2024-02-10
Wrapped libraries refactor

rajdakin:
This PR refactors various common operations on wrapped libraries initialization/finalization:
- calling `getMy(lib)` in `CUSTOM_INIT` and `freeMy()` in `CUSTOM_FINI`, toggled by the `HAS_MY` macro;
- calling `SETALT` in `CUSTOM_INIT` into the `ALTMY` macro;
- calling `setNeededLibs(lib, n, ...)` in `CUSTOM_INIT`, called with the correct `n` using the `NEEDED_LIBS` macro.


src/wrapped/wrappedgdk3.c
```diff
@@ -181,22 +179,12 @@ EXPORT uint32_t my3_gdk_threads_add_timeout_full(x64emu_t* emu, int priotity, ui
     return my->gdk_threads_add_timeout_full(priotity, interval, findGSourceFunc(f), data, findGDestroyNotifyFct(d));
 }
 
-
 #define PRE_INIT    \
     if(box64_nogtk) \
         return -1;
 
-#define CUSTOM_INIT \
-    libname = lib->name;            \
```
ptitSeb, line 190: this line about libname disapeared?

src/wrapped/wrappedgtkx112.c
```diff
@@ -20,7 +20,7 @@
 #include "gtkclass.h"
 
 const char* gtkx112Name = "libgtk-x11-2.0.so.0";
-static char* libname = NULL;
```
ptitSeb, line 23: this was here because it use to be able to load gtk-1 also. I guess this is obsolete and can be remove completly, especially on 64bits?

src/wrapped/wrappedgdk3.c
rajdakin: Yes, `libname` is unused in the entire file.

src/wrapped/wrappedgtkx112.c
rajdakin: Yes. `wrappedgdkx112.c` still requires it, but `wrappedgtkx112.c` does not.

PR #1260, opened 2024-02-10
more wrapped functions in libxcb-randr and libcap

Coreforge:
These functions are needed for SteamVR's vrcompositor and vrcompositor-launcher to run.


ptitSeb:
You are able to run SteamVR on an arm platform with box64?

Coreforge:
I'm able to run it (with some options like `BOX64_DYNAREC_STRONGMEM` set, though I haven't tested yet which ones are actually required), but I haven't gotten frame submission to work yet.

PR #1274, opened 2024-02-21
[LIBTOOL] Fixed my_pthread_attr_init

ksco:


src/libtools/threads.c
```diff
@@ -251,8 +251,8 @@ static void* pthread_routine(void* p)
 #define PTHREAD_ATTR_UNALIGN(A)
 #define PTHREAD_ATTR(A) 	A
 #else
-#define PTHREAD_ATTR_ALIGN(A) pthread_attr_t aligned_attr = {0}; if(A) memcpy(&aligned_attr, A, 56)
-#define PTHREAD_ATTR_UNALIGN(A) if(A) memcpy(A, &aligned_attr, 56)
+#define PTHREAD_ATTR_ALIGN(A) pthread_attr_t aligned_attr = {0}; if(A) memcpy(&aligned_attr, A, sizeof(pthread_attr_t))
+#define PTHREAD_ATTR_UNALIGN(A) if(A) memcpy(A, &aligned_attr, sizeof(pthread_attr_t))
```
ksco, line 255: This is unnecessary, but it has mo harm anyway.

ptitSeb: I'm not sure this is correct. Especially the "unalign" part: a pthread_attr_t is 56 bytes on x86_64, if you copy more than 56 bytes (like 64 on arm64), you will overwrite some memory not belonging to the pthread_attr_t

So, Im' pretty sure you will break things with this change.

src/libtools/threads.c
```diff
@@ -366,6 +366,7 @@ EXPORT int my_pthread_attr_init(x64emu_t* emu, pthread_attr_t* attr)
 	PTHREAD_ATTR_ALIGN(attr);
 	int ret = pthread_attr_init(PTHREAD_ATTR(attr));
 	PTHREAD_ATTR_UNALIGN(attr);
+	return ret;
```
ptitSeb, line 369: Oh, missing ret is bad

src/libtools/threads.c
ptitSeb: Yes, they are bigger so they can handled some miss-alignment.
But here, the it will be created on the stack and will be correctly aligned because of that, so the extrat space is not use and so the actual content is on front and can be just copy back and forth with no issue.

PR #1275, opened 2024-02-21
[DYNAREC_RV64] Minor OF flag fixes and optimizations for emit shift utils

ksco:


src/dynarec/rv64/dynarec_rv64_emit_shift.c
```diff
@@ -38,7 +38,7 @@ void emit_shl32(dynarec_rv64_t* dyn, int ninst, rex_t rex, int s1, int s2, int s
         SUBI(s5, s2, rex.w?64:32);
         NEG(s5, s5);
         SRL(s3, s1, s5);
-        ANDI(s5, s3, 1); // F_CF
+        ANDI(s5, s3, 1); // LSB == F_CF
```
ptitSeb, line 41: Why not moving that ̀`ANDI` inside the `IFX`?

ksco:
Interestingly, test.vmp.exe is now working on RISC-V, the div by zero hack is not required.

src/dynarec/rv64/dynarec_rv64_emit_shift.c
ksco: Doing that, we will need another ANDI below where the `s5` is used again.

src/dynarec/rv64/dynarec_rv64_emit_shift.c
```diff
@@ -94,12 +97,10 @@ void emit_shl32c(dynarec_rv64_t* dyn, int ninst, rex_t rex, int s1, uint32_t c,
     IFX(X_CF|X_OF) {
         if (c > 0) {
             SRLI(s3, s1, (rex.w?64:32)-c);
-            ANDI(s5, s3, 1); // F_CF
+            ANDI(s5, s3, 1); // LSB == F_CF
```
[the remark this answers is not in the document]
ptitSeb: Ah no, maybe it's used for OF also.

src/dynarec/rv64/dynarec_rv64_emit_shift.c
ptitSeb: Ah no, it's used for OF also

ptitSeb:
> Interestingly, test.vmp.exe is now working on RISC-V, the div by zero hack is not required.

Ah, just accurate OF was enough? Good!
(still I suggest you add the div0 hack later)

PR #1276, opened 2024-02-22

ksco:


src/dynarec/rv64/dynarec_rv64_0f.c
```diff
@@ -1874,24 +1886,6 @@ uintptr_t dynarec64_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 SH(x3, gback, gdoffset + 2 * i);
             }
             break;
-            INST_NAME("PSUBSW Gm,Em");
```
ptitSeb, line 1877: That seems to be "F9" opcode

ptitSeb: But it's alreay there. Strange

ksco: > But it's alreay there. Strange

It's a mistake from before.

PR #1297, opened 2024-02-29
[LONGAARCH_DYNAREC] Added 00...05 and 28...2D opcodes

factfinding:
[LONGAARCH] More renamed arch to LA64
[LONGAARCH_DYNAREC] Added 00...05 and 28...2D opcodes


ksco:
.. so we don't have to deal with eflags and x87 emulation, it's slow and painful.

ptitSeb:
this extension is present on all cpu right?

ksco:
> this extension is present on all cpu right?

Yes, I think so.

factfinding:
Sorry, I am not an internal employee of Loongson Corporation, so I can only use the Loongarch instruction set that Loongson has already made public to implement dynamic binary. The LBT instruction is not publicly available, so I don't know how to use it. Loongarch and riscv are very similar, so I just imitated riscv to implement dynamic binary.

ksco:
Me neither ;) but there are some reverse-engineered documents we can refer: https://github.com/jiegec/la-inst/blob/master/LBT.md, https://github.com/loongson-community/loongarch-opcodes/blob/develop/lbt.txt. I'll look into it and see if it's enough for us to use LBT.

ptitSeb:
As I see it, it's fine to use RiSCV implementation as a 1st draft for the Loongarch dynarec.
LBT extension is clearly a plus, but it can come after, as a replacement. And probably not all cases are covert by LBT anyway.

PR #1298, opened 2024-02-29
[CI] Added loongarch64

ksco:


ptitSeb:
Ah thanks,  it will be needed!

PR #1304, opened 2024-02-29
[LA64_DYNAREC] Added more opcodes and fixed more things

ksco:


src/dynarec/la64/dynarec_la64_00.c
```diff
@@ -193,6 +209,18 @@ uintptr_t dynarec64_00(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 SMWRITELOCK(lock);
             }
             break;
+        case 0x8B:
+            INST_NAME("MOV Gd, Ed");
+            nextop=F8;
+            GETGD;
+            if(MODREG) {
+                MVxw(gd, xRAX+(nextop&7)+(rex.b<<3));
```
ptitSeb, line 217: Probably need `TO_LA64`macro here

ksco: Ah yes, sorry

ksco: It's a bit annoying the register number can continuous.

ptitSeb: An alternative way could be also: `(nextop&7)+((rex.b<<3)?xR8:xRAX)` not sure it's usefull, but that can be a possibility.

ptitSeb: > It's a bit annoying the register number cannot be continuous.

I agree, but there is no choice :(

src/dynarec/la64/dynarec_la64_emit_math.c
```diff
@@ -171,9 +170,8 @@ void emit_add32c(dynarec_la64_t* dyn, int ninst, rex_t rex, int s1, int64_t c, i
     IFX(X_CF)
     {
         if (rex.w) {
-            MOV32w(x2, 0xffffffff);
```
ptitSeb, line 174: oh, yeah, that was just wrong before. mixing `x2` direct access register with `s1`/`s2` arbitrary scratch register is a sure fail at some point!

PR #1307, opened 2024-03-01
[LA64_DYNAREC] Added 70-7F Jcc opcodes, refine printer and some fixes too

ksco:


ksco:
Hmmm, qemu-loongarch64 does not have the LBT extension?!

ksco:
Proposals:

1. Disable LA64 Dynarec CI.
2. Make LBT extension optional and only run the tests without LBT -- which means we need to implement eflags handling and x87 infra twice.

Maybe someday we will be able to setup a self hosted Loongarch64 runner.

PR #1315, opened 2024-03-03
We can use cpucfg to get la64_lbt_x64 support

factfinding:


ksco:
The BOX64_DYNAREC_LA64NOEXT option is there to be used manually -- to enable/disable the LBT code path, because it would be interesting to see the performance difference w/ or w/o LBT.

ksco:
Plus, isn’t LBT always available on all the LoongArch CPUs?

ksco:
Okay, I was told that LBT does not exist on some low-end LoongArch CPUs like 2k1000. Probably nobody would run box64 on those CPUs, it would be nice to detect if the extension is supported anyway.

Could you revert current change and instead add a `LA64_Detect_Function()` like what's done in RISC-V?

factfinding:
Okay, you are right. Almost all the LoongArch CPUs have LBT, the pull request is useless.

PR #1319, opened 2024-03-03
[LA64_DYNAREC] Added more opcodes and more fixes, SuperHexagon is working

ksco:
test01 fully implemented.


ksco:
Hmmm, superhexagon stops working after I pushed some new commits...

ptitSeb:
Ok, I'll wait then. Ping me when it works again

ksco:
superhexagon is back to party.

PR #1324, opened 2024-03-04
[LA64_DYNAREC] Utilizing bitmanip instructions because they're great

ksco:
More rework on the eflags computation TBD.

### DIFF
--- a/src/dynarec/la64/dynarec_la64_00.c
+++ b/src/dynarec/la64/dynarec_la64_00.c
@@ -94,9 +94,7 @@ uintptr_t dynarec64_00(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             u8 = F8;
             ANDI(x1, xRAX, 0xff);
             emit_add8c(dyn, ninst, x1, u8, x3, x4, x5);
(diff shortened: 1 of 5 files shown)


src/dynarec/la64/dynarec_la64_00.c
```diff
@@ -476,26 +472,8 @@ uintptr_t dynarec64_00(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
-                        }
-                        OR(eb1, eb1, x4);
-                    }
-                } else {
-                    ADDI_W(x3, xZR, 0xf00); // mask ffffffffffffff00
-                    AND(eb1, eb1, x3);
-                    ORI(eb1, eb1, u8);
-                }
+                MOV32w(x3, u8);
+                BSTRINS_D(eb1, x3, eb2 * 8 + 7, 8);
```
ptitSeb, line 476: shouldn't this be
```C
 BSTRINS_D(eb1, x3, eb2 * 8 + 7, eb2 * 8);
```
instead?

ksco: Yeah, thank you for saving the future me.

PR #1327, opened 2024-03-04
[LA64_DYNAREC] Added more opcodes

ksco:


src/dynarec/la64/la64_emitter.h
```diff
@@ -221,19 +221,28 @@ f24-f31  fs0-fs7   Static registers                Callee
 
 #define NOP() ANDI(xZR, xZR, 0)
 
-// tmp = SLL(GR[rj][31:0], GR[rk][4:0])
+// tmp = SLL(GR[rj][31:0], ui5)
```
ptitSeb, line 224: why this change in the comment? The previous version seemed more precise than the new one.

ksco: It is directly copied from the spec -- but I agree with you, the spec seems not consistent.

PR #1330, opened 2024-03-05
[LA64_DYNAREC] Added more opcodes and some fixes in the printer

ksco:


ksco:
Note: Although LL/SC can specify an offset, we cannot use it in the `LOCK` prefix instructions because the offset must be aligned to 4 bytes, while on x86 the offset can be arbitrary. To use it, some modifications to `geted` is needed.

PR #1335, opened 2024-03-06
Fixed a warning

ksco:


src/libtools/threads.c
```diff
@@ -253,7 +253,7 @@ static void* pthread_routine(void* p)
 #else
 #define PTHREAD_ATTR_ALIGN(A) pthread_attr_t aligned_attr = {0}; if(A) memcpy(&aligned_attr, A, 56)
 #define PTHREAD_ATTR_UNALIGN(A) if(A) memcpy(A, &aligned_attr, 56)
-#define PTHREAD_ATTR(A)		(A)?&aligned_attr:NULL
```
ksco, line 256: Ah so this is a warning we have to live with.

ksco:
Hmmm, we need a retry on the LoongArch qemu too.

PR #1339, opened 2024-03-07
[LA64_DYNAREC] Added C1 /4 and /6 opcodes

factfinding:


src/dynarec/la64/dynarec_la64_00.c
```diff
@@ -657,6 +657,15 @@ uintptr_t dynarec64_00(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                     emit_shr32c(dyn, ninst, rex, ed, u8, x3, x4);
                     if (u8) { WBACK; }
                     break;
+                case 4:
+                case 6:
+                    INST_NAME("SHL Ed, Ib");
+                    SETFLAGS(X_ALL, SF_SET_PENDING); // some flags are left undefined
```
ptitSeb, line 663: Flags should only be set if masked u8 is not 0. look at ARM64 to see how to do it.

PR #1370, opened 2024-03-19
Add 2 more function to gtk3

LiZhuoheng-coder:
Add gtk_file_filter_add_custom and gtk_style_context_get_style


src/wrapped/wrappedgtk3_private.h
```diff
@@ -2828,7 +2828,7 @@ GO(gtk_style_context_get_scale, iFp)
 GO(gtk_style_context_get_screen, pFp)
 GO(gtk_style_context_get_section, pFpp)
 GO(gtk_style_context_get_state, uFp)
-//GO2(gtk_style_context_get_style, vFpV, gtk_style_context_get_style_valist)
+GO2(gtk_style_context_get_style, vFpV, gtk_style_context_get_style_valist)
```
ptitSeb, line 2831: No, that doesn't work like that on x86_64, only on x86 will this trick work (that's why it was commented)

src/wrapped/generated/wrappedgtk3types.h
```diff
@@ -66,6 +66,7 @@ typedef void* (*pFpipppppppi_t)(void*, int32_t, void*, void*, void*, void*, void
 	GO(gtk_print_job_send, vFpppp_t) \
 	GO(gtk_stock_set_translate_func, vFpppp_t) \
 	GO(gtk_tree_sortable_set_default_sort_func, vFpppp_t) \
+	GO(gtk_file_filter_add_custom, vFpuppp_t) \
```
ptitSeb, line 69: did you add the type by hand? it should be all automatic in the generated folder

PR #1389, opened 2024-03-28
Feature: build libbox64.so and expose 3 APIs.

Howard0o0:
In some cases we need to use box64 as a library.

For example, we have a python program depending on some x64 libraries. If we run this python program on arm64 platform, it would fail on opening the x64 libraries.   

To run this python program, we could emulate the python interpreter, which means emulating the whole python code and the x64 libraries. It works but not efficient. Cause we only need to emulate the x64 libraries, emulating the whole python code is not necessary.  In this case, we could create wrapper libraries, which kidnap the python code calling to these x64 libraries. 

libbox64_use_case


ptitSeb:
Building box64 as a lib should be optionnal, and if activated, box64 binary should not be built.

I expect building box64 as a lib to require source code also, to define name of the lib and setup the API with wrapped function. In my point of view, box64 as lib is a submodule of the lib itself. so if you wrap "libmyapi.so", you will generate an actual "libmyapi.so" when building box64 as a lib, that will load the x86 version of "libmyapi.so" somewhere.

Howard0o0:
Thank you for the prompt response! :D

The initial point is clear; building box64 as lib should be optional.

Could you elaborate on the second point? Specifically, building box64 as a lib to require source code? Providing some examples would be helpful. Does this imply that exposing API is unnecessary?

PR #1407, opened 2024-04-03
[RV64_DYNAREC] Add 1E/1F opcode

LiZhuoheng-coder:


ksco:
You're pushing tabs again.

ksco:
Use `git clang-format --style=file` to pick up the format rules.

src/dynarec/rv64/dynarec_rv64_00_0.c
```diff
@@ -317,7 +321,27 @@ uintptr_t dynarec64_00_0(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+                DEFAULT;
+            }
+	    break;
+        case 0x1F:
+            if (rex.is32bits) {
+                INST_NAME("POP DS");
+                SMREAD();
+                POP1_32(x1);
+                SH(x1, xEmu, offsetof(x64emu_t, segs[_DS]));
+                SW(x1, xEmu, offsetof(x64emu_t, segs_serial[_DS]));
```
ksco, line 340: Save x1 to `segs_serial` does not make sense to me, other `POP`s implemented in previous PR share the same issue.

ksco: Looking at ARM64 DynaRec, it's saving xZR to `segs_serial` to indicate there need a re-compute on the segment offsets. We should do the same.

ksco (approved):

PR #1410, opened 2024-04-04
[LA64_DYNAREC] Added more SSE/SSE2 instructions

factfinding:


src/dynarec/la64/dynarec_la64_f30f.c
```diff
@@ -53,6 +53,22 @@ uintptr_t dynarec64_F30F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+                v1 = sse_get_reg(dyn, ninst, x1, (nextop & 7) + (rex.b << 3), 0);
+                GETGX_empty(v0);
+                FMOV_D(v0, v1);
+            } else {
+                GETGX_empty(v0);
+                SMREAD();
+                addr = geted(dyn, addr, ninst, nextop, &ed, x1, x2, &fixedaddress, rex, NULL, 1, 0);
+                FLD_D(v0, ed, fixedaddress);
+            }
+            ST_D(xZR, xEmu, offsetof(x64emu_t, xmm[gd]) + 8);
```
ksco, line 70: (Sorry for the late review, I took a day off yesterday.)

I don't think this is correct, instead of clear the upper part with a `ST_D`, you should do it on the LSX register directly. Unlike the scalar instructions, on the SSE/SSE2 part, RV64 and LA64 DynaRec are very very different.

src/dynarec/la64/dynarec_la64_f20f.c
```diff
@@ -0,0 +1,85 @@
...
+                v0 = sse_get_reg(dyn, ninst, x1, gd, 1);
+                d0 = sse_get_reg(dyn, ninst, x1, ed, 0);
+                FMOV_D(v0, d0);
+            } else {
+                SMREAD();
+                v0 = sse_get_reg_empty(dyn, ninst, x1, gd);
+                addr = geted(dyn, addr, ninst, nextop, &ed, x1, x2, &fixedaddress, rex, NULL, 8, 0);
+                FLD_D(v0, ed, fixedaddress);
+                // reset upper part
+                ST_D(xZR, xEmu, offsetof(x64emu_t, xmm[gd])+8);
```
ksco, line 62: Again, clear the upper part on the LSX register directly.

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -195,6 +195,17 @@
...
 
+// Get Ex as a double, not a quad (warning, x1 get used, x2 might too)
+#define GETEXSD(a, D)                                                                        \
+    if (MODREG) {                                                                            \
+        a = sse_get_reg(dyn, ninst, x1, (nextop & 7) + (rex.b << 3), 0);                     \
+    } else {                                                                                 \
+        SMREAD();                                                                            \
+        a = fpu_get_scratch(dyn);                                                            \
+        addr = geted(dyn, addr, ninst, nextop, &ed, x1, x2, &fixedaddress, rex, NULL, 1, D); \
+        FLD_D(a, ed, fixedaddress);                                                            \
```
ksco, line 206: This seems incorrect too, the upper 64 bits are left undefined unfortunately.

ksco (commented):
Not sure how this is tested, but the I found some places confusing.

src/dynarec/la64/dynarec_la64_f30f.c
ptitSeb: Ah yes. As a general rules, the model for SSEx opcode on LA64 should be ARM64, not RV64.

factfinding:
I just test by tests/benchfloat and these code compile by gcc test.c -o test -msse2 -O2
```
#include <stdio.h>
#include <emmintrin.h>
 
int main()
{
    double a = 2.578;
    double b = -4.2597;
    double c = 0.0;
    double d = 0.0;
    __m128d sa = _mm_load_sd(&a);
    __m128d sb = _mm_load_sd(&b);
    __m128d sc = _mm_add_sd(sa, sb);
    __m128d sd = _mm_sub_sd(sa, sb);
    _mm_store_sd(&c, sc);
    _mm_store_sd(&d, sd);
    printf("%lf %lf\n", c, d);
    return 0;
}
```
The results look all good, so I think code is correct. It looks like I forget some boundary problem. By a way, my tests are all running on the qemu-loongarch, perhaps the results of virtual machines and real machines are different?

ptitSeb:
This sample code will not show you the issue.
Try to load 4 packed float as `a` instead of a single float, do a add_sd on a than print the 4 float value in a. That should highlight the issue.

ksco:
> my tests are all running on the qemu-loongarch, perhaps the results of virtual machines and real machines are different?

Not sure, but qemu-user will definitely not enough for box64 development.

PR #1421, opened 2024-04-06
[LA64_DYNAREC] Added more SSE/SSE2 instructions

factfinding:
* Added F2 0F 11 MOVSD opcode

* Added F2 0F 59 MULSD opcode

* Added F2 0F 5E DIVSD opcode

* Added some log information

### DIFF
--- a/src/dynarec/la64/dynarec_la64_f20f.c
+++ b/src/dynarec/la64/dynarec_la64_f20f.c
@@ -62,6 +62,20 @@ uintptr_t dynarec64_F20F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int 
             }
             VEXTRINS_D(v0, v1, 0); // v0[63:0] = v1[63:0]
             break;
+        case 0x11:
+            INST_NAME("MOVSD Ex, Gx");
+            nextop = F8;
+            GETG;
+            v0 = sse_get_reg(dyn, ninst, x1, gd, 0);
+            if(MODREG) {
+                ed = (nextop&7)+ (rex.b<<3);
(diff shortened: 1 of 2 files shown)


src/dynarec/la64/dynarec_la64_f20f.c
```diff
@@ -62,6 +62,21 @@ uintptr_t dynarec64_F20F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
             break;
+        case 0x11:
+            INST_NAME("MOVSD Ex, Gx");
+            nextop = F8;
+            GETG;
+            v0 = sse_get_reg(dyn, ninst, x1, gd, 0);
+            if(MODREG) {
+                ed = (nextop&7)+ (rex.b<<3);
+                d0 = sse_get_reg(dyn, ninst, x1, ed, 0);
+                FMOV_D(d0, v0);
```
ksco, line 73: This is again not correct, the upper 64 bit of the corresponding lsx are left undefined. You should really be careful on the SSE instructions. Use `VEXTRINS_D` instead.

ksco: (I'll fix this next PR)

ksco: No, the `vextrins.d` retains the untouched part (and that's what we want here, keep the upper part unchanged), but `fmov.d` make the untouched part undefined. See: https://chipsandcheese.com/2023/02/26/loongsons-lsx-and-lasx-vector-extensions/

factfinding: I see, thinks a lot!

PR #1440, opened 2024-04-12
[RV64_DYNAREC] Fixed 66 C7 MOV opcode

ksco:
Copy-paste is the new way of code review. This issue can be traced back to the very beginning, but we discovered it only now, and not through debugging.


ksco:
Hmm what happened to the CI, why it does not start...

ptitSeb (approved):
Oh, Shift Right instead of Shift Left... Nice find!

PR #1444, opened 2024-04-14
[LA64_DYNAREC] Added 1 more opcode and some fixes too

ksco:


src/dynarec/la64/dynarec_la64_emit_shift.c
```diff
@@ -444,3 +444,58 @@ void emit_ror32c(dynarec_la64_t* dyn, int ninst, rex_t rex, int s1, uint32_t c,
...
+    }
+
+    if (la64_lbt) {
+        IFX (X_ALL) {
+            if (rex.w)
+                X64_ROTL_D(s1, s2);
+            else
+                X64_ROTL_W(s1, s2);
+        }
+    }
```
ptitSeb, line 467: missing an else of something? that's strange s1 get rol'd if la64_lbt and IFX, and rol'd again before line 485 were it finaly return.

ksco: > Or is that just flags update and not the value itself?

Yes, LBT instructions prefixed with `X64_` only update flags. The design philosophy of LBT is to keep the RISC instructions RISCy and add additional x86-related simulations on top of it.

Issue #1450, opened 2024-04-17
A Qt program can't start


mio-19:
Debian package https://down.115.com/client/115pc/lin/115pc_2.0.9.3.deb AUR package https://aur.archlinux.org/packages/115pc
```
/opt/115pc/115
Dynarec for ARM64, with extension: ASIMD AES CRC32 PMULL ATOMICS SHA1 SHA2 USCAT FLAGM FLAGM2 FRINT PageSize:4096 Running on Cortex-A510 Cortex-A715 Cortex-A710 - with 8 Cores
Will use Hardware counter measured at 19.2 MHz emulating 614 MHz
Params database has 62 entries
Box64 with Dynarec v0.2.7 067cf75c built on Apr 15 2024 06:25:38
BOX64: Didn't detect 48bits of address space, considering it's 39bits
Counted 43 Env var
BOX64 LIB PATH: ./:lib/:lib64/:x86_64/:bin64/:libs64/:/lib/x86_64-linux-gnu/:/usr/lib/x86_64-linux-gnu/
BOX64 BIN PATH: ./:bin/:/home/user/.local/bin/:/sbin/:/bin/:/usr/bin/:/usr/local/bin/:/usr/local/sbin/:/usr/lib/jvm/default/bin/:/usr/bin/site_perl/:/usr/bin/vendor_perl/:/usr/bin/core_perl/
Looking for /opt/115pc/115
Rename process to "115"
Using emulated /opt/115pc/lib/libQt5Widgets.so.5
Using emulated /opt/115pc/lib/libQt5Gui.so.5
Using emulated /opt/115pc/lib/libQt5Network.so.5
Using emulated /opt/115pc/lib/libQt5Xml.so.5
Using emulated /opt/115pc/lib/libQt5Sql.so.5
Using emulated /opt/115pc/lib/libQt5DBus.so.5
Using emulated /opt/115pc/lib/libQt5Core.so.5
Using native(wrapped) libpthread.so.0
Using emulated /lib/x86_64-linux-gnu/libstdc++.so.6
Using emulated /lib/x86_64-linux-gnu/libgcc_s.so.1
Using native(wrapped) libc.so.6
Using native(wrapped) ld-linux-x86-64.so.2
Using native(wrapped) libdl.so.2
Using native(wrapped) libutil.so.1
Using native(wrapped) librt.so.1
Using native(wrapped) libbsd.so.0
Using native(wrapped) libm.so.6
Look for arc4random in loaded elfs
[...]

PR #1453, opened 2024-04-17
[DYNAREC] Fixed CF flag computation

ksco:
A very subtle error, the source register does not guarantee that the high bits are clean. Found while trying to run the program in #1450 on RV64.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_emit_math.c
+++ b/src/dynarec/rv64/dynarec_rv64_emit_math.c
@@ -45,7 +45,9 @@ void emit_add32(dynarec_rv64_t* dyn, int ninst, rex_t rex, int s1, int s2, int s
             BEQZ(s5, 8);
             ORI(xFlags, xFlags, 1 << F_CF);
         } else {
-            ADD(s5, s1, s2);
+            AND(s3, s1, xMASK);
+            AND(s4, s2, xMASK);
+            ADD(s5, s3, s4);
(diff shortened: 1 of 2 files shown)


src/dynarec/rv64/dynarec_rv64_emit_math.c
```diff
@@ -45,7 +45,9 @@ void emit_add32(dynarec_rv64_t* dyn, int ninst, rex_t rex, int s1, int s2, int s
             BEQZ(s5, 8);
             ORI(xFlags, xFlags, 1 << F_CF);
         } else {
-            ADD(s5, s1, s2);
+            AND(s3, s1, xMASK);
+            AND(s4, s2, xMASK);
+            ADD(s5, s3, s4);
```
ptitSeb, line 50: only 1 and with xMASK is needed actually...

ksco: hmm why?

ptitSeb: s3 will have upper bit to 0. anything and that will still have the upper bit to 0.

ksco: oh I think you misread line 50, it’s add, not and.

ptitSeb: ah right, add, not and :(

PR #1479, opened 2024-04-30
[LA64_DYNAREC] Fixed dynarec infra

ksco:


ptitSeb:
I have updated `fpuCacheTransform(...)` latelly, on arm64 (a few days ago) and rv64 (yesterday). Pretty sure it needs some changes on la64 too?

ksco:
Yeah, it's x87 stuff right? Will do it when adding x87 support.

PR #1518, opened 2024-05-22
[DYNAREC] Simplified emit_pf helper

ksco:


src/emu/x64emu.c
```diff
@@ -48,11 +48,6 @@ static uint32_t x86emu_parity_tab[8] =
 	0x69969669,
 };
 
-uint32_t* GetParityTab()
```
ptitSeb, line 51: The function is still used on LA64 backend I guess

ptitSeb: Ah no ok, rv64 and la64 are both done, my bad.

PR #1521, opened 2024-05-22
[LA64_DYNAREC] Fixed more issues here and there

ksco:
A Short Hike renders correctly.

### DIFF
--- a/src/dynarec/la64/dynarec_la64_660f.c
+++ b/src/dynarec/la64/dynarec_la64_660f.c
@@ -701,10 +701,15 @@ uintptr_t dynarec64_660F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int 
             if (MODREG) {
                 v1 = sse_get_reg_empty(dyn, ninst, x1, (nextop & 7) + (rex.b << 3));
                 q0 = fpu_get_scratch(dyn);
-                // v1 can be same as v0!
-                VXOR_V(q0, q0, q0);
-                VEXTRINS_D(q0, v0, 0);
-                VOR_V(v1, q0, q0);
+                if (v0 == v1) {
+                    // clear upper bits
+                    VXOR_V(q0, q0, q0);
+                    VEXTRINS_D(q0, v0, 0);
+                    VOR_V(v1, q0, q0);
+                } else {
+                    VXOR_V(v1, v1, v1);
+                    VEXTRINS_D(v1, v0, 0);
+                }
             } else {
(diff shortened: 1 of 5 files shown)


src/dynarec/la64/dynarec_la64_660f.c
```diff
@@ -700,8 +700,11 @@ uintptr_t dynarec64_660F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
             GETGX(v0, 0);
             if (MODREG) {
                 v1 = sse_get_reg_empty(dyn, ninst, x1, (nextop & 7) + (rex.b << 3));
-                VXOR_V(v1, v1, v1);
-                VEXTRINS_D(v1, v0, 0);
+                q0 = fpu_get_scratch(dyn);
```
ptitSeb, line 703: if ex==gx, than v0==v1, and the operationis a NOP...

src/dynarec/la64/dynarec_la64_f30f.c
```diff
@@ -230,7 +230,6 @@ uintptr_t dynarec64_F30F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
             INST_NAME("MOVQ Gx, Ex");
             nextop = F8;
             GETGX_empty(v0);
-            VXOR_V(v0, v0, v0);
             if (MODREG) {
                 v1 = sse_get_reg(dyn, ninst, x1, (nextop & 7) + (rex.b << 3), 0);
```
ptitSeb, line 234: This should not work in the case EX==GX, because GX is fetched as "empty" first, so EX (that is GX), will not be fetched becasue it's after the GETGX_empty(). It only work because GX has already been fetched in a previous opcode and so the cache is present (and GETGX_empty does nothing, just return the cached registred

src/dynarec/la64/dynarec_la64_660f.c
ksco: not really, the upper bits still need to be cleared.

src/dynarec/la64/dynarec_la64_f30f.c
ksco: Ah yes, so the `GETGX_empty` need to be put after `sse_get_reg`.

src/dynarec/la64/dynarec_la64_660f.c
ksco: anyway, if v0 != v1, 1 instruction can be saved, I pushed the optimization.

src/dynarec/la64/dynarec_la64_f30f.c
ksco: Fixed.

PR #1537, opened 2024-05-30
Added more MMX opcodes and some optimizations too

xctan:


xctan:
https://github.com/ptitSeb/box64/pull/1537/commits/39283ff77aee6c34031ad582b6c9f332cd5ee40c is a bit tricky, so I want to keep this PR small to ease future debugging.

PR #1538, opened 2024-05-30
Wrapped libssh2

zohanzephyr:


ptitSeb:
No need for wrapped function here? What's the use case?

zohanzephyr:
> No need for wrapped function here? What's the use case?

I test an application when used libssh2 library, which many functions are relocated, but the function of the real implementation of the library is wrapped so I have to wrap the libssh2 to be able to find the actual address of the function, the subsequent increase in functions I will add!

ptitSeb:
> > No need for wrapped function here? What's the use case?
> 
> I test an application when used libssh2 library, which many functions are relocated, but the function of the real implementation of the library is wrapped so I have to wrap the libssh2 to be able to find the actual address of the function, the subsequent increase in functions I will add!

you can do something like `nm -D /PATH/TO/libssh2.so.1 | grep " [DBTIW] "` to get the name of the function available in the lib to pre-fill the `_private.h`

zohanzephyr:
> > > No need for wrapped function here? What's the use case?
> > 
> > 
> > I test an application when used libssh2 library, which many functions are relocated, but the function of the real implementation of the library is wrapped so I have to wrap the libssh2 to be able to find the actual address of the function, the subsequent increase in functions I will add!
> 
> you can do something like `nm -D /PATH/TO/libssh2.so.1 | grep " [DBTIW] "` to get the name of the function available in the lib to pre-fill the `_private.h`

Okay, I'll add to that.

PR #1542, opened 2024-05-31
[RV64_DYNAREC] Added more MMX opcodes and some optimizations too

xctan:


src/dynarec/rv64/dynarec_rv64_0f.c
```diff
@@ -2182,12 +2314,93 @@ uintptr_t dynarec64_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             GETEM(x2, 0);
             MMX_LOOP_WS(x3, x4, MULW(x3, x3, x4));
             break;
+        case 0xD7:
```
ksco, line 2317: This looks so complicated, personally I would prefer simpler implementation for the scalar version.

src/dynarec/rv64/dynarec_rv64_660f.c
```diff
@@ -2755,13 +2755,59 @@ uintptr_t dynarec64_660F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
             GETEX(x2, 0);
             GETGD;
-            MV(gd, xZR);
-            for (int i = 0; i < 16; ++i) {
-                LB(x1, wback, fixedaddress + i);
-                SLT(x3, x1, xZR);
-                if (i > 0) SLLI(x3, x3, i);
-                OR(gd, gd, x3);
-            }
+            #define MYGO(s)            \
```
ptitSeb, line 2758: Is this an optimisation?

xctan: Yes. This instruction is computed in the reverse direction so we can make use of the SH1ADD instruction.

PR #1547, opened 2024-06-01
[INTERP] Rework on the 16b emulation for LA64

ksco:


ksco:
Oops, the compiler does not recognize `sc.q`, afk now I'll fix it later.

ksco:
@ptitSeb ping

PR #1552, opened 2024-06-02
Try to fix mmap64 already defined error in custommem.c

rtissera:
As you suggested, I added a barebones custommap.c file and it passes compilation and linking in my buildroot GCC 13.3 cross-compilation toolchain to arm64 / rv64.

It can probably be tidied up.


ksco (approved):
LGTM.

(Maybe we should upgrade the CI gcc version to 13.x or even 14.x.)

PR #1556, opened 2024-06-03
Develop

zohanzephyr:


ptitSeb:
This is a C++ library, and cannot be wrapped easily. There is no mecanism for now, in box, to handle vtable and overloadeing, so this wrapping will not work, unless it stays empty, but I assume it's not the goal...

zohanzephyr:
> This is a C++ library, and cannot be wrapped easily. There is no mecanism for now, in box, to handle vtable and overloadeing, so this wrapping will not work, unless it stays empty, but I assume it's not the goal...

Yes, but it will be used when other library functions are relocated, and the current state of keeping it empty can satisfy the use of it, is it possible to first wrapped, and then add the function at a later stage?

ksco:
> Yes, but it will be used when other library functions are relocated, and the current state of keeping it empty can satisfy the use of it, is it possible to first wrapped, and then add the function at a later stage?

hmmm, what's the point if it stays empty? makes no sense.

zohanzephyr:
> > Yes, but it will be used when other library functions are relocated, and the current state of keeping it empty can satisfy the use of it, is it possible to first wrapped, and then add the function at a later stage?
> 
> hmmm, what's the point if it stays empty? makes no sense.

Screenshot from 2024-06-03 18-04-58
It solves the problem of function relocation

ksco:
maybe you misunderstood the error message? for example the symbol `g_bytes_new_with_free_func` does not come from libvips-cpp, it's the caller, not callee.

ptitSeb:
All this doesn't changed the fact that c++ lib cannot be wrapped for now. I cannot accept this PR.

It would be more constructive too understand why there is a SEGFAUT when using wrapped glibc.

PR #1561, opened 2024-06-05
[RV64_DYNAREC] Fixed some opcodes caught by cosim

xctan:


src/dynarec/rv64/dynarec_rv64_emit_shift.c
```diff
@@ -1429,6 +1435,9 @@ void emit_shld32(dynarec_rv64_t* dyn, int ninst, rex_t rex, int s1, int s2, int
     } else IFX(X_ALL) {
         SET_DFNONE();
     }
+
+    BEQZ_MARK(s5);
```
ksco, line 1439: It seems for now `emit_shld32` only gets called once in `SHLD Ed, Gd, CL`, and there is already a `BEQZ_NEXT` on s5 before the function call.

ptitSeb:
So it's ready to merge?

src/dynarec/rv64/dynarec_rv64_00_2.c
```diff
@@ -547,7 +547,7 @@ uintptr_t dynarec64_00_2(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
         case 0x9F:
             INST_NAME("LAHF");
             READFLAGS(X_CF | X_PF | X_AF | X_ZF | X_SF);
-            ANDI(x1, xFlags, 0xFF);
```
ptitSeb, line 550: well, that OF2 flags doesn't seems handled in this opcode?

xctan: OF2 has taken the place of res3 but it is not read here, so the mask below will strip the extra OF2, as I confirmed in the cosim.

xctan:
> So it's ready to merge?

Yup. I cannot find any other prominent differences between dynarec and the interpreter when running VMP with cosim.

src/dynarec/rv64/dynarec_rv64_00_2.c
ptitSeb: ah yes, OF is to high to get there. All good then :)

PR #1572, opened 2024-06-09
Update README.md

Pro42good:
Usage link was broken so i fixed it, Usage.md is under docs, not under main

Previously:
```ptitSeb/box64/blob/main/USAGE.md``` 

Now
```ptitSeb/box64/blob/main/docs/USAGE.md```


ptitSeb:
Thanks.

PR #1573, opened 2024-06-09
Update README.md

Pro42good:
Added Strong memory mode to the hint under unity games because freezing is also a common issue on PI-4 and PI-5 for games, also added the mention of PI-5 on the hint because the issues aren't limited to the PI-4 and also apply to the Pi-5.

Just quality of life changes to the README.md


Pro42good:
I also fixed grammar there

PR #1574, opened 2024-06-10
Quality of life note and grammar fix

Pro42good:
I added a singular comma.
Added a note on the vulkan section informing most people that use raspberry pi os that vulkan drivers don't come pre-installed with the OS to stop unnecessary issues being created, should fix confusion of why the graphics drivers aren't working, games aren't being launched, and why open GL core isn't starting.

Side Note:
Yes you need to install the vulkan drivers for the newest openGL to be supported on Raspbian and also to fix open gl core not being found


Pro42good:
Issues with the pull Request, had to back out

Issue #1581, opened 2024-06-12
Wine's Wayland driver does not load


strobo5:
I'm trying to run wine (wine-9.9) on aarch64, in a Ubuntu 24.04 userspace within Docker since the main OS uses muslibc. Wine was compiled manually on a Ubuntu 24.04 x86_64 machine with Wayland and syswow64 support:
```
./configure --enable-archs=i386,x86_64 --enable-win64 --with-wayland=yes
```
I enable Wayland like this:
```
wine reg.exe add HKCU\\Software\\Wine\\Drivers /v Graphics /d wayland,x11
```
I can start `winecfg`, it works by falling back to X11, but when force it to use Wayland by having an empty `DISPLAY` environment variable it fails to start.

I created a log with `WAYLAND_DEBUG=1 WINEDEBUG=+loaddll,+module,+imports,+waylanddrv,+x11drv` and attached the part of it where `winewayland.drv` is being loaded. Unfortunately I don't have an understanding of what is happening here under the hood. Should I not see a line like this somewhere?
```
Using native(wrapped) libwayland-client.so.0
```

[wine-9.9_winecfg_WINEDEBUG_loaddll_module_imports_waylanddrv.log](https://github.com/user-attachments/files/15811823/wine-9.9_winecfg_WINEDEBUG_loaddll_module_imports_waylanddrv.log)

Some notes:
- aarch64 Wayland apps work within the Docker container (tested: `foot`, [hello-wayland](https://github.com/emersion/hello-wayland))
- the same "hello-wayland", but compiled on x86_64 and run through `box64`, crashes with a `SIGBUS` exception (another bug?)
- in `gdb` I can do `file box64` and `run wine winecfg` but I have no idea what to do here lol
- the wine Wayland driver _does_ work on the x86_64 system where I built wine

Attaching more uncut gzipped logs because it's free, but they are large:
[winecfg_WINEDEBUG_waylanddrv_loaddll_BOX64_LOG_3.log.gz](https://github.com/user-attachments/files/15811946/winecfg_WINEDEBUG_waylanddrv_loaddll_BOX64_LOG_3.log.gz)
[winecfg_WINEDEBUG_waylanddrv_loaddll_module_imports.log.gz](https://github.com/user-attachments/files/15811947/winecfg_WINEDEBUG_waylanddrv_loaddll_module_imports.log.gz)
[...]

PR #1599, opened 2024-06-18
Add missing wrappings for winewayland.so

strobo5:
This was a late-night attempt to fix issue #1581.

I still don't really know how the wrappings work (I should take time and learn) so I just did what seemed right. For `size_t` and `bool` I guessed how they are specified ('L' and 'c'??).

Unsuprisingly, it does not work at this point! Now there is a SIGILL during/after mapping winewayland.so.

### DIFF
--- a/src/wrapped/wrappedwaylandclient.c
+++ b/src/wrapped/wrappedwaylandclient.c
@@ -475,6 +475,57 @@ static void* find_zwp_relative_pointer_v1_listener_Fct(void* fct)
     printf_log(LOG_NONE, "Warning, no more slot for wayland-client zwp_relative_pointer_v1_listener callback\n");
     return NULL;
 }
+// zxdg_output_v1_listener ...
+typedef struct my_zxdg_output_v1_listener_s {
+    uintptr_t   logical_position; //vFppii
+    uintptr_t   logical_size; //vFppii
+    uintptr_t   done; //vFpp
+    uintptr_t   name; //vFppp
+    uintptr_t   description; //vFppp
+} my_zxdg_output_v1_listener_t;
+#define GO(A)   \
+static my_zxdg_output_v1_listener_t* ref_zxdg_output_v1_listener_##A = NULL;                                              \
+static void* my_zxdg_output_v1_listener_logical_position_##A(void* a, void* b, int32_t c, int32_t d)  \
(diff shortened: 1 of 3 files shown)


ptitSeb:
> This was a late-night attempt to fix issue #1581.
> 
> I still don't really know how the wrappings work (I should take time and learn) so I just did what seemed right. For `size_t` and `bool` I guessed how they are specified ('L' and 'c'??).
> 
> Unsuprisingly, it does not work at this point! Now there is a SIGILL during/after mapping winewayland.so.

For bool, I would use `u` instead. `c` would be signed char, I don't think it's the best approach (`C` unsigned char would be better).

For the SIGILL, it might be something else. Check if you have some `BOX64: Error, Wayland-client, add_listener to XXXX unknown, will crash soon!` in the log, that might be just that.

strobo5:
Alriight, thanks for checking it! I'm changing the bool to `u` then.

Also I found exactly that error message. I just had to set a GDB breakpoint on that line since I still can't get the log messages printed on my terminal.

Now I'm trying to add the missing `zxdg_output_v1` listener proxy thing. It's not working yet, I'll get back later..

ptitSeb:
Listenner are more tricky to implement. if you don't have enough example, tell me, and I'll add it.

strobo5:
That would be great I think. This was my attempt and I don't see any obvious problems without a bit more understanding about the mechanisms.

ptitSeb:
According to https://searchfox.org/mozilla-central/source/widget/gtk/wayland/xdg-output-unstable-v1-client-protocol.h#191 you are missing "done", vFpp, between "logical_size" and "name"

strobo5:
Haha yes, this was it. I had seen it [here](https://wayland.app/protocols/xdg-output-unstable-v1#zxdg_output_v1:event:done), read something about "deprecated", did not see it in the winewayland.so symbols and decided to omit it.

Of course this is not how box64 works, and this Wayland listener interface..
Now I know. Sort of.

The result is that I can now start `regedit` and `winecfg`. Now I will test some games.

ptitSeb:
thanks, and good work wrapping the listener ;)

strobo5:
only possible with your quick feedback! thanks, this was fun

Issue #1576, opened 2024-06-10
Precompiled builds no longer work. 


rom4ster:
The instructions for ryanfortner precomp builds result in "Unable to locate package". 

Probably has to do with him redirecting his domain or something. 

Could I suggest making a pipeline or something for precompiled builds. I really dont want to have to compile anything on my raspberry pi because it takes forever.

ptitSeb:
Cab you open a ticket at ryanfortne repo about this?

About the pipeline, you want to do a PR bout it?

rom4ster:
I dont know how. I cant even get this to work on my local PI .

m1m1k4tz:
It is kinda weird how the artifacts don’t get uploaded I checked the release.yml and I can’t figure out why. Unable to locate package sounds like an apt issue though are you using the correct distribution and are 32 bit packages enabled?

PR #1605, opened 2024-06-19
switch to maintained Pi-Apps-Coders apt repo

theofficialgman:
closes https://github.com/ptitSeb/box64/issues/1576

at pi-apps we have been using our own apt repo which is better maintained and we keep strict rules of not breaking old installs (eg: through package renaming or otherwise). this has been the case for a year now. Our efforts (including both Ryan and Itai) are consolidated in these repos that we all have access to.

### DIFF
--- a/docs/COMPILE.md
+++ b/docs/COMPILE.md
@@ -2,12 +2,34 @@ Compiling/Installing
 ----
 
 ### Debian-based Linux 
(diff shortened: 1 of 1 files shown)


ptitSeb:
are there some hardware specific build also? or just the generic-arm version?

theofficialgman:
> are there some hardware specific build also? or just the generic-arm version?

some hardware specific builds are there. if the user wants to install one of them they can click on the link to see the list

PR #1616, opened 2024-06-25
[RV64_DYNAREC] Fixed regression on D9 E5 FXAM opcode

ksco:


src/dynarec/rv64/dynarec_rv64_d9.c
```diff
@@ -186,7 +192,6 @@ uintptr_t dynarec64_D9(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             MARK3;
             // Extract signa & Update SW
             SRLI(x1, x2, 63);
-            ANDI(x4, x4, ~(1<<9));
```
ptitSeb, line 189: the the bit 9 needs to be cleared beforre the OR

ptitSeb: or is it already 0 no mater what?

ksco: That bit in x4 are always zero because it’ constructed manually above.

PR #1619, opened 2024-06-26
[RV64_DYNAREC] Detect vector extension

ksco:
Dynarec for RISC-V With extension: I M A F D C Zba Zbb Zbc Zbs Vector PageSize:4096 Running on Spacemit(R) X60 with 8 Cores

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -108,6 +108,7 @@ int rv64_zba = 0;
 int rv64_zbb = 0;
 int rv64_zbc = 0;
 int rv64_zbs = 0;
+int rv64_vector = 0;
 int rv64_xtheadba = 0;
 int rv64_xtheadbb = 0;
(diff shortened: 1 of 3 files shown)


ptitSeb:
We might want to check if it's v0.7 or v1.0 extension?

ksco:
> We might want to check if it's v0.7 or v1.0 extension?

Yeah this checks, there is no vcsr in v0.7.

ptitSeb:
ok, but what if we want to support v0.7? there is a lot of harsware out there with 0.7

ksco:
I'm thinking of adding v1.0 first, then adapting the code to v0.7.1 (or xtheadvector for its new name) if necessary.

ptitSeb:
Ok, I wont be abble to help you there, I don't have any v1.0 copliant hardware.

PR #1631, opened 2024-07-03
[RV64_DYNAREC] Added DF default /4 opcode

zohanzephyr:


src/dynarec/rv64/dynarec_rv64_df.c
```diff
@@ -179,6 +179,33 @@ uintptr_t dynarec64_DF(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                     SH(x4, wback, fixedaddress);
                     X87_POP_OR_FAIL(dyn, ninst, x3);
                     break;
+		case 4:
```
ksco, line 182: Please do not use tabs.

PR #1632, opened 2024-07-03
[RV64_DYNAREC] Added preliminary RVV infra and PXOR opcode for demonstration

ksco:


src/dynarec/rv64/dynarec_rv64_66.c
```diff
@@ -125,7 +125,13 @@ uintptr_t dynarec64_66(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             break;
         case 0x0F:
             switch(rep) {
-                case 0: addr = dynarec64_660F(dyn, addr, ip, ninst, rex, ok, need_epilog); break;
+                case 0: {
+                    if (rv64_vector)
+                        addr = dynarec64_660F_vector(dyn, addr, ip, ninst, rex, ok, need_epilog);
```
ptitSeb, line 130: That means we need to duplicate ALL the opcode in _vector and regular... Why not having regular version as a fallback if it's implemented in the _vector one instead?

src/dynarec/rv64/dynarec_rv64_private.h
```diff
@@ -18,20 +18,28 @@ typedef struct instsize_s instsize_t;
...
     };
 } ext_cache_t;
+
 typedef union sse_cache_s {
-    int8_t      v;
+    int16_t v;
     struct {
-        uint8_t     reg:7;
-        uint8_t     single:1;
+        uint16_t reg : 7;
```
ptitSeb, line 35: 7 bits for reg? so a number between 0 and 127? that seems a bit too much.

src/dynarec/rv64/dynarec_rv64_66.c
ksco: mmm not sure what you mean, this PR does have the regular version as fallback.

src/dynarec/rv64/dynarec_rv64_660f_vector.c
```diff
@@ -0,0 +1,71 @@
...
+    MAYUSE(q1);
+    MAYUSE(eb1);
+    MAYUSE(eb2);
+    MAYUSE(j64);
+    switch (opcode) {
+        case 0xEF:
+            INST_NAME("PXOR Gx, Ex");
+            nextop = F8;
+            // FIXME: we should try to minimize vsetvl usage as it may hurts performance a lot.
+            vector_vsetvl_emul1(dyn, ninst, x1, VECTOR_SEW8);
```
ptitSeb, line 53: xor (and all other logical operations) doesn't care about the size of the vector element, so setting it up is probably not usefull.

And yeah, this info needs to be tracked, like the defered flags status, to avoid changing at each opcodes

src/dynarec/rv64/dynarec_rv64_helper.c
```diff
@@ -1837,16 +1950,35 @@ static int findCacheSlot(dynarec_rv64_t* dyn, int ninst, int t, int n, extcache_
...
+            MESSAGE(LOG_DUMP, "\t  - Moving %d <- %d\n", i, j);
+            VOR_VV(i, j, j, VECTOR_UNMASKED);
+            cache->extcache[i].v = cache->extcache[j].v;
+            cache->extcache[j].v = 0;
+            return;
+        }
+        // SWAP
+        ext_cache_t tmp;
+        MESSAGE(LOG_DUMP, "\t  - Swapping %d <-> %d\n", i, j);
+#define SCRATCH 31
```
ptitSeb, line 1967: Try to use XOR to swap the register, without using a scratch regs, it might be usefull for AVX implementation later for hardware that don't support 256bits lenght vectors

src/dynarec/rv64/dynarec_rv64_660f_vector.c
ksco: yes you are right, but we need to call the vsetvli at least once, to keep the vpu in a good state so the following vector instructions do not sigill.

src/dynarec/rv64/dynarec_rv64_66.c
ptitSeb: What happens if the program have 66 0F 5B opcode for example? for now, it will be run by the interpreter, not the dynarec.

src/dynarec/rv64/dynarec_rv64_private.h
ksco: right, there are 16 registers, so 5 bits is enough (an extra bit for not conflict with the special -1), we can keep it in a uint8_t.

src/dynarec/rv64/dynarec_rv64_66.c
ksco: it will not, see the end of `dynarec64_660F_vector()`, called `dynarec64_660F` as fallback there. maybe you have a better way to do the fallback in mind?

src/dynarec/rv64/dynarec_rv64_helper.c
ksco: Done.

src/dynarec/rv64/dynarec_rv64_66.c
ptitSeb: Oh, ok. but I don't think it's the best place, because there can be many "default" there....

src/dynarec/rv64/dynarec_rv64_660f_vector.c
ptitSeb: Ok. That can be enough for now... but in the future, when this is info is tracked, logical operation should use waterver size is already setup...

ksco: yeah, we may have element width requirements like 8, 16, 32, 64, and agnostic, the last one means the opcode does not care.

src/dynarec/rv64/dynarec_rv64_private.h
```diff
@@ -18,20 +18,27 @@ typedef struct instsize_s instsize_t;
...
     };
 } ext_cache_t;
+
 typedef union sse_cache_s {
-    int8_t      v;
+    int8_t v;
     struct {
-        uint8_t     reg:7;
-        uint8_t     single:1;
+        uint8_t reg:5;
```
ptitSeb, line 35: If there is only 16 regs then we have a problem, and cannot use fixed register position as for arm64, because that would leave 0 scratch register!
We need sommething more dynamic, like YMM handling on ARM64, because we need more than 16regs!

src/dynarec/rv64/dynarec_rv64_66.c
ksco: so you have a better place? maybe in here, and use some pointer variable to indicate if it needs to fallback?

src/dynarec/rv64/dynarec_rv64_private.h
ksco: wait... it stores the native register number, but the sse register number, then there is 32 of them, and we need a uint16_t.

ksco: I reverted the change, sorry.

ptitSeb: reverted partly... 5bits should be enough yes.

src/dynarec/rv64/dynarec_rv64_66.c
ptitSeb: yes, here is a better place.
there is already a way to return if the opcode was not handled, with `ok`... But you can also return 0 instead of addr when opcode is not handled, and have here the return value stored in a temp value, if it's not 0, return temp, else, falback to regular version.

src/dynarec/rv64/dynarec_rv64_private.h
ksco: 5 bits are not enough I think, because we use -1 to indicate the register is forgotten, so there needs at least 1 bit free to not conflict with -1.

src/dynarec/rv64/dynarec_rv64_66.c
ksco: Done.

ksco:
@ptitSeb please do another round of review, thanks.

ptitSeb (approved):

src/dynarec/rv64/dynarec_rv64_private.h
ksco: sorry not sure I followed, but what you suggested on "revert partly" is this?

```
       uint16_t reg:5;
        uint16_t vector:1;
        uint16_t single:1;
        uint16_t write:1;
        uint16_t isfree:1;
        uint16_t unused:7;
```

ptitSeb: Yes. And this `isfree` need to be set to 0 when used...

PR #1641, opened 2024-07-04
[VULKAN] Add Wrapped VK_KHR_android_surface

KreitinnSoftware:


KreitinnSoftware:
I had tested only on Android build, later I will try to fix build to another builds

ptitSeb:
I think you are simply missing the generated files in your commit.

KreitinnSoftware:
> I think you are simply missing the generated files in your commit.

Probably

PR #1643, opened 2024-07-04
[VULKAN] Add Wrapped VK_KHR_android_surface

KreitinnSoftware:


rajdakin:
@KreitinnSoftware Just so you are aware, you don't have to close old PR and open new ones with the same content except with one more change, since 1. ptitSeb uses the "squash and merge pull request" functionality of GitHub (which merges all commits into one before adding it to the main branch), and 2. even if you wanted to only have a single commit in a PR, you can just force-push into the branch you have selected for merging and the PR is updated automatically (`git push --force`, as told by the error message).

PR #1644, opened 2024-07-05
Should not follow symlinks on `argv[0]` path resolution

Coekjan:
Many programs (e.g. busybox, ex, bash) can have different behaviors on different `argv[0]`. We should only resolve the `argv[0]` (program name)  to its absolute path but not realpath, i.e. don't resolve the symlinks.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1930,7 +1930,7 @@ int initialize(int argc, const char **argv, char** env, x64emu_t** emulator, elf
     PrintCollection(&my_context->box64_path, "BOX64 BIN PATH");
     // lets build argc/argv stuff
     printf_log(LOG_INFO, "Looking for %s\n", prog);
-    my_context->argv[0] = ResolveFile(prog, &my_context->box64_path);
+    my_context->argv[0] = ResolveFileSoft(prog, &my_context->box64_path);
     // check if box86 is present
     {
         my_context->box86path = box_strdup(my_context->box64path);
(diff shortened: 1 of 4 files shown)


ptitSeb:
> Many programs (e.g. busybox, ex, bash) can have different behaviors on different `argv[0]`. We should only resolve the `argv[0]` (program name) to its absolute path but not realpath, i.e. don't resolve the symlinks.

Ok, I agree. But I'm afraid adding it here will have some side effect on other things. This function is used in many places, not just arg[0] resolution.

Coekjan:
> Ok, I agree. But I'm afraid making the change here will have some side effect on other things. This function is used in many places, not just arg[0] resolution.

Maybe we can have a "soft" version of `ResolveFile`? Like `ResolveFileSoft`, indicating that it does not resolve symlinks.

ptitSeb:
> > Ok, I agree. But I'm afraid making the change here will have some side effect on other things. This function is used in many places, not just arg[0] resolution.
> 
> Maybe we can have a "soft" version of `ResolveFile`? Like `ResolveFileSoft`, indicating that it does not resolve symlinks.

Yeah, that would limit side-effect I think.

Coekjan:
> Yeah, that would limit side-effect I think.

Done :) Please review the changes.

ptitSeb:
Ok, looks good now, thanks.

PR #1653, opened 2024-07-08
Mask `rs2` when using `bext` instruction if `rex.w` is not set

### LINKED ISSUE #1652: Unexpected Behaviors when Using RISC-V Extension "zbs"
### Description & Investigation

I tried to run python3.12 on box64, and it crashed with SIGSEGV. Using `BOX64_DYNAREC_TEST=1`, I saw many differences about RIP:

```txt
Warning, difference between x64 Interpreter and Dynarec in 0x3f001e44b1 (0f 82 82 03 00 00 8d b0)
=======================================
DIFF: Dynarec |  Interpreter
----------------------
RIP: 0000003f001e44b7 | 0000003f001e4839
Warning, difference between x64 Interpreter and Dynarec in 0x3f001e4405 (0f 82 2d 01 00 00 83 ea)
=======================================
DIFF: Dynarec |  Interpreter
----------------------
RIP: 0000003f001e440b | 0000003f001e4538
Warning, difference between x64 Interpreter and Dynarec in 0x3f002683c7 (72 2b 2d 00 01 00 00 83)
=======================================
DIFF: Dynarec |  Interpreter
(( omitted ))
```

These addresses are all in `libpython3.12.so` and `objdump` told me **they all almost look like**:

```asm
bt     %eax,%ecx
jb     (( somewhere ))
```

The `bt` instruction should set the CF flag and the behavior of the following `jb` instruction depends on it. So I read current code:

https://github.com/ptitSeb/box64/blob/d6b9fd2860dc097de8aebf147b3aa6d1741ea550/src/dynarec/rv64/dynarec_rv64_0f.c#L1720-L1739

and saw `BEXT` will emit to `bext` instruction when "zbs" extension enabled.

---

When turning off the "zbs" extension, `libpython3.12` still crashed but the test log showed other differences irrelative with `bt` instruction:

```txt
(( BEGIN ))
Warning, difference between x64 Interpreter and Dynarec in 0x3f001ea3f3 (7e 17 83 fe 63 7f 12 89)
=======================================
DIFF: Dynarec |  Interpreter
----------------------
RIP: 0000003f001ea40c | 0000003f001ea3f5
Warning, difference between x64 Interpreter and Dynarec in 0x3f001ea36f (0f 8f bb 00 00 00 39 c3)
=======================================
DIFF: Dynarec |  Interpreter
----------------------
RIP: 0000003f001ea430 | 0000003f001ea375
Warning, difference between x64 Interpreter and Dynarec in 0x3f001ea3f3 (7e 17 83 fe 63 7f 12 89)
=======================================
DIFF: Dynarec |  Interpreter
----------------------
RIP: 0000003f001ea40c | 0000003f001ea3f5
Warning, difference between x64 Interpreter and Dynarec in 0x3f001ea36f (0f 8f bb 00 00 00 39 c3)
=======================================
DIFF: Dynarec |  Interpreter
----------------------
RIP: 0000003f001ea430 | 0000003f001ea375
(( END ))
```

Again, `objdump` told me **they all look like**:

```asm
cmp    (( %esi or $0x63 here )),%eax
jle/jg    (( somewhere ))
```

---

Maybe at least we can say the `bt` instruction with "zbs" extension results in the my first "dynarec test" difference of this issue. And the latter one is still in investigation.

Coekjan:
Fixes #1652 .

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_660f.c
+++ b/src/dynarec/rv64/dynarec_rv64_660f.c
@@ -2463,9 +2463,14 @@ uintptr_t dynarec64_660F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int 
                 ed = x1;
                 wback = x3;
             }
-            if (rv64_zbs)
-                BEXT_(x4, ed, gd);
-            else {
+            if (rv64_zbs) {
+                if (rex.w) {
+                    BEXT_(x4, ed, gd);
+                } else {
+                    ANDI(x2, gd, 0xf);
+                    BEXT_(x4, ed, x2);
+                }
+            } else {
                 ANDI(x2, gd, rex.w ? 0x3f : 0xf);
                 SRL(x4, ed, x2);
                 ANDI(x4, x4, 1);
--- a/src/dynarec/rv64/rv64_emitter.h
+++ b/src/dynarec/rv64/rv64_emitter.h
@@ -965,11 +965,16 @@ f28–31  ft8–11  FP temporaries                  Caller
 // Single-bit Set (Immediate)
 #define BSETI(rd, rs1, imm) EMIT(R_type(0b0010100, imm, rs1, 0b001, rd, 0b0010011))
 
-// Single-bit Extract (Register), s0 can be the same as rs2
+// Single-bit Extract (Register)
 #define BEXT(rd, rs1, rs2, s0)              \
-    if (rv64_zbs)                           \
-        BEXT_(rd, rs1, rs2);                \
-    else {                                  \
+    if (rv64_zbs) {                         \
+        if (rex.w) {                        \
+            BEXT_(rd, rs1, rs2);            \
+        } else {                            \
+            ANDI(s0, rs2, 0x1f);            \
+            BEXT_(rd, rs1, s0);             \
+        }                                   \
+    } else {                                \
         ANDI(s0, rs2, rex.w ? 0x3f : 0x1f); \
         SRL(rd, rs1, s0);                   \
         ANDI(rd, rd, 1);                    \


ptitSeb:
Mmmm, is that a fix or a workaround?

doesn't BEXT works on 32bits input?

Coekjan:
`bext` on rv64 does not handle 32bits input. This pseudo code describes how it works:

```txt
let index = X(rs2) & (XLEN - 1);
X(rd) = (X(rs1) >> index) & 1;
```

In the first step, `X(rs2) & (XLEN - 1)` is actually `X(rs2) & (64 - 1)` in rv64. When `rex.w` is not set, what we need to do is `X(rs2) & (32 - 1)` instead.

ptitSeb:
Then why not using the 's0' scratch reg to do the mask and still use BEXT, that would still be 1 opcode less then the fallback?

Coekjan:
> Then why not using the 's0' scratch reg to do the mask and still use BEXT, that would still be 1 opcode less then the fallback?

This can be better. I will push a new version and then please review the changes again. Thanks!

src/dynarec/rv64/dynarec_rv64_660f.c
```diff
@@ -2463,9 +2463,14 @@ uintptr_t dynarec64_660F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
-                BEXT_(x4, ed, gd);
-            else {
+            if (rv64_zbs) {
+                if (rex.w) {
+                    BEXT_(x4, ed, gd);
+                } else {
+                    ANDI(x2, gd, 0x1f);
+                    BEXT_(x4, ed, x2);
+                }
+            } else {
```
ksco, line 2473: Maybe replace this with the `BEXT` macro?

src/dynarec/rv64/dynarec_rv64_660f.c
```diff
@@ -2463,9 +2463,14 @@ uintptr_t dynarec64_660F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
                 ANDI(x2, gd, rex.w ? 0x3f : 0xf);
```
Coekjan, line 2474: @ksco I don't know why here it is `ANDI(x2, gd, rex.w ? 0x3f : 0xf);` but not `ANDI(x2, gd, rex.w ? 0x3f : 0x1f);` (difference on the mask). So I did not replace it before.

ksco: Oh it must be a terrible typo 😄

Coekjan: Okay. Let's fix it in this PR.

src/dynarec/rv64/rv64_emitter.h
```diff
@@ -967,9 +967,14 @@ f28–31  ft8–11  FP temporaries                  Caller
 
 // Single-bit Extract (Register), s0 can be the same as rs2
```
ksco, line 968: I think the `s0 can be the same as rs2` statement can be a bit confusing, maybe remove it.

ksco (approved):

src/dynarec/rv64/rv64_emitter.h
Coekjan: Done

src/dynarec/rv64/dynarec_rv64_660f.c
```diff
@@ -2463,13 +2463,7 @@ uintptr_t dynarec64_660F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
                 ed = x1;
                 wback = x3;
             }
-            if (rv64_zbs)
-                BEXT_(x4, ed, gd);
```
ptitSeb, line 2467: Sorry, but the BEXT macro cannot be used because this is the "66" version of the opcode, so operating on 16bits register. The non-rex.w mask is 0xf and not 0x1f here.

ksco: Oh, I'm sorry.... then:

```
            if (rv64_zbs) {
                if (rex.w) {
                    BEXT_(x4, ed, gd);
                } else {
                    ANDI(x2, gd, 0xf);
                    BEXT_(x4, ed, x2);
                }
            } else {
                ANDI(x2, gd, rex.w ? 0x3f : 0xf);
                SRL(x4, ed, x2);
                ANDI(x4, x4, 1);
            }
```

Coekjan: Fixed in the newest code. Thanks!

PR #1658, opened 2024-07-08
[LA64_DYNAREC] Fixed ADC/SBC usage

ksco:


src/dynarec/la64/dynarec_la64_emit_math.c
```diff
@@ -1038,17 +1038,17 @@ void emit_adc32(dynarec_la64_t* dyn, int ninst, rex_t rex, int s1, int s2, int s
...
+            ADC_W(s3, s1, s2);
 
         IFX (X_ALL) {
             if (rex.w)
                 X64_ADC_D(s1, s2);
             else
                 X64_ADC_W(s1, s2);
         }
-
+        MV(s1, s3);
```
ptitSeb, line 1051: I would duplicate the code to avoid the MV in the case there is no flags... but that's fine.

PR #1659, opened 2024-07-08
Fixes

rajdakin:
This PR fixes most warnings that I could find (leaving out those which are either false positive or too bothersome to fix while being minor). The first commit also changes some tabs into spaces.


ksco (approved):
I checked rv64 and la64 parts, look good, thanks! Also maybe we could add some more strict compiler flags to the CMakeLists.txt, at least optional? Would be great to catch these kinds of warnings/errors when writing the code.

Coekjan:
> Also maybe we could add some more strict compiler flags to the CMakeLists.txt, at least optional? Would be great to catch these kinds of warnings/errors when writing the code.

Agree. Something like `-Werror`, `-Wall`. And for code-style, I think we could check the code style in CI (we have `.clang-format` already).

ksco:
About clang-format, please refer to #1387.

PR #1666, opened 2024-07-09
optimize avx

salt431:
consolidate the conditionals in /src/emu/x64runavxf20f38.c along with a slight refactor. should boost performance a little bit, builds passing.


ptitSeb:
I'm sorry, but this is is literaly the exact same code with sligt style variation. There iis no optimisation there at all.

PR #1670, opened 2024-07-11
[RV64_DYNAREC] Cleanup for removed rv64_lock_cas_dq

iv-m:
Since rv64_lock_cas_dq is removed, we should also remove its global symbol declaration and its prototype. Otherwise, we get and undefined symbol reference, which breaks ELF checks that some distros employ.

Fixes: 9dbd7fc7e0615f48c9287305610bf9aed1cc24f3


iv-m:
@ksco @ptitSeb please take a look.

iv-m:
@ptitSeb , thank you)

PR #1671, opened 2024-07-12
[Wrapped] Added sqlite3_exec

kaixindeken:
This adds the `sqlite3_exec` function.


ksco:
This is not correct. sqlite_exec takes callback parameter.

kaixindeken:
> This is not correct. sqlite_exec takes callback parameter.

you are right, my bad.

PR #1674, opened 2024-07-12
Wrap `sqlite3_exec` for libsqlite3

Coekjan:


Coekjan:
I imitated the callback function processing methods in other wrapped libraries.

Please review my changes. If I am wrong, kindly let me know.

ksco:
IIRC wpp uses this function, can you verify if that application works with wrapped libsqlite3 now?

Coekjan:
> IIRC wpp (wps presentation) uses this function, can you verify if that application works with wrapped libsqlite3 now? btw PR itself looks good to me.

Seems that WPP works good.

ksco (approved):

src/wrapped/wrappedlibsqlite3.c
```diff
@@ -25,8 +25,40 @@ const char* libsqlite3Name = "libsqlite3.so.0";
...
+GO(0)   \
+GO(1)   \
+GO(2)   \
+GO(3)   \
+GO(4)
+
+// sqlite3_exec ...
+#define GO(A) \
+static uintptr_t my_sqlite3_exec_fct_##A = 0; \
+static int* my_sqlite3_exec_##A(void* a, int b, char** c, char** d) { \
```
ptitSeb, line 40: Type of the function is
```C
int sqlite3_exec(
  sqlite3*,                                  /* An open database */
  const char *sql,                           /* SQL to be evaluated */
  int (*callback)(void*,int,char**,char**),  /* Callback function */
  void *,                                    /* 1st argument to callback */
  char **errmsg                              /* Error msg written here */
);
```
return type if the callback is `int` not `int*`

ptitSeb (changes requested):

src/wrapped/wrappedlibsqlite3.c
Coekjan: Sorry for my mistake. I have fixed it in my newest commit.

ptitSeb (approved):

PR #1694, opened 2024-07-18
[RV64_DYNAREC] Added vlen detection

ksco:
We require VLEN >= 128 for the convenience of implementing SSE 2/3/4 opcodes, this PR checks it.


ptitSeb:
Note that if vlen is not big enough, it will silently disable RVV without any message...

PR #1697, opened 2024-07-18
[RV64_DYNAREC] Added 66 0F 38 00 PSHUFB for vector

ksco:
Look, we can shuffle in RISC-V!


ptitSeb:
Nice.
It will be time to cache the vector size/element size definition now, don't you think?

ksco:
Yes, it's next, but I'm going to switch to something else, so it should be tomorrow.

PR #1705, opened 2024-07-22
[RV64_DYNAREC] Fixed vector infra

ksco:
For vlen>128, we cannot use load/store whole vector instructions! Sorry for this mistake. This PR replaced all the whole vector load/stores with the normal ones, which requires the `vtype` to be set correctly using `SET_ELEMENT_WIDTH()`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,7 +45,7 @@ jobs:
           - platform: TERMUX
             type: StaticBuild
           - platform: X64
-            type: StaticBuild  
+            type: StaticBuild
 
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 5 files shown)


ksco:
Related to this PR:

I think there are certain limitations in the current vector infrastructure -- mainly for opcodes that are essentially element-width-agnostic.

For example, `MOVDQA` can be implemented with `vle8.v`, `vle16.v`, `vle32.v`, and `vle64.v`, it will be good if we know the predecessor's element width in compile time to choose the best one.

ptitSeb:
> Related to this PR:
> 
> I think there are certain limitations in the current vector infrastructure -- mainly for opcodes that are essentially element-width-agnostic.
> 
> For example, `MOVDQA` can be implemented with `vle8.v`, `vle16.v`, `vle32.v`, and `vle64.v`, it will be good if we know the predecessor's element width in compile time to choose the best one.
> 
> Previously I implemented `MOVDQA` with "load whole vector", but that unfortunately only works on vlen=128.

You can use a ecanism similar to float/i64/double of Float, but SSE handling is quite different, so an other mecanism should be used.

PR #1717, opened 2024-08-08
[INTERP] Fixed some undefined behaviour

ksco:
It's probably just fine, but just in case...

### DIFF
--- a/src/emu/x64primop.h
+++ b/src/emu/x64primop.h
@@ -311,7 +311,7 @@ static inline uint8_t shl8(x64emu_t *emu, uint8_t d, uint8_t s)
 		emu->df = d_shl8;
 		emu->op1.u8 = d;
 		emu->op2.u8 = s;
-		emu->res.u8 = d << s;
+		emu->res.u8 = (s>7) ? 0 : (d<<s);
 
 		return emu->res.u8;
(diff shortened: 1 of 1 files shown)


src/emu/x64primop.h
```diff
@@ -311,7 +311,7 @@ static inline uint8_t shl8(x64emu_t *emu, uint8_t d, uint8_t s)
 		emu->df = d_shl8;
 		emu->op1.u8 = d;
 		emu->op2.u8 = s;
-		emu->res.u8 = d << s;
+		emu->res.u8 = s > 7 ? 0 : d << s;
```
ptitSeb, line 314: that's fine, but I think you need paranthesis, I don't think `:` as more priority than `<<` (at least it's more clear that way).

ptitSeb: so `(s > 7) ? 0 : (d <<s)` (I also find the space to be useless with the paranthesis, but it's just personal taste)

ptitSeb (changes requested):
please add some `(` `)`

ksco:
Will do, I'm out for dinner.

ksco:
done.

ptitSeb (approved):

PR #1718, opened 2024-08-08
[RV64_DYNAREC] Fixed register conflict with GETEB macro

xctan:


src/dynarec/rv64/dynarec_rv64_00_3.c
```diff
@@ -499,6 +499,7 @@ uintptr_t dynarec64_00_3(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
             nextop = F8;
             switch((nextop>>3)&7) {
                 case 0:
+                    GETEB(x1, 0);
```
ptitSeb, line 502: Putting code before the "INST_NAME" is not recommanded as it will make the dump a bit strange to read. Can't it be put after?

ksco:
Can we merge this now? I would like to test more with this merged.

PR #1720, opened 2024-08-09
[RV64_DYNAREC] Removed TODOs on GETEX and GETEM macros

ksco:


ptitSeb:
That one should help reduce the size of generated code a bit :)

PR #1722, opened 2024-08-09
[RV64_DYNAREC] Fixed fpu_flags handling and enabled cosim in CI

ksco:


ptitSeb:
Test 6 about threads seems to fail with cosim... or is it random?

ksco:
> Test 6 about threads seems to fail with cosim... or is it random?

It seems random, not sure why, I'll add `--repeat until-pass` option for now.

ksco:
Ah ah, it's okay, tests are running pretty fast, and fail pretty fast too, this PR made the PR 2 minutes longer or so ;)

PR #1743, opened 2024-08-17
[RV64_DYNAREC] Slightly improved LOCK CMPXCHG16B opcode

ksco:


src/dynarec/rv64/dynarec_rv64_f0.c
```diff
@@ -287,6 +287,7 @@ uintptr_t dynarec64_F0(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                         case 0:
                             if (rex.w) {
                                 INST_NAME("LOCK CMPXCHG16B Gq, Eq");
+                                PASS3(dynarec_log(LOG_NONE, "Warning, LOCK CMPXCHG16B is not well supported on RISC-V and issues are expected.\n"));
```
ptitSeb, line 290: Maybe only 1 warning per elf is enough?

ksco: okay, but how? some static variable?

ptitSeb: Yeah, local `static int`...

PR #1749, opened 2024-08-24
[BOX32] [ANDROID] Some Fixes for Android Build

KreitinnSoftware:


KreitinnSoftware:
It's compile correctly but a simple program (test01) don't load

Always give `Faillure to create lib => fail` with BOX64_LOG=2

ptitSeb:
> It's compile correctly but a simple program (test01) don't load
> 
> Always give `Faillure to create lib => fail` with BOX64_LOG=2

Yeah, I ave not ported the `__libc_start` (or whatever name it is, I forgot) wrapped function for Android, that's why it will always fail for now.

KreitinnSoftware:
> > It's compile correctly but a simple program (test01) don't load
> > Always give `Faillure to create lib => fail` with BOX64_LOG=2
> 
> Yeah, I ave not ported the `__libc_start` (or whatever name it is, I forgot) wrapped function for Android, that's why it will always fail for now.

Ah ok

PR #1754, opened 2024-08-24
[ANDROID] Not use 'libbsd.so' as needed lib and don't define NEEDED_LIBS_234 for Android Too, Use BOX64_LD_LIBRARY_PATH instead of LD_LIBRARY_PATH on CTEST

KreitinnSoftware:


ptitSeb:
All Android / Termux tests seems to crash now, can you check?

PR #1756, opened 2024-08-25
[ANDROID] Remove 'libbsd' and 'libpthread', and use their functions directly on libc (since android don't have separated pthread library and libbsd)

KreitinnSoftware:


ptitSeb:
that should not be usefulll actually: the whole point of the NEEDEDLIBS was to force libpthread and libbsd to be loaded with libc. And because wrapped libpthread and libbsd do not load actuall lib but use `dlopen(NULL, ...)` it use current open elf (box64 and libs, so libc and whatever is linkled with it). The current mecanism should do exactly the same. So just add libbsd and libpthread as needed with libc, and ensure the libpthread and libbsd use `dlopen(NULL,...)` in `PRE_INIT` macro and it will just work.

KreitinnSoftware:
> that should not be usefulll actually: the whole point of the NEEDEDLIBS was to force libpthread and libbsd to be loaded with libc. And because wrapped libpthread and libbsd do not load actuall lib but use `dlopen(NULL, ...)` it use current open elf (box64 and libs, so libc and whatever is linkled with it). The current mecanism should do exactly the same. So just add libbsd and libpthread as needed with libc, and ensure the libpthread and libbsd use `dlopen(NULL,...)` in `PRE_INIT` macro and it will just work.

That's true, Then Just revert 'a7c61d8e20d8176ca956b58a4f837c54760ada0b'

PR #1762, opened 2024-08-26
Revert "[ANDROID] Not use 'libbsd.so' as needed lib and don't define NEEDED_LIBS_234 for Android Too, Use BOX64_LD_LIBRARY_PATH instead of LD_LIBRARY_PATH on CTEST"

KreitinnSoftware:
Reverts ptitSeb/box64#1754


KreitinnSoftware:
Some functions on libbsd are needed by wine and it's on android libc, I done a mistake thinking that's no needed

ptitSeb:
But you probably want to re-apply to box64_ld_library_path change later...

KreitinnSoftware:
> But you probably want to re-apply the box64_ld_library_path change later...

Yes, this I will do

PR #1773, opened 2024-08-29
[BOX32][LA64_DYNAREC] Added preliminary box32 support

ksco:


src/dynarec/la64/dynarec_la64_64.c
```diff
@@ -319,9 +319,14 @@ uintptr_t dynarec64_64(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
-                addr = geted(dyn, addr, ninst, nextop, &ed, x2, x1, &fixedaddress, rex, NULL, 1, 0);
-                ADD_D(x4, ed, x4);
-                SDxw(gd, x4, fixedaddress);
+                if (rex.is32bits) {
+                    addr = geted(dyn, addr, ninst, nextop, &ed, x2, x1, &fixedaddress, rex, NULL, 1, 0);
+                    ADDz(x4, ed, x4);
+                    SDxw(gd, x4, fixedaddress);
+                } else {
+                    addr = geted(dyn, addr, ninst, nextop, &ed, x2, x1, &fixedaddress, rex, NULL, 0, 0);
+                    SDXxw(gd, ed, x4);
```
ptitSeb, line 328: How will `SDXxw(...)` works here? x4 will be a full 64bits offset and ed wiil be signed 32bits value. Or is it already signe extended to 64bits?

ksco: addr = ed + x4, all in 64 bit, no matter the data width, so we can’t use it for 32bit.

ptitSeb: so if accessing `gs:[-4]`, x4 will be offsetbase of GS (in unsigned 64bits) and `ed` will be `x2` with a content of 64bits `-4` (and not 32bits `-4` with zeroup)?

ptitSeb: because, on arm64, I'm using `_SXTW` variant of `LDR` for this case...

ksco: yes, but we don't have that variant in LA64 unfortunately.

ptitSeb: There is something I don't understand there. How `ed` can be sign etended here, and not sign extended in other place?

ksco: because if is32bits is true, the memory address has to be restricted in 32bit, else the memory address is a 64 bit value, so no restriction?

ptitSeb: not really: `SDXxw(gd, ed, x4)` will store 32bits value `gd` at 64bits address `ed + x4`.
In the case of `gs:[-4]` , if gs base is, for example `0x100000`, you will store the value at `0x100000+0xfffffffc` so `0x1000ffffc` instead of  `0xffffc`

ptitSeb: but then, what about `[ESP]` with `esp=0xffffffdc` ? will it be also expanded to `0xffffffffffffffdc`?

ptitSeb: there is a specific path for 32bits, ok

PR #1776, opened 2024-08-30
Wrapped 1 more function for libxcb

ksco:


ptitSeb:
Enrelated to this PR: Box32 test on RISC-V is often timing out... Maybe there is something wrong there somewhere?

PR #1786, opened 2024-09-01
[SDORYON1] Oryon1 (Snapdragon Elite) support

powpingdone:
Some compiler flags for the Oryon1 support. I still need to test this. I don't believe that this cpu has 32 bit support, though I'm not 100% sure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,7 @@ option(PHYTIUM "Set to ON if targeting an Phytium (D2000 or FT2000/4) based devi
 option(SD845 "Set to ON if targeting a Snapragon 845 based device" ${SD845})
 option(SD888 "Set to ON if targeting a Snapragon 888 based device" ${SD888})
 option(SD8G2 "Set to ON if targeting a Snapragon 8 Gen 2 based device" ${SD8G2})
+option(SDORYON1 "Set to ON if targeting a Snapdragon Oryon 1 (X1E80100/X1E78100) based device" ${SDORYON1})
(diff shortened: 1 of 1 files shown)


ptitSeb:
There is no 32bits support in Oryon1 CPU Cores.

powpingdone:
That's kinda what I figured, but I didn't find a solid source anywhere.

ptitSeb:
That's a trending with new chips anyway. Cortex X, M1, Oryon... All have droped the 32bits support to save some silicon.

Side note, the official naming is "X", with 2 derivative: X-Elite and X-Plus.

powpingdone:
If you're talking about the `X1E` naming stuff, I'll go ahead and modify it to be something more readable.

powpingdone:
I'm sorry that I have to ask this, but what's the best way to test these flags for compatibility?

ptitSeb:
> I'm sorry that I have to ask this, but what's the best way to test these flags for compatibility?

What do you mean? Testing if the binary build and run on an X-Elite/X-Plus laptop running linux?

powpingdone:
Yeah, kinda. Making sure I don't run into instructions that aren't in the CPU. I might have blown away my previous Gentoo install with these compiler flags, even though I copied them from the GCC repo.

powpingdone:
It does pass the test suite with `-fsanitize={undefined,address}` and no sanitizer, so it might be ok.

powpingdone:
I'm going to presume if it passes the tests, then it's "ok."

ptitSeb:
looks good to me. I'll try to test later (tomorrow?)

PR #1788, opened 2024-09-03
[WRAPPED] Added back some libc function wrappers

ksco:


rajdakin (approved):
Looks good to me (since everything is aliased with everything else)

PR #1791, opened 2024-09-04
[X64LIBS] Changed x86/x64 libs install directory to prevent conflicts

ksco:
I'm not sure about the `pressure_vessel` hack, do we need to do some changes there too?


ptitSeb:
Yes, `BOX64_LD_LIBRARY_PATH` should be setup with that new path there.

ksco:
Note there is also a `BOX86_LD_LIBRARY_PATH` in function `pressure_vessel`, is this a mistake?

PR #1795, opened 2024-09-05
[CI] Refactored CI

ksco:
- Upgraded QEMU
- Removed `BOX64_DYNAREC_TEST=2` as it's not useful..
- Mark certain tests as `nocosim` to not test them under cosim

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -165,44 +165,53 @@ jobs:
           fi
 
           cd build
+          export CTEST_OPTIONS="-j$(nproc) --timeout 60 --output-on-failure --repeat until-pass:20"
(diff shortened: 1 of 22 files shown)


ptitSeb:
Missing removing of dynarec_test=2 in `USAGE.md`?

ptitSeb:
Also, la64 box32 test doesn't seem to like the newer qemu?

ksco:
> Also, la64 box32 test doesn't seem to like the newer qemu?

It does not like the old qemu either, that's why I tried to upgrade it, and it works on my forked repo CI, but that seems just me getting lucky.. Anyway, it works on my local machine, so seems something is wrong with qemu. I excluded this test from LA64.

> Missing removing of dynarec_test=2 in `USAGE.md`?

Indeed, removed.

PR #1796, opened 2024-09-05
box64.box64rc: Extracted mobox's box64rc game list

EDLLT:
While I was dissecting mobox and updating it, I found out that it has a custom box64rc with games that aren't found here so I combined both.


ptitSeb:
Thank you for the PR.
But I would prefer for the author of Mobox to do this PR. Also, there are settings that are Qualcomm specific (and I suspect many settings are now obsolete on current box64 build).

EDLLT:
> Thank you for the PR.
> But I would prefer for the author of Mobox to do this PR. Also, there are settings that are Qualcomm specific (and I suspect many settings are now obsolete on current box64 build).

Oh alright, I'll close this pr then

PR #1801, opened 2024-09-06
Eliminate duplicate hash calculations for `getSymbolInSymbolMaps`

Coekjan:
## Introduction

WPS Presentation can be run on RISC-V Lichee Pi 4A with the power of box64. And I did profiling for it, in order to find the performance bottleneck. The profiling result is shown below:

image

So I investigated why `getSymbolInSymbolMaps` (`kh_get_symbolmap` & `kh_get_symbol2map`) costs so much. I found that `getSymbolInSymbolMaps` calls `kh_get` 6 times at the worst case, and the most expensive calculation in `kh_get` is "hashing". Symbols are strings and current string hashing function needs to traversal the whole string.

My idea is to eliminate the duplicate "hashing", since the pass-in symbol name does not change during `getSymbolInSymbolMaps`.

## Implementation

The optimization is straightforward. That is, just calculate the hash once and reuse the result. To achieve this, I added two APIs called `kh_hash` & `kh_get_with_hash`, and applied them for `getSymbolInSymbolMaps`.

## Evaluation

I evaluated the start-up time of WPS Presentation. Results are shown here:

image

where the "logo" column means the time when the logo displays, and the "full" column means the time when the WPS Presentation fully starts up.

So this simple optimization does improve the performance a lot in my use case. And I am sending this PR and sharing the results with you.

### DIFF
--- a/src/include/khash.h
+++ b/src/include/khash.h
@@ -217,6 +217,22 @@ static const double __ac_HASH_UPPER = 0.77;
 			h->size = h->n_occupied = 0;								\
 		}																\
 	}																	\
+	SCOPE khint_t kh_hash_##name(khkey_t key) {							\
+		return __hash_func(key);										\
+	}																	\
+	SCOPE khint_t kh_get_##name##_with_hash(const kh_##name##_t *h, khkey_t key, khint_t hash) { \
+		if (h->n_buckets) {												\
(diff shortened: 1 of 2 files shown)


ptitSeb:
Interesting analysis. Thanks for the break down and the optimisation. That's a good one.

That looks like a good loading time speedup.

PR #1806, opened 2024-09-08
[WRAPPED] Fixed some libc wrapped functions

rajdakin:
This PR fixes some wrong wrapped functions in libc (`scanf` among others).


src/wrapped/wrappedlibc_private.h
```diff
@@ -307,7 +307,7 @@ GOW(erand48_r, iFppp)
 GOM(err, vFEipV)
 //DATAB(errno, 
 GO(__errno_location, pFv)
-GOW(error, vFiipV)
+//GOWM(error, vFiipV)
```
ptitSeb, line 310: This one should not be disabled, and a wrapped "my_" should be created, prformation the fmt using vsprintf in a local buffer kind of call

src/wrapped/wrappedlibc_private.h
```diff
@@ -2237,7 +2237,7 @@ GOW(ulckpwdf, iFv)
 #ifdef STATICBUILD
 //GOW(ulimit, lFiV)
 #else
-GOW(ulimit, lFiV)
+//GOWM(ulimit, lFiV)
```
ptitSeb, line 2240: should not be disabled. Wrapping it to lFil should be fine.

src/wrapped/wrappedlibc_private.h
```diff
@@ -2273,8 +2273,8 @@ GOWM(vasprintf, iFEppA)
 GOM(__vasprintf_chk, iFEpipp)
 GOM(vdprintf, iFEipA)
 GOM(__vdprintf_chk, iFEiipA)
-GO(verr, vFipA)
-GO(verrx, vFipA)
+GOM(verr, vFEipA)
+GOM(verrx, vFEipA)
```
ptitSeb, line 2277: my_verrx is missing it seems

PR #1833, opened 2024-09-18
[RV64_DYNAREC] Added more opcodes for vector

ksco:


src/dynarec/rv64/dynarec_rv64_660f_vector.c
```diff
@@ -655,12 +655,62 @@ uintptr_t dynarec64_660F_vector(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t i
...
+            VRGATHER_VV(d1, v0, q1, VECTOR_UNMASKED);
+            VMERGE_VVM(q0, d1, d0);
+            break;
+        case 0x63:
+            INST_NAME("PACKSSWB Gx, Ex");
+            nextop = F8;
+            SET_ELEMENT_WIDTH(x1, VECTOR_SEW16, 1);
+            GETGX_vector(q0, 1, VECTOR_SEW16);
+            GETEX_vector(q1, 0, 0, VECTOR_SEW16);
+            fpu_get_scratch(dyn); // HACK: skip v3, for vector register group alignment!
```
ptitSeb, line 680: I don't understand this. Especialy because EX can be a memory and so a scratch register already.

ptitSeb: if you need d0/d1 to be consecutive and with some other constraint, create a function to grab 2 consecutive scratch instead (I think there is one on arm64 already)

src/dynarec/rv64/dynarec_rv64_660f_vector.c
```diff
@@ -677,32 +727,82 @@ uintptr_t dynarec64_660F_vector(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t i
...
             VRGATHER_VV(d1, v0, q1, VECTOR_UNMASKED);
             VMERGE_VVM(q0, d1, d0);
             break;
+        case 0x6B:
+            INST_NAME("PACKSSDW Gx, Ex");
+            nextop = F8;
+            SET_ELEMENT_WIDTH(x1, VECTOR_SEW32, 1);
+            GETGX_vector(q0, 1, VECTOR_SEW32);
+            GETEX_vector(q1, 0, 0, VECTOR_SEW32);
+            fpu_get_scratch(dyn); // HACK: skip v3, for vector register group alignment!
```
ptitSeb, line 787: same here.

src/dynarec/rv64/dynarec_rv64_660f_vector.c
ksco: Indeed, it will be wrong if EX is from memory, I'll fix this as you suggested, merci!

ksco: Done.

ptitSeb:
c'est mieux ;)

PR #1834, opened 2024-09-18
[RV64_DYNAREC] Fixed some found issues in vector

ksco:


ksco:
I'm working on the CI failures.

ksco:
BTW, dav1d works after this, but it's slower than the scalar version.

ksco:
Oh, I got it wrong. It's already a little faster than the scalar version. There are many fallbacks to scalar cases, so it's expected.

ptitSeb:
Yeah, speed will not be great as long as register are save / restored... Soon :D ...

PR #1856, opened 2024-09-23
[ARM64_DYNAREC] Fixed sse_forget_reg function

zohanzephyr:
If sse and neon are mapped one-to-one, is it possible not to use a for loop?


ptitSeb:
XMM are matched one to one, but not the hight part of YMM.

zohanzephyr:
> So this change will not work for AVX programs.

Okay, I got it, thanks. So what's the plan for simulating AVX instructions later for risc-v dynarec, will it still be simulated with scalar instructions now that we have vector instructions?

PR #1874, opened 2024-09-26
[BOX32][WRAPPER] Refined SDL2 display mode wrapping

ksco:


src/wrapped32/wrappedsdl2.c
```diff
@@ -96,9 +96,9 @@ static void* find_AudioCallback_Fct(void* fct)
 
 EXPORT int my32_2_SDL_GetDesktopDisplayMode(int displayIndex, void* mode)
 {
-    inplace_SDL2_DisplayMode_to_64(mode);
-    int ret = my->SDL_GetDesktopDisplayMode(displayIndex, mode);
-    inplace_SDL2_DisplayMode_to_32(mode);
+    my_SDL2_DisplayMode_t my_mode;
```
ptitSeb, line 99: better initialize to `{0}` so if the function fail and does not fill the structure, the covertsion of all 0 will work (it might fail if it's converting garbage, with too large pointer or stuffs like that).

ksco: sure

PR #1887, opened 2024-09-29
[ANDROID][BOX32] Fix __libc_init, Fix Some Warnings

KreitinnSoftware:


src/wrapped/wrappedopenal.c
```diff
@@ -63,9 +63,7 @@ void my_alRequestFoldbackStop(x64emu_t* emu);
 
 #include "wrappedlib_init.h"
 
-
-
-void fillALProcWrapper()
+void fillALProcWrapper(box64context_t* context)
```
ptitSeb, line 66: Don't change the signature of this function. Use my_context if you nedd the context (but for what?)

KreitinnSoftware: I changed it because of the compiler warning, so it compiled without the warning, I didn't know it was intentional to leave it as it was

KreitinnSoftware: /home/pablo/box64/src/wrapped/wrappedopenal.c:116:39: warning: too many arguments in call to 'fillALProcWrapper'
        fillALProcWrapper(emu->context);
        ~~~~~~~~~~~~~~~~~             ^
/home/pablo/box64/src/wrapped/wrappedopenal.c:116:26: warning: passing arguments to 'fillALProcWrapper' without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
        fillALProcWrapper(emu->context);
                         ^
2 warnings generated.

KreitinnSoftware: Well, I'll just return to how it was then

ptitSeb: it should be fixed at the warning, removog the parameter in call, that's the mystake

src/wrapped/wrappedopenal.c
```diff
@@ -149,7 +147,7 @@ EXPORT void* my_alcGetProcAddress(x64emu_t* emu, void* device, void* name)
     const char* rname = (const char*)name;
     printf_log(LOG_DEBUG, "Calling alcGetProcAddress(%p, %s)\n", device, rname);
     if(!emu->context->alwrappers)   // could be moved in "my" structure...
-        fillALProcWrapper();
+        fillALProcWrapper(NULL);
```
ptitSeb: no need for NULL here now

KreitinnSoftware: Already removed on latest commit

PR #1889, opened 2024-09-30
Various box32 wrapped libs and script fixes

rajdakin:
This PR fixes some signatures in `libc` in 32bits, and fixes the `rebuild_wrappers_32.py` script when generating the type of `GOM` functions with structure pointers in the `*types.h` files.


src/wrapped32/wrappedlibc_private.h
```diff
@@ -2045,7 +2045,7 @@ GOWM(writev, lFEipi)
 // xdr_authdes_cred
 // xdr_authdes_verf
 // xdr_authunix_parms
-GO(xdr_bool, iFpp)
+//GO(xdr_bool, iFpp)
```
ptitSeb, line 2048: what was wrong on those?

src/wrapped32/wrappedlibc_private.h
```diff
@@ -2073,7 +2073,7 @@ GO(xdr_bool, iFpp)
 // xdr_longlong_t
 // xdrmem_create
 // xdr_netnamestr
-GO(xdr_netobj, iFpbup_)
+//GO(xdr_netobj, iFpbup_)
```
ptitSeb, line 2076: and this

src/wrapped32/wrappedlibc_private.h
```diff
@@ -2095,7 +2095,7 @@ GO(xdr_netobj, iFpbup_)
 // xdr_string
 // xdr_u_char
 // xdr_u_hyper
-GO(xdr_u_int, iFpp)
+//GO(xdr_u_int, iFpp)
```
ptitSeb, line 2098: and this?

src/wrapped32/wrappedlibc_private.h
rajdakin: I believe I commented this since the `XDR` structure is self-referencing, which would need a manual realignment.
The simplified declaration of the wrapperhelper is:
```
Declaration: xdr_bool -> ...: int32_t((* struct __rpc_xdr <with 6 members, with self-recursion> { <snip> }), (* int32_t))
```
i.e. `int32_t xddr_bool(struct __rpc_xdr*, int32_t*)` with `struct __rpc_xdr` a self-referencing structure.

rajdakin: Reverted.

src/wrapped32/wrappedlibc_private.h
rajdakin: Reverted.

src/wrapped32/wrappedlibc_private.h
rajdakin: Reverted.

PR #1909, opened 2024-10-06
[DYNAREC] Optimized CALL/RET RAS for out of dynablock jumps

ksco:
Related PR: #1907 

For out of block jumps, instead of jump to the table directly, add one more jump to make the RAS optimization valid.

A simple benchmark as below shows this fixes the performance loss in out-of-block-jumps introduced in PR #1907, while keeps the in-block-jumps as fast.

```
#include <limits.h>

int foo(int a) {
  return a + a;
}

int bar(int a) {
  return a + 2;
}

int baz(int a) {
  return a + 3;
}

int indirect_version(int (*fn)(int), int (*fn2)(int), int (*fn3)(int)) {
  int i, b = 0;

  for (i = 0; i < INT_MAX >> 2; ++i) {
    b += fn(b) + fn2(b) - fn3(b);
  }

  return b;
}

int main(int argc, char *argv[]) {
  return indirect_version(&foo, &bar, &baz);
}
```

### DIFF
--- a/src/dynarec/arm64/dynarec_arm64_00.c
+++ b/src/dynarec/arm64/dynarec_arm64_00.c
@@ -3162,10 +3162,9 @@ uintptr_t dynarec64_00(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
                             ADR_S20(x4, j64);
                             MESSAGE(LOG_NONE, "\tCALLRET set return to +%di\n", j64>>2);
                         } else {
-                            MESSAGE(LOG_NONE, "\tCALLRET set return to Jmptable(%p)\n", (void*)addr);
-                            j64 = getJumpTableAddress64(addr);
-                            TABLE64(x4, j64);
-                            LDRx_U12(x4, x4, 0);
+                            j64 = (dyn->insts)?(GETMARK-(dyn->native_size)):0;
+                            ADR_S20(x4, j64);
+                            MESSAGE(LOG_NONE, "\tCALLRET set return to +%di\n", j64>>2);
                         }
                         STPx_S7_preindex(x4, x2, xSP, -16);
                     } else {
@@ -3177,6 +3176,14 @@ uintptr_t dynarec64_00(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
                     else
                         j64 = addr+i32;
                     jump_to_next(dyn, j64, 0, ninst, rex.is32bits);
+                    if(box64_dynarec_callret && addr >= (dyn->start + dyn->isize)) {
+                        // jumps out of current dynablock...
+                        MARK;
+                        j64 = getJumpTableAddress64(addr);
+                        TABLE64(x4, j64);
+                        LDRx_U12(x4, x4, 0);
+                        BR(x4);
+                    }
                     break;
             }
             break;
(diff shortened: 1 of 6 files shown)


src/dynarec/arm64/dynarec_arm64_00.c
```diff
@@ -3177,6 +3176,14 @@ uintptr_t dynarec64_00(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
                     else
                         j64 = addr+i32;
                     jump_to_next(dyn, j64, 0, ninst, rex.is32bits);
+                    if(box64_dynarec_callret && addr >= (dyn->start + dyn->isize)) {
```
ptitSeb, line 3179: I don't think this block is usefull. Why not simply jump to the `jump_to_next` that is just generated before?

ksco: It's there for the `RET` to return.

Before this patch, the out-of-block jump "returned" to the jump table, this violates the RAS assumption and hurts the performance as I tested.

With this patch, the out-of-block jump "returned" to the end of `jump_to_next`, right after the `BLR`, so it's native PC + 4, same as what's in the LR register, which matches the RAS assumption.

ptitSeb: Actually, maybe the whole test on addr out of the block is useless, it should just return there to either continue or jump_to_next?!

ksco: Oh are you saying that `&& addr >= (dyn->start + dyn->isize)` is unnecessary? well, yeah but it saves some memory if that matters...

ptitSeb: well, code compacity also helps to stay more in instruction cache... and if it's useless, it's useless..

ksco: Sure, I'll remove it.

src/dynarec/arm64/dynarec_arm64_00.c
```diff
@@ -3176,7 +3176,7 @@ uintptr_t dynarec64_00(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
                     else
                         j64 = addr+i32;
                     jump_to_next(dyn, j64, 0, ninst, rex.is32bits);
-                    if(box64_dynarec_callret && addr >= (dyn->start + dyn->isize)) {
+                    if(box64_dynarec_callret) {
                         // jumps out of current dynablock...
```
ptitSeb, line 3180: That's not what I meant.
I think the block can be changed to:
```C
                   MARK;
                   jump_to_next(dyn, j64, 0, ninst, rex.is32bits);
                    break;
```
Because the `if(...)` is basicaly `jump_to_next` with a fixed address.

ksco: Oh, that will not work, the return point must be right after the `jump_to_next` (`PC + 4`), to be consistent with the last instruction of `jump_to_next`, which is `BLR x2`, which sets the `LR` to `PC + 4`.

ptitSeb: mmm, then I guess the if should be back to it's previous more complete test

ksco: yeah, already force-pushed.

PR #1914, opened 2024-10-08
[RV64_DYNAREC] Added more opcodes and small optimizations

ksco:


src/dynarec/rv64/dynarec_rv64_emit_logic.c
```diff
@@ -422,7 +422,7 @@ void emit_and32c(dynarec_rv64_t* dyn, int ninst, rex_t rex, int s1, int64_t c, i
         MOV64xw(s3, c);
         AND(s1, s1, s3); // res = s1 & s2
     }
-    if (!rex.w) ZEROUP(s1);
```
ptitSeb, line 425: I'm unsure why the ZEROUP should be needed here, s1 should already be 32bits only, so unless the result is sign extended (and in that case thoses optims are not ok), the upper part will never be filled.

ksco: for example `AND EAX, Id`, s1 is just `xRAX`, which is not 32bits only
```
        case 0x25:
            INST_NAME("AND EAX, Id");
            SETFLAGS(X_ALL, SF_SET_PENDING);
            i64 = F32S;
            emit_and32c(dyn, ninst, rex, xRAX, i64, x3, x4);
            break;
```

ptitSeb: But then, the same optims done in OR and XOR might not be correct if the same trick is used?

ptitSeb: and I see this is the case...

ksco: > But then, the same optims done in OR and XOR might not be correct if the same trick is used?

sorry, I must lost my mind...

Issue #1938, opened 2024-10-14
/box64/src/libtools/my_x11_xevent.c:260:13. cannot build


iHad168:
root@localhost ~/b/build (main) [2]# make install
[  0%] Built target WRAPPERS32
[  0%] Built target WRAPPERS
[ 11%] Built target interpreter
[ 12%] Building C object CMakeFiles/box64.dir/src/libtools/my_x11_xevent.c.o
/root/box64/src/libtools/my_x11_xevent.c: In function ‘convertXEvent’:
/root/box64/src/libtools/my_x11_xevent.c:260:13: error: a label can only be part of a statement and a declaration is not a statement
  260 |             register_events_t* head = register_events_head;
      |             ^~~~~~~~~~~~~~~~~
/root/box64/src/libtools/my_x11_xevent.c: In function ‘unconvertXEvent’:
/root/box64/src/libtools/my_x11_xevent.c:496:13: error: a label can only be part of a statement and a declaration is not a statement
  496 |             register_events_t* head = register_events_head;
      |             ^~~~~~~~~~~~~~~~~
make[2]: *** [CMakeFiles/box64.dir/build.make:1270: CMakeFiles/box64.dir/src/libtools/my_x11_xevent.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:102: CMakeFiles/box64.dir/all] Error 2
make: *** [Makefile:182: all] Error 2

m1m1k4tz:
What hardware do you have? How do you reproduce this?

ptitSeb:
it's more it's an old compiler (or just very strict). I'll try to fix this later, but it's just anoying to do.

PR #1952, opened 2024-10-22
Fix compilation error in my_x11_xevent.c (#1938) and  dynarec_rv64_helper.c

devarajabc:
Resolved a compilation error due to variable declaration within a switch-case statement without enclosing braces. 
This error occurred with older GCC compilers, which require declarations to be part of a block within switch-case constructs.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_helper.c
+++ b/src/dynarec/rv64/dynarec_rv64_helper.c
@@ -2694,11 +2694,12 @@ void vector_loadmask(dynarec_rv64_t* dyn, int ninst, int vreg, uint64_t imm, int
                     ADDI(s1, xZR, 1);
                     VMV_S_X(vreg, s1);
                     return;
-                case 2:
+                case 2: {
                     int scratch = fpu_get_scratch(dyn);
                     VMV_V_I(scratch, 1);
                     VSLIDE1UP_VX(vreg, scratch, xZR, VECTOR_UNMASKED);
                     return;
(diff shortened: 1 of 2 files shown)


src/dynarec/rv64/dynarec_rv64_helper.c
```diff
@@ -2694,11 +2694,12 @@ void vector_loadmask(dynarec_rv64_t* dyn, int ninst, int vreg, uint64_t imm, int
                     ADDI(s1, xZR, 1);
                     VMV_S_X(vreg, s1);
                     return;
-                case 2:
+                case 2:{
```
jserv, line 2697: For the sake of consistent style, you should write as `case 2: {`.
See the file `src/libtools/my_x11_xevent.c` for example.

src/libtools/my_x11_xevent.c
```diff
@@ -256,7 +256,7 @@ void convertXEvent(my_XEvent_32_t* dst, my_XEvent_t* src)
             dst->xcookie.cookie = src->xcookie.cookie;
             dst->xcookie.data = to_ptrv_silent(src->xcookie.data);  // in case data are not initialized
             break;
-        default:
+        default:{
```
jserv, line 259: Ditto. Be aware of the consistent style.

ptitSeb:
A thanks. I'm using modern gcc only, so I didn't noticed this until it was too late...

PR #1954, opened 2024-10-24
[RV64_DYNAREC] Modify some functions related to caching and SSE to support AVX

zohanzephyr:


ksco (changes requested):
AVX support for RV64 will be more cumbersome and different than the ARM64 counterpart. I don't think it's any useful to port a small amount of code directly from ARM64 to "getting started", it has to be a working PoC.

BTW @xctan is interested in MMX support, so after finishing the SSE vector support, I'll start working on AVX soon.

zohanzephyr:
> AVX support for RV64 will be more cumbersome and different than the ARM64 counterpart. I don't think it's any useful to port a small amount of code directly from ARM64 to "getting started", it has to be a working PoC.
> 
> BTW @xctan is interested in MMX support, so after finishing the SSE vector support, I'll start working on AVX soon.
I know what you mean, I'm also working on using RV64 scalar instructions to emulate AVX instructions, these are only part of the content, while the rest I haven't debugged through yet, will continue to submit when I do!

zohanzephyr:
> AVX support for RV64 will be more cumbersome and different than the ARM64 counterpart. I don't think it's any useful to port a small amount of code directly from ARM64 to "getting started", it has to be a working PoC.
> 
> BTW @xctan is interested in MMX support, so after finishing the SSE vector support, I'll start working on AVX soon.

I know what you mean, I'm also looking into using RV64 scalar instructions to emulate AVX instructions, these are just some of the parts that won't affect the rest of the content, while the rest I haven't debugged through yet, will continue to submit when I do!

ksco:
I would suggest you wait for the AVX infrastructure for RV64 to finish if you want to get involved. We want box64 to support various profiles (e.g. w/ or w/o RVV, RVV with different VLENs), so it's not as simple as coping code from ARM64.

ksco:
I've added a very basic AVX infra in #1962, closing this.

PR #1967, opened 2024-10-28
[RV64_DYNAREC] Added AVX.0F 10/11 opcode

zohanzephyr:


src/dynarec/rv64/dynarec_rv64_avx_0f.c
```diff
@@ -0,0 +1,97 @@
...
+
+        case 0x10:
+            INST_NAME("VMOVUPS Gx,Ex");
+            nextop = F8;
+            GETGX();
+            GETEX(x2, 0, 8);
+            SSE_LOOP_MV_Q(x3);
+            if(vex.l) {
+                GETGY();
+                GETEY(x2, 0, 16);
```
ksco, line 75: Hmm, this seems not correct, how is the 16 calculated? Also, `GETEY` seems (very) expensive.

src/dynarec/rv64/dynarec_rv64_avx_0f.c
```diff
@@ -0,0 +1,97 @@
...
+            break;
+        case 0x11:
+            INST_NAME("VMOVUPS Ex,Gx");
+            nextop = F8;
+            GETGX();
+            GETEX(x2, 0, 8);
+            SSE_LOOP_MV_Q2(x3);
+            if(vex.l) {
+                GETGY();
+                GETEY(x2, 0, 16);
```
ksco, line 88: Hmm, this seems not correct, how is the 16 calculated? Also, GETEY seems (very) expensive.

src/dynarec/rv64/dynarec_rv64_avx_0f.c
```diff
@@ -0,0 +1,97 @@
...
+    MAYUSE(q0);
+    MAYUSE(q1);
+    MAYUSE(d0);
+    MAYUSE(d1);
+    MAYUSE(s0);
+    MAYUSE(j64);
+    MAYUSE(cacheupd);
+    #if STEP > 1
+    static const int8_t mask_shift8[] = { -7, -6, -5, -4, -3, -2, -1, 0 };
+    #endif
```
ksco, line 61: this is not used.

src/dynarec/rv64/dynarec_rv64_avx_0f.c
```diff
@@ -0,0 +1,97 @@
...
+
+    rex_t rex = vex.rex;
+
+    switch(opcode) {
+
+        case 0x10:
+            INST_NAME("VMOVUPS Gx,Ex");
```
ksco, line 68: `s/Gx,Ex/Gx, Ex/`

src/dynarec/rv64/dynarec_rv64_avx_0f.c
```diff
@@ -0,0 +1,97 @@
...
+            SSE_LOOP_MV_Q(x3);
+            if(vex.l) {
+                GETGY();
+                GETEY(x2, 0, 16);
+                SSE_LOOP_MV_Q(x3);
+            }
+            if(!vex.l) YMM0(gd);
+            break;
+        case 0x11:
+            INST_NAME("VMOVUPS Ex,Gx");
```
ksco, line 81: `s/Ex,Gx/Ex, Gx/`

src/dynarec/rv64/dynarec_rv64_helper.h
```diff
@@ -501,6 +501,24 @@
...
+#define GETEY(a, D, I12)                                                                         \
+    if (MODREG) {                                                                                \
+        ed = (nextop & 7) + (rex.b << 3);                                                        \
+        fixedaddress = offsetof(x64emu_t, ymm[ed]);                                              \
+        wback = xEmu;                                                                            \
+    } else {                                                                                     \
+        SMREAD();                                                                                \
+        ed = 16;                                                                                 \
+        addr = geted(dyn, addr, ninst, nextop, &wback, a, x3, &fixedaddress, rex, NULL, I12, D); \
+        fixedaddress+=16;                                                                        \
```
ksco, line 519: I don't think we can do `fixedaddress += 16` here, the value will overflow.

ksco (changes requested):
Requested changes. Also, would be great to have `clang-format`ted for all the newly added RV64 codes, thanks. 

(Again, AVX infra for RV64 is not ready yet, things might change dramatically, would be great if you could wait for the infra to finish if you want to get involved. Meanwhile, SSE vector infra is very mature and welcomes contributions for new opcodes).

src/dynarec/rv64/dynarec_rv64_helper.h
zohanzephyr: > I don't think we can do `fixedaddress += 16` here, the value will overflow.

When fetching 256 bits of data, it needs to be offset by 128 bits

src/dynarec/rv64/dynarec_rv64_avx_0f.c
zohanzephyr: > this is not used.

Yes, it's not being used at the moment

src/dynarec/rv64/dynarec_rv64_avx_0f.c
zohanzephyr: > Hmm, this seems not correct, how is the 16 calculated? Also, `GETEY` seems (very) expensive.

Because the operation is on 256 bits of data

src/dynarec/rv64/dynarec_rv64_helper.h
zohanzephyr: > I don't think we can do `fixedaddress += 16` here, the value will overflow.

It won't overflow, I did test it with the demo and the result is correct

src/dynarec/rv64/dynarec_rv64_avx_0f.c
ksco: it should be 24. but like I said, the GETEY is too expensive and probably wrong, just reuse the result from GETEX.

zohanzephyr: > Running you current `GETEY` after `GETEX` will endup running `geted` 2 times for the opcode, wich might result is reading more bytes from the opcode than actually exist. You should really looks athow it's done on the ARM64 side: geted is only called at the GETEX side, not on GETEY. Also, with what program or game are you testing your changed?

It is also possible to call it only on the GETGX side, the fixedaddress address is the same, but fixedaddress +=16 is correct, offset by 128 to get the value of ymm.

zohanzephyr: > Running you current `GETEY` after `GETEX` will endup running `geted` 2 times for the opcode, wich might result is reading more bytes from the opcode than actually exist. You should really looks athow it's done on the ARM64 side: geted is only called at the GETEX side, not on GETEY. Also, with what program or game are you testing your changed?

A test demo I wrote myself

#include <stdio.h>

int main() {
    float src[8] = {1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f, 7.0f, 8.0f};
    float dst[8] = {0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f}; 

    __asm__ volatile (
        "vmovups (%1), %%ymm0\n"  
        "vmovups %%ymm0, (%0)\n"  
        :                        
        : "r" (dst), "r" (src)  
        : "%ymm0" 
    );

    printf("Source data: %f %f %f %f %f %f %f %f\n", src[0], src[1], src[2], src[3], src[4], src[5], src[6], src[7]);
    printf("Destination data: %f %f %f %f %f %f %f %f\n", dst[0], dst[1], dst[2], dst[3], dst[4], dst[5], dst[6], dst[7]);

    return 0;
}
~

ptitSeb: looks like a valid sample code (did you look at the box64 dump to see what is generated code for it?).
Feel free to add a few more opcode and add this the box64 test suite if it makes sense.

zohanzephyr: > looks like a valid sample code (did you look at the box64 dump to see what is generated code for it?). Feel free to add a few more opcode and add this the box64 test suite if it makes sense.

Ok, geted should indeed be executed once, I've made the change

zohanzephyr: > looks like a valid sample code (did you look at the box64 dump to see what is generated code for it?). Feel free to add a few more opcode and add this the box64 test suite if it makes sense.

zohanzephyr: > looks like a valid sample code (did you look at the box64 dump to see what is generated code for it?). Feel free to add a few more opcode and add this the box64 test suite if it makes sense.

ptitSeb: So, the VMOVUPS didn't have SIB or Offset involved in this sample, that's why calling 2* times geted still worked. It would have crash and burn if an offset was involved in the opcode for example. Maybe try to use only 1 single array, instead of two?

(pro tips: use `less -R dump.txt` to see the dump in all it's colorfull glory)

src/dynarec/rv64/dynarec_rv64_helper.h
```diff
@@ -501,6 +501,21 @@
         addr = geted(dyn, addr, ninst, nextop, &wback, a, x3, &fixedaddress, rex, NULL, I12, D); \
     }
 
+
+#define GETGY()                                 \
+    gd = ((nextop & 0x38) >> 3) + (rex.r << 3); \
+    gback = xEmu;                               \
+    gdoffset = offsetof(x64emu_t, ymm[gd])
+
+#define GETEY(a, D, I12)                            \
```
ksco, line 510: `I12` is useless.

src/dynarec/rv64/dynarec_rv64_avx_0f.c
```diff
@@ -0,0 +1,95 @@
...
+
+    rex_t rex = vex.rex;
+
+    switch (opcode) {
+
+        case 0x10:
+            INST_NAME("VMOVUPS Gx,Ex");
+            nextop = F8;
+            GETGX();
+            GETEX(x2, 0, 8);
```
ksco, line 69: 8 is not correct, it should be 24.

ksco (changes requested):

src/dynarec/rv64/dynarec_rv64_helper.h
ksco: please make sure you understand what the i12 is for.

ksco:
this PR is still not correct, the extcache should be kept up to date.

ptitSeb:
@zohanzephyr there are still some changes to be done on this PR.

ptitSeb:
Closing, at it seems to be abandoned.

PR #1998, opened 2024-11-04
[WRAPPER] Replace wrapped libandroid-shmem with wrapped functions in …

olegos2:
…libc for -DANDROID


olegos2:
There's no way to get tests working using libandroid for now.

PR #2000, opened 2024-11-05
[WRAPPER] Wrapped 3 more symbols for libc

ksco:


src/wrapped/generated/wrappedlibctypes.h
```diff
@@ -114,6 +114,7 @@ typedef int32_t (*iFppipppp_t)(void*, void*, int32_t, void*, void*, void*, void*
 	GO(fork, iFv_t) \
 	GO(vfork, iFv_t) \
 	GO(iopl, iFi_t) \
+	GO(__libc_dlclose, iFp_t) \
```
ptitSeb, line 117: I don't think it is safe to wrap dlclose, dlsym and dlopen without a my_ function, to wrap the emulated libraries (or else, nothing will actually work).

ksco: This is auto-generated, the GOM part is down below.

PR #2042, opened 2024-11-18
[WRAPPER]my_dbus_message_get_args() args type error

liuli0217:


ptitSeb:
You need to also added wrapper generated code to the commit

PR #2050, opened 2024-11-20
update pressure_vessel function with additional valid BOX86_LD_LIBRAR…

theofficialgman:
…Y_PATH

needed after https://github.com/ptitSeb/box86/pull/1014


ptitSeb:
Indeed, thanks.

PR #2056, opened 2024-11-22
[LA64_DYNAREC] Initial implement missing opcode for VMOVD Gx, Ed

xiangzhai:
Hi,

Testcase:

```
export BOX64_DYNAREC_MISSING=1
/yourpath/build/box64 /yourpath/x64/jdk8/bin/java -version

0x60030297: Dynarec stopped because of Opcode C5 F9 6E C1 C5 F9 70 C0 00 C4 E3 7D 18 C0 01
```

After implemented `VMOVD Gx, Ed`:

```
0x6003029b: Dynarec stopped because of Opcode C5 F9 70 C0 00 C4 E3 7D 18 C0 01 C5 FE 6F F8
Dynarec unimplemented AVX opcode size 128 prefix 66 map 0F opcode 70
```

I would implement `VPSHUFD Gx, Ex, Ib` in other patchset trying be equivalent to `VZIP` etc.[2]

Please review my initial patch.

[1] Linux/x64 JDK8:  https://www.azul.com/downloads/?version=java-8-lts&os=linux&architecture=x86-64-bit&package=jdk#zulu
[2] https://github.com/ptitSeb/box64/blob/main/src/dynarec/arm64/dynarec_arm64_avx_66_0f.c#L830

Thanks,
Leslie Zhai

### DIFF
--- a/src/dynarec/la64/dynarec_la64_avx_66_0f.c
+++ b/src/dynarec/la64/dynarec_la64_avx_66_0f.c
@@ -1,0 +1,85 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <stddef.h>
+#include <errno.h>
+
+#include "debug.h"
+#include "box64context.h"
+#include "dynarec.h"
+#include "emu/x64emu_private.h"
+#include "emu/x64run_private.h"
+#include "x64run.h"
+#include "x64emu.h"
+#include "box64stack.h"
+#include "callback.h"
+#include "emu/x64run_private.h"
+#include "x64trace.h"
+#include "dynarec_native.h"
+#include "my_cpuid.h"
+#include "emu/x87emu_private.h"
+#include "emu/x64shaext.h"
+
+#include "la64_printer.h"
+#include "dynarec_la64_private.h"
+#include "dynarec_la64_functions.h"
+#include "../dynarec_helper.h"
+
+uintptr_t dynarec64_AVX_66_0F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ninst, vex_t vex, int* ok, int* need_epilog)
+{
+    (void)ip; (void)need_epilog;
+
+    uint8_t opcode = F8;
+    uint8_t nextop, u8;
+    uint8_t gd, ed;
+    uint8_t wback, wb1, wb2;
+    uint8_t eb1, eb2, gb1, gb2;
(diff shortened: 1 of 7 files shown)


ptitSeb:
The commit looks fine, but I think some of the infrastructure is missing, and will need to be added later, to handle upper 128bits part of ymm0 (load from `x64emu_t`, store to `x64emu_t`, etc...)

src/dynarec/la64/dynarec_la64_avx_66_0f.c
```diff
@@ -0,0 +1,85 @@
...
+
+    switch(opcode) {
+        case 0x6E:
+            INST_NAME("VMOVD Gx, Ed");
+            nextop = F8;
+            GETGX_empty(v0);
+            GETED(0);
+            if (rex.w) {
+                MOVGR2FR_D(v0, ed);
+                FFINT_D_L(v0, v0);
```
ksco, line 74: This is not correct, the upper bit should be cleared.

src/dynarec/la64/dynarec_la64_avx_66_0f.c
```diff
@@ -0,0 +1,85 @@
...
+            } else {
+                VXOR_V(v0, v0, v0);
+                VINSGR2VR_W(v0, ed, 0);
+            }
+            YMM0(gd);
```
ksco, line 79: We have LASX at hand, the YMM0 trick is simply not needed on LoongArch, it should be done with LASX instructions.

ksco (changes requested):
The loongarch AVX intra should be done completely different than ARM64.

ksco:
> The commit looks fine, but I think some of the infrastructure is missing, and will need to be added later, to handle upper 128bits part of ymm0 (load from `x64emu_t`, store to `x64emu_t`, etc...)

Nah, the commit is not fine, because we don't need the YMM0 trick on LoongArch.

ptitSeb:
Also, not that a Dynablock missing opcode doesn't always means that this opcode will be executed. On RV64 and LA64 architecture, AVX/AVX2 is not exposed by default and you need to use `BOX64_AVX=2` to do so. It's only activated by default on Dynarec ARM64 configurations.

xiangzhai:
Offline discussion: implement general instruction at first.

Thanks,
Leslie Zhai

PR #2060, opened 2024-11-23
[RBTREE] Document the rationale for memory management

devarajabc:
Enhanced the rbtree.h header with detailed documentation comments for each function. 
These comments explain the purpose, parameters, and expected behavior of the functions.

### DIFF
--- a/src/include/rbtree.h
+++ b/src/include/rbtree.h
@@ -1,3 +1,27 @@
+/*
+ * The primary function of red-black trees in Box64 is to provide a fast and efficient method (O(log(n))) for managing memory mappings. 
+ * Each red-black tree node includes two additional pointers, "start" and "end," to denote a specific memory range. 
+ * When a range of memory is mapped from the system, it is recorded in red-black trees based on the type and characteristics of that memory range.
+ *
+ * Box64 utilizes four red-black trees, each serving distinct purposes:
+ * 1. memprot: Manages memory protection settings where each node's data field represents the permissions of the memory range. 
+ *    The "rb_set" function within this tree can set or change permissions for a specified range.
+ * 2. mapallmem: Tracks all memory mappings. Its nodes' data field can indicate whether the memory is included in this mapping.
+ * 3. mmapmem: Similar to "mapallmem" but specifically for memory mapped through the mmap system call. 
+ *    The nodes' data fields here differentiate whether the memory is exclusively mapped by mmap.
+ * 4. blockstree: Specifically contains memory ranges for blocks, where each node represents a different block. 
+ *    The "data" field in this tree represents the index of the block in the "blocklist_s" array "p_blocks".
+ *
+ * Currently, there is an overlap between "mapallmem" and "mmapmem", where if memory is mapped in "mmapmem", it is also considered mapped in "mapallmem". 
+ * If "mapallmem" and "mmapmem" were to be merged in the future, the "data" field could then take values indicating:
+ * - 0: Memory is not mapped.
+ * - 1: Memory is only mapped in "mapallmem".
+ * - 2: Memory is mapped in both "mapallmem" and "mmapmem".
+ * 
+ * Before the introduction of the red-black tree in Box64, the rationale for memory management was a sparse array, which takes O(n) complexity for accessing data.
+ * After transitioning from a sparse array to a red-black tree, the memory usage has decreased slightly for processes that consumed a lot of RAM (for example, Steam uses about 100 MB less memory, and each Wine process uses about 15 MB less).
+ */
+
 #ifndef RBTREE_H
 #define RBTREE_H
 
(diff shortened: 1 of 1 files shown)


jserv (changes requested):
Rebase the latest `main` branch.

jserv:
The subject can be shortened to `[RBTREE] Document the rationale for memory management`.
Read [How to Write a Git Commit Message](https://cbea.ms/git-commit/) carefully and enforce the rules.

src/include/rbtree.h
```diff
@@ -2,16 +2,162 @@
 #define RBTREE_H
 
 #include <stdint.h>
-
+/*
```
jserv, line 5: Add a newline between `#include` and the comment.

src/include/rbtree.h
```diff
@@ -2,16 +2,162 @@
...
+The main functionality of the red-black trees is to provide a fast,
```
jserv, line 6: Specify the abbreviated form of red-black tree as "rbtree" in the first occurrence, so that "rbtree" can appear in the latter context, making comments more clear.

src/include/rbtree.h
```diff
@@ -2,16 +2,162 @@
 #define RBTREE_H
 
 #include <stdint.h>
-
+/*
+The main functionality of the red-black trees is to provide a fast, 
+cheap way to map ranges of addresses to a number and acccessing by providing an address in this range.
```
jserv: Use more detailed wording for the rationale of utilizing red-black trees for memory management. You can discuss the prior situation when red-black trees were not introduced, where sparse arrays were used, resulting in memory waste. With the introduction of red-black trees, explain the considerations and benefits below.

src/include/rbtree.h
```diff
@@ -1,17 +1,165 @@
+/*
+ * The main functionality of the red-black trees is to provide a fast, 
+ * cheap way to map ranges of addresses to a number and acccessing by providing an address in this range.
+ */
```
jserv, line 4: Use more detailed wording for the rationale of utilizing red-black trees for memory management. You can discuss the prior situation when red-black trees were not introduced, where sparse arrays were used, resulting in memory waste. With the introduction of red-black trees, explain the considerations and benefits below.

Make it informative.

ptitSeb: Memory was one goal for sure. But speed of access was also another one. The old method using sparse arrays was an `O(n)` complexity for accessing data, while the rbtree is `O(ln(n)` and because the `memprot` can have hundreds of entries and is used quite often, the speed difference is measurable there.

src/include/rbtree.h
```diff
@@ -1,17 +1,173 @@
...
+ * The primary function of red-black trees is to provide a fast (O(ln(n))) and cost-effective method for mapping address ranges to numbers and accessing them by specifying an address within that range.
+ * Each RB-node includes two additional pointers, "start" and "end," to denote a memory range.
+ * When a range of memory is mapped from the system, it is recorded by multiple red-black trees with the information of that memory range (e.g., permissions).
+ * That's it! Each node encapsulates a memory range along with detailed information about that range.
+ * Before the introduction of the red-black tree in Box64, 
+ * the rationale for memory management was a sparse array, 
+ * which take O(n) complexity for accessing data.
+ * After transitioning from a sparse array to a red-black tree, 
+ * the memory usage has decreased slightly for processes that consumed a lot of RAM 
+ * (for example, Steam uses about 100 MB less memory, and each Wine process uses about 15 MB less).
```
rajdakin, line 11: This is really weird to read: the beginning has no break mid-sentence, then midway through there are line breaks after every comma. Please either always add breaks or keep every sentence in one line.
```suggestion
 * Before the introduction of the red-black tree in Box64, the rationale for memory management was a sparse array, which take O(n) complexity for accessing data.
 * After transitioning from a sparse array to a red-black tree, the memory usage has decreased slightly for processes that consumed a lot of RAM (for example, Steam uses about 100 MB less memory, and each Wine process uses about 15 MB less).
```

src/include/rbtree.h
```diff
@@ -1,17 +1,173 @@
...
+ * rb_set() - Set an address range in a red-black tree.
+ * @tree: Pointer to the red-black tree where the address range will be set.
+ * @start: The starting address of the range to be set.
+ * @end: The ending address of the range to be set.
+ * @data: The data value to associate with the address range.
+ *
+ * This function adds a new address range with associated data to the red-black tree. 
+ * If the range overlaps with existing nodes, it intelligently merges or modifies nodes to maintain non-overlapping, contiguous ranges.
+ * This includes extending, splitting, or merging nodes based on the overlap conditions and data consistency. 
+ * It handles multiple edge cases:
```
rajdakin, line 73: There is also the case of adjacency with no overlap:
```
+---------+
| data A  |
+---------+
             +---------+
             | data A  |
             +---------+
           ||
          \||/
           \/
+---------+---------+
|      data A       |
+---------+---------+
```

src/include/rbtree.h
```diff
@@ -1,17 +1,173 @@
...
+ *    and the overlapped part will be overwritten with the data of the new address range.
+ * 
+ * (case 1)
+ *       +---------+---------+---------+---------+---------+
+ *       |                   data A                        |
+ *       +---------+---------+---------+---------+---------+
+ *                  ^.....       overlap      ....^
+ *                  +---------+---------+---------+
+ *                  |            data B           |
+ *                  +---------+---------+---------+
```
rajdakin, line 85: This is nitpicking, but this is misaligned. Also, the `overlap` is not centered, even though it could be (there are five dits and seven spaces on the left, six spaces and four dots on the right).
```suggestion
 *                 ^.....      overlap      .....^
 *                 +---------+---------+---------+
 *                 |            data B           |
 *                 +---------+---------+---------+
```

src/include/rbtree.h
```diff
@@ -1,17 +1,173 @@
...
+ *                            |                   data B                        |
+ *                            +---------+---------+---------+---------+---------+
+ * 
+ *                                ||
+ *                               \||/
+ *                                \/
+ *   
+ *       +---------+---------+ +-----------------------------+---------+---------+
+ *       |      data A       | |                     data B                      |
+ *       +---------+---------+ +-------------------+---------+---------+---------+
```
rajdakin, line 127: This is nitpicking as well: the `data B` block is missing some `+` signs (they got replaced by `-`).

rajdakin (changes requested):
Overall this is fine, but please fix at least the first two issues I've reported.

rajdakin:
Also, you can have multiple commits, they will simply be squashed into one when they are merged.

devarajabc:
> Also, you can have multiple commits, they will simply be squashed into one when they are merged.

Thanks. I also want to add more detail to the scenario of 'mapping memory' and clarify the relation between the dynamic block and the red-black tree, since the abstract is a bit ambiguous.

src/include/rbtree.h
```diff
@@ -1,17 +1,207 @@
+/*
+ * The primary function of red-black trees is to provide a fast (O(ln(n))) and cost-effective method for mapping address ranges to numbers and accessing them by specifying an address within that range.
+ * Each RB-node includes two additional pointers, "start" and "end," to denote a memory range.
+ * When a range of memory is mapped from the system, it is recorded by multiple red-black trees with the information of that memory range.
+ * 
+ * In fact, there are foru red-black tree in custommem.c: memprot, mapallmem, mmapmem and blockstree.
```
rajdakin, line 6: Typo
```suggestion
 * In fact, there are four red-black trees in custommem.c: memprot, mapallmem, mmapmem and blockstree. 
```

src/include/rbtree.h
```diff
@@ -1,17 +1,207 @@
+/*
+ * The primary function of red-black trees is to provide a fast (O(ln(n))) and cost-effective method for mapping address ranges to numbers and accessing them by specifying an address within that range.
+ * Each RB-node includes two additional pointers, "start" and "end," to denote a memory range.
+ * When a range of memory is mapped from the system, it is recorded by multiple red-black trees with the information of that memory range.
+ * 
+ * In fact, there are foru red-black tree in custommem.c: memprot, mapallmem, mmapmem and blockstree. 
+ * The "blockstree" specifically contains memory ranges for blocks, where each node represents a different block and "data" represent the index of the block in blocklist_s array "p_blocks".
+ * The other three trees serve as general-purpose memory tracking mechanisms. 
+ * In "memprot", data represents the permissions of that memory range, and the function rb_set can set or change permissions for a range of memory.
+ * In the case of "mapallmem" and "mmapmem", "data" can be '1' or '2', depending on factors that are not specified.
```
rajdakin, line 10: This is false for now. Also, `mapallmem` contains `mmapmem` (i.e. `mapallmem` is 1 if `mmapmem` is 1), so if/when `mapallmem` and `mmapmem` are merged, `data` could take values `0` (nowhere), `1` (only in `mapallmem`) or `2` (`mapallmem` + `mmapmem`).
Note that there is still the question of whether this merge should happen.

rajdakin (changes requested):
Please fix the abstract, the rest looks good to me.

jserv (approved):

rajdakin (approved):
Looks good to me.

PR #2070, opened 2024-11-25
[ARM64_DYNAREC] Added DYNAREC_PAUSE option for hint instructions

ksco:


src/dynarec/arm64/dynarec_arm64_helper.h
```diff
@@ -1140,6 +1140,8 @@
 #endif
 #define CLEARIP()   dyn->last_ip=0
 
+#define SKIP_SEVL(val) if (dyn->insts[dyn->insts[ninst].x64.jmp_insts].wfe) i32 += 4;
```
ptitSeb, line 1143: it's `val+=4;` I guess here.

ksco: :O Thanks!

src/dynarec/arm64/dynarec_arm64_00.c
```diff
@@ -1427,7 +1429,15 @@ uintptr_t dynarec64_00(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
             if(gd==xRAX) {
                 if (rep == 2) {
                     INST_NAME("PAUSE");
-                    YIELD;
+                    if (box64_dynarec_pause == 1)
```
ptitSeb, line 1432: a `switch(box64_dynarec_pause)` would make more sense

PR #2076, opened 2024-11-26
[LA64_DYNAREC] Added SBB opcodes

xiangzhai:
Hi,

Passed testcase for `SBB Gb, Eb`:

```
#include <stdio.h>
#include <stdint.h>
#include <assert.h>

void test_sbb8(uint8_t n, uint8_t m, uint8_t expected_al, uint8_t expected_dl) {
  uint8_t al;
  uint8_t dl;
  __asm__ __volatile__ (
      "movb  %[n], %%dl\n" 
      "movb  %[m], %%al\n" 
      "subb  %[m], %%al\n" 
      "sbbb  %[n], %%dl\n" 
      ".byte 0x1A\n"                   // SBB Gb, Eb
      ".byte 0xCA\n" 
      "movb  %%al, %[al]\n" 
      "movb  %%dl, %[dl]\n" 
      : [al] "=r" (al), [dl] "=r" (dl)
      : [n] "r" (n), [m] "r" (m)
      : "al", "dl", "memory");
  printf("al=0x%x\n", al);
  printf("dl=0x%x\n", dl);
  assert(al == expected_al && dl == expected_dl);
}

int main(int argc, char* argv[]) {
  test_sbb8(0x1, 0x0, 0x0, 0x0);
}
```

Please review my patch.

Thanks,
Leslie Zhai

### DIFF
--- a/src/dynarec/la64/dynarec_la64_helper.h
+++ b/src/dynarec/la64/dynarec_la64_helper.h
@@ -250,6 +250,28 @@
         wb1 = 1;                                                                                \
         ed = i;                                                                                 \
     }
+// GETEB32 will use i for ed, and can use r3 for wback.
+#define GETEB32(i, D)                                                                             \
+    if (MODREG) {                                                                                 \
+        if (rex.rex) {                                                                            \
+            wback = TO_LA64((nextop & 7) + (rex.b << 3));                                         \
+            wb2 = 0;                                                                              \
+        } else {                                                                                  \
+            wback = (nextop & 7);                                                                 \
+            wb2 = (wback >> 2) * 8;                                                               \
+            wback = TO_LA64(wback & 3);                                                           \
+        }                                                                                         \
+        BSTRPICK_D(i, wback, wb2 + 7, wb2);                                                       \
+        wb1 = 0;                                                                                  \
+        ed = i;                                                                                   \
+    } else {                                                                                      \
+        SMREAD();                                                                                 \
+        addr = geted32(dyn, addr, ninst, nextop, &wback, x3, x2, &fixedaddress, rex, NULL, 1, D); \
+        LD_BU(i, wback, fixedaddress);                                                            \
+        wb1 = 1;                                                                                  \
+        ed = i;                                                                                   \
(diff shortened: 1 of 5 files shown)


src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -250,6 +250,28 @@
         wb1 = 1;                                                                                \
         ed = i;                                                                                 \
     }
+// GETEB32 will use i for ed, and can use r3 for wback.
+#define GETEB32(i, D)                                                                             \
+    if (MODREG) {                                                                                 \
+        if (rex.rex) {                                                                            \
+            wback = xRAX + (nextop & 7) + (rex.b << 3);                                           \
```
ksco, line 257: This is not correct, `TO_LA64` should be used here.

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -250,6 +250,28 @@
...
+            wb2 = 0;                                                                              \
+        } else {                                                                                  \
+            wback = (nextop & 7);                                                                 \
+            wb2 = (wback >> 2) * 8;                                                               \
+            wback = xRAX + (wback & 3);                                                           \
```
ksco, line 262: Ditto.

ksco (changes requested):
Adding opcodes this way, IMHO, is dangerous. Because the submitted code is hardly tested, it is hard to find all the problems just by reviewing it.

A more practical approach would be to use BOX64_DYNAREC_MISSING to indicate the opcodes used by your real world x86_64 programs, and test with as many different programs as possible (mostly games).

(Please remember to clang-format the code.)

ksco:
Please update the commit message and PR title to: [LA64_DYNAREC] Added SBB opcodes

src/dynarec/la64/dynarec_la64_helper.h
ksco: There is a similar issue in last PR, please fix that too.

ksco (approved):

PR #2083, opened 2024-11-27
[LA64_DYNAREC] Act like INT 3

xiangzhai:
Hi,

Testcase:
```
#include <stdio.h>
#include <stdlib.h>

int main(int argc, char* argv[]) {
  __asm__ __volatile__ ("int3\n");
  char* p = getenv("BOX64_IGNOREINT3");
  if (p && p[0] == '1') {
    printf("You can see me\n");
  } else {
    printf("You can *NOT* see me\n");
  }
  return 0;
}
```

Before:
```
You can *NOT* see me
```

After applied the patch:
```
export BOX64_IGNOREINT3=1
You can see me

export BOX64_IGNOREINT3=0
Trace/breakpoint trap
```

Please review my patch.

Thanks,
Leslie Zhai


ptitSeb:
BOX64_IGNOREINT3=1 is here to ignore int3. Why would you want to remove it's effect?

PR #2091, opened 2024-11-28
[RV64] Added nan propagation emulation for interpreter and DynaRec

### LINKED ISSUE #1971: [RISCV] Sample program fails on RISC-V DynaRec and Interpreter
```wget ksco.cool/EmDe```

This program prints `CA9E4439` on the x86-64 native, ARM, LoongArch, and x86-64 interpreter but `A494CECB` on the RISC-V, both Dynarec and interpreter.

ksco:
fixed #1971


ksco:
Ah, this makes cosim unhappy, I'll add DynaRec support in this PR then.

ksco:
This should be done for the SIMD instructions too, but it can be delayed to the point when we add fastnan emulation for `0F` opcodes, I guess.

ptitSeb:
tricky one. I guess the `NAN_PROPAGATION` macro will not work for AVX 3-operands type of opcodes. But that's for later.

PR #2092, opened 2024-11-29
[RV64_DYNAREC] Added gp as a dedicated register for future usage

ksco:


ksco:
Not sure why but static build is broken, I'll fix it tomorrow..

ksco:
Sad, the compiler always tries to use `gp` when possible, even when we initialize `gp` in the `epilog`. I believe it's possible to tell the compiler not to do this, but it's too fragile and not a good practice.

PR #2111, opened 2024-12-05
[DYNAREC] Introduced TO_NAT to ease register mapping changes in future

ksco:
... also, I did a massive reformat on rv64 backend and part of the arm64 macros as well.


ptitSeb:
Note that `FROM_NAT` will also be needed.

ksco:
> Note that `FROM_NAT` will also be needed.

where?

ksco:
> > > Note that `FROM_NAT` will also be needed.
> > 
> > 
> > where?
> 
> Note sure. I thought I needed the reverse operation at some point, maybe on some hlper.c stuffs, but maybe it's not needed at all.

Okay. I’ll be careful on this!

PR #2117, opened 2024-12-06
[LA64_DYNAREC] Rename GETEX to GETEX_vector, GETGX to GETGX_vector and GETGX_empty to GETGX_empty_vector

xiangzhai:
Hi,

Please review my patch.

Thanks,
Leslie Zhai

### DIFF
--- a/src/dynarec/la64/dynarec_la64_0f.c
+++ b/src/dynarec/la64/dynarec_la64_0f.c
@@ -156,21 +156,21 @@ uintptr_t dynarec64_0F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             nextop = F8;
             if (MODREG) {
                 INST_NAME("MOVHLPS Gx,Ex");
-                GETGX(v0, 1);
+                GETGX_vector(v0, 1);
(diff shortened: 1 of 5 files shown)


ksco (changes requested):
This seems pointless, can you explain why?

xiangzhai:
>  This seems pointless, can you explain why?

There is [GETGX](https://github.com/ptitSeb/box64/blob/main/src/dynarec/rv64/dynarec_rv64_helper.h#L454) macro for RV64 general register usage and [GETGX_vector](https://github.com/ptitSeb/box64/blob/main/src/dynarec/rv64/dynarec_rv64_helper.h#L479) for vector register, but there is only `GETGX` for LA64 vector register. [保持队形](https://github.com/ptitSeb/box64/blob/main/src/dynarec/rv64/dynarec_rv64_660f38.c#L771) when adding `AES` opcodes.

Thanks,
Leslie Zhai

ksco:
RISC-V did that for a reason -- there are two different set of implementations: scalar and vector. This does not adapt to LoongArch, where LSX/LASX is mandatory. Closing.

PR #2122, opened 2024-12-07
[LA64_DYNAREC] Added AES opcodes

xiangzhai:
Hi,

Please review my patch.

Thanks,
Leslie Zhai


ksco (changes requested):
RV64 implementations does not always apply to LA64, I believe there is no need to do this while you have LSX at hand.

ksco:
Check how ARM64 did it, and adapt it to LA64.

ksco (approved):
Better.

xiangzhai:
Thanks!

PR #2134, opened 2024-12-10
[LA64] SIGRTMAX is 127 for ABI 1.0

xiangzhai:
Hi,

Reproduced:
```
box64 /yourpath/x64-jdk23 some testcase
```

`Java_sun_nio_ch_NativeThread_init` [src/java.base/unix/native/libnio/ch/NativeThread.c](https://github.com/openjdk/jdk/blob/master/src/java.base/unix/native/libnio/ch/NativeThread.c#L70) thrown Invalid argument exception.

Reduced Testcase:
```
#include <stdio.h>
#include <signal.h>
#include <string.h>
#include <errno.h>

#define INTERRUPT_SIGNAL (SIGRTMAX - 2)

static void
nullHandler(int sig)
{
}

int main(int argc, char* argv[])
{
    struct sigaction sa, osa;
    sa.sa_handler = nullHandler;
    sa.sa_flags = 0;
    sigemptyset(&sa.sa_mask);
    if (sigaction(INTERRUPT_SIGNAL, &sa, &osa) < 0)
        printf("%s INTERRUPT_SIGNAL: %d\n", strerror(errno), INTERRUPT_SIGNAL);
    return 0;
}
```

The root cause is `SIGRTMAX` equals to 127 for LA64 ABI 1.0, so meet `signum>MAX_SIGNAL` condition:
```
int EXPORT my_sigaction(x64emu_t* emu, int signum, const x64_sigaction_t *act, x64_sigaction_t *oldact)
{
    printf_log(LOG_DEBUG, "Sigaction(signum=%d, act=%p(f=%p, flags=0x%x), old=%p)\n", signum, act, act?act->_u._sa_handler:NULL, act?act->sa_flags:0, oldact);
=>    if(signum<0 || signum>MAX_SIGNAL) {
        errno = EINVAL;
        return -1;
    }
...
```

Please review my patch.

Thanks,
Leslie Zhai


ksco:
Hmmm, I expect a CMake option for this, since it's a for the ABI 1.0

src/include/box64context.h
```diff
@@ -52,7 +52,7 @@ typedef struct kh_dynablocks_s  kh_dynablocks_t;
 typedef void* (*procaddress_t)(const char* name);
 typedef void* (*vkprocaddess_t)(void* instance, const char* name);
 
-#ifdef __loongarch64
+#ifdef LA64_ABI_1
 #define MAX_SIGNAL 127
```
ptitSeb, line 56: use 128 and not 127 here, it should be MAX_SIG+1

Issue #2136, opened 2024-12-11
[WRAPPER] The function or data cairo_image_surface_create is declared in multiple files


xiangzhai:
Hi,

x64 openjdk [src/java.desktop/unix/native/libawt_xawt/awt/gtk3_interface.c]( https://github.com/openjdk/jdk/blob/master/src/java.desktop/unix/native/libawt_xawt/awt/gtk3_interface.c#L347) `gtk3_load` not only check gtk related API but also check cairo:
```
static void* dl_symbol(const char* name)
{
    void* result = dlsym(gtk3_libhandle, name);
    if (!result) {
        longjmp(j, NO_SYMBOL_EXCEPTION);
    }

    return result;
}

GtkApi* gtk3_load(JNIEnv *env, const char* lib_name)
{
...
    fp_gtk_check_version = dl_symbol("gtk_check_version");
...
=>  fp_cairo_image_surface_create = dl_symbol("cairo_image_surface_create");
...
    /* Now we have only one kind of exceptions: NO_SYMBOL_EXCEPTION
     * Otherwise we can check the return value of setjmp method.
     */
    else
    {
        dlclose(gtk3_libhandle);
        gtk3_libhandle = NULL;

        dlclose(gthread_libhandle);
        gthread_libhandle = NULL;

        return NULL;
    }
...
```

if added cairo related API wrapper into gtk3:
```
diff --git a/src/wrapped/wrappedgtk3_private.h b/src/wrapped/wrappedgtk3_private.h
index ac87f92d..63f69330 100644
--- a/src/wrapped/wrappedgtk3_private.h
+++ b/src/wrapped/wrappedgtk3_private.h
@@ -4591,3 +4591,5 @@ GO(gtk_wrap_mode_get_type, LFv)
 GO(dummy_iFppdd, iFppdd)    // for GtkEventController wrapping
 GO(dummy_iFppUup, iFppUup)
 GO(dummy_pFpLi, pFpLi)
+
+GO(cairo_image_surface_create, pFiii)
```

rebuild_wrappers.py halt:
```
The function or data cairo_image_surface_create is declared in multiple files (wrappedcairo_private.h/wrappedgtk3_private.h) [extra note in the script]
```

So I just comment `halt_required = True` and rebuilt box64, it works :) but open an issue about is it able to add cario related API into gtk3?

Thanks,
Leslie Zhai

ptitSeb:
[...]

PR #2137, opened 2024-12-11
[WRAPPER] Added libcairo.so.2 as one of NEEDED_LIBS for gtk3 (#2136)

xiangzhai:
Hi,

Please review my patch.

Thanks,
Leslie Zhai


ptitSeb:
Why not adding the the dependancy to libpangocairo instead? that would make more sense. I haven't checked the dependancis of the native libs but I doubt gtk3 has a direct dependancy to cairo tbh

ptitSeb:
image

ptitSeb:
image

ptitSeb:
actually, there is a direct dependancy, my bad!

PR #2139, opened 2024-12-11
[RV64_DYNAREC] New register mapping

ksco:
isSimpleWrapper is pretty limited now, it will be addressed next, to be as good as the Arm one.


ksco:
Noticed regressions on wine games, either revert this or disable RVV will fix the issue, could be something else unrelated.

ptitSeb (approved):

PR #2147, opened 2024-12-12
[RV64_DYNAREC] Reworked ZEROUP and freed t0

xctan:


ksco (approved):

ksco:
(we will use the newly available register for something else, probably for callret? not sure yet.)

ptitSeb:
You should remove the `#define xMask` also.

So you will slow down rv64gc version with this change.

xctan:
> So you will slow down rv64gc version with this change.

Recent hardware either has zba or xtheadbb, so it shouldn't slow down on actual devices.

PR #2150, opened 2024-12-13
[RV64_DYNAREC] Optimize push/pop with xtheadmemidx

ksco:


ptitSeb:
Some tests (idiv_32 and tlsData_32) seems to fail on box32 build. But not all the time?

ksco:
> Some tests (idiv_32 and tlsData_32) seems to fail on box32 build. But not all the time?

The 11th run failed, which is `BOX64_DYNAREC_TEST=1 QEMU_CPU=thead-c906 ctest $CTEST_OPTIONS -E nocosim`.
... I reproduced this issue locally.
... Okay, fixed.

ptitSeb:
Oh, there was some sign extension, I see.
But now, I'm a bit woring of the POP1_16 fix. Is RSP will be truncated to 16bits now?

ksco:
> Oh, there was some sign extension, I see. But now, I'm a bit woring of the POP1_16 fix. Is RSP will be truncated to 16bits now?

nah, the addition is always full-sized, just like an `ADDI`.

PR #2159, opened 2024-12-17
[CMAKE] Fixed linking issue with mold

xctan:


ptitSeb:
Link works for me with MOLD as-is, but only if I add `-f-use-ld=mold` to link settings.

xctan:
When `WITH_MOLD` was on, the linking flags were overwritten by clang style ones, so on my machine mold would complain about missing symbols in libresolv.

PR #2160, opened 2024-12-17
[RV64_DYNAREC] Extended simple wrapper for more int types

xctan:


ksco (approved):

ptitSeb:
How is the speed increase?

xctan:
> How is the speed increase?

Tested on Banana Pi BPI-F3:

| test case | before new register mapping | now |
| -- | -- | -- |
| `getpid()` | 1140444 calls/s | 1293757 calls/s |
| `free(NULL)` | 15438979 call/s | 19245900 call/s |

ptitSeb:
not bad

ksco:
The improvements might be diluted for functions tend to do more, but it's nice to have anyway.

PR #2161, opened 2024-12-17
[LA64_DYNAREC] Added CLC, PMOVZXBW and POPCNT opcodes

xiangzhai:
Hi,

Please review my patch.

Thanks,
Leslie Zhai

### DIFF
--- a/src/dynarec/la64/dynarec_la64_00.c
+++ b/src/dynarec/la64/dynarec_la64_00.c
@@ -2477,6 +2477,12 @@ uintptr_t dynarec64_00(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                     DEFAULT;
             }
             break;
+        case 0xF8:
+            INST_NAME("CLC");
+            SETFLAGS(X_CF, SF_SUBSET);
+            SET_DFNONE();
+            BSTRINS_D(xFlags, xZR, F_CF, F_CF);
+            break;
         case 0xFC:
             INST_NAME("CLD");
             BSTRINS_D(xFlags, xZR, F_DF, F_DF);
--- a/src/dynarec/la64/dynarec_la64_00.c
+++ b/src/dynarec/la64/dynarec_la64_00.c
@@ -2481,7 +2481,10 @@ uintptr_t dynarec64_00(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             INST_NAME("CLC");
             SETFLAGS(X_CF, SF_SUBSET);
             SET_DFNONE();
-            BSTRINS_D(xFlags, xZR, F_CF, F_CF);
+            if (la64_lbt)
+                X64_SET_EFLAGS(xZR, X_CF);
+            else
+                BSTRINS_D(xFlags, xZR, F_CF, F_CF);
             break;
         case 0xFC:
             INST_NAME("CLD");
--- a/src/dynarec/la64/dynarec_la64_f30f.c
+++ b/src/dynarec/la64/dynarec_la64_f30f.c
@@ -336,14 +336,31 @@ uintptr_t dynarec64_F30F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int 
             SETFLAGS(X_ALL, SF_SET);
             SET_DFNONE();
             nextop = F8;
+            v1 = fpu_get_scratch(dyn);
             GETED(0);
             GETGD;
-            if (!rex.w && MODREG) {
-                ZEROUP2(x4, ed);
-                ed = x4;
+            if (MODREG) {
+                if (rex.w) {
+                    VXOR_V(v1, v1, v1);
+                    VINSGR2VR_D(v1, ed, 0);
+                } else {
+                    ZEROUP2(x4, ed); // need to clear uper part
+                    ed = x4;
+                    VXOR_V(v1, v1, v1);
+                    VINSGR2VR_D(v1, ed, 0);
+                }
+            } else {
+                if (rex.w) {
+                    SMREAD();
+                    addr = geted(dyn, addr, ninst, nextop, &ed, x1, x2, &fixedaddress, rex, NULL, 0, 0);
+                    FLD_D(v1, ed, fixedaddress);
+                } else {
+                    VXOR_V(v1, v1, v1);
+                    VINSGR2VR_D(v1, ed, 0);
+                }
             }
             CLEAR_FLAGS(x2);
-            BNE_MARK(ed, xZR);
+            BNEZ_MARK(ed);
             if (la64_lbt) {
                 ADDI_D(x3, xZR, 1 << F_ZF);
                 X64_SET_EFLAGS(x3, X_ZF);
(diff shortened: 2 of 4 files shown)


src/dynarec/la64/dynarec_la64_00.c
```diff
@@ -2477,6 +2477,12 @@ uintptr_t dynarec64_00(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                     DEFAULT;
             }
             break;
+        case 0xF8:
+            INST_NAME("CLC");
+            SETFLAGS(X_CF, SF_SUBSET);
+            SET_DFNONE();
+            BSTRINS_D(xFlags, xZR, F_CF, F_CF);
```
ksco, line 2484: This will break LBT.

src/dynarec/la64/dynarec_la64_f30f.c
```diff
@@ -331,6 +331,51 @@ uintptr_t dynarec64_F30F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+                    VLD(v1, ed, fixedaddress);
+                } else {
+                    GETED(0); // just load and clear the upper part
+                    VINSGR2VR_D(v1, ed, 0);
+                    VINSGR2VR_D(v1, xZR, 1);
+                }
+            }
+            CLEAR_FLAGS(x2);
+            BNEZ_MARK(ed);
+            ORI(xFlags, xFlags, 1 << F_ZF);
```
ksco, line 365: This will break LBT.

src/dynarec/la64/dynarec_la64_f30f.c
```diff
@@ -331,6 +331,51 @@ uintptr_t dynarec64_F30F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+            if (MODREG) {
+                if (rex.w) {
+                    VINSGR2VR_D(v1, ed, 0);
+                    VINSGR2VR_D(v1, xZR, 1);
+                } else {
+                    ZEROUP2(x4, ed); // need to clear uper part
+                    ed = x4;
+                    VINSGR2VR_D(v1, ed, 0);
+                    VINSGR2VR_D(v1, xZR, 1);
+                }
```
ksco, line 351: Clear v1 first with `VXOR_V` first, then you can use `VINSGR2VR_W/D` to do the job.

src/dynarec/la64/dynarec_la64_f30f.c
```diff
@@ -331,6 +331,51 @@ uintptr_t dynarec64_F30F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+            } else {
+                if (rex.w) {
+                    SMREAD();
+                    addr = geted(dyn, addr, ninst, nextop, &ed, x1, x2, &fixedaddress, rex, NULL, 0, 0);
+                    VLD(v1, ed, fixedaddress);
```
ksco, line 356: This is wrong.

ksco (changes requested):
The whole implementation of `POPCNT` seems to be a farfetched translation from the ARM64 backend, which I cannot approve.

ksco (changes requested):
Please use `vpcnt.d` to implement this one.

I believe that copying code from RV64 or ARM64 without fully understanding it might not be the best practice and could lead to issues down the line. I'll do my best to review the code to ensure it makes sense anyway, but it would help a lot if we can avoid blind copying from other backends.

Thanks.

xiangzhai:
> Please use vpcnt.d to implement this one.

Sorry for my stupid farfetched translation about `UADDLV` ...

src/dynarec/la64/dynarec_la64_f30f.c
```diff
@@ -331,6 +331,48 @@ uintptr_t dynarec64_F30F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+            GETGD;
+            if (MODREG) {
+                if (rex.w) {
+                    VXOR_V(v1, v1, v1);
+                    VINSGR2VR_D(v1, ed, 0);
+                } else {
+                    ZEROUP2(x4, ed); // need to clear uper part
+                    ed = x4;
+                    VXOR_V(v1, v1, v1);
+                    VINSGR2VR_D(v1, ed, 0);
```
ksco, line 350: Use the _W variant here, so you don’t have to clear upper bits.

src/dynarec/la64/dynarec_la64_f30f.c
```diff
@@ -331,6 +331,48 @@ uintptr_t dynarec64_F30F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+                }
+            } else {
+                if (rex.w) {
+                    SMREAD();
+                    addr = geted(dyn, addr, ninst, nextop, &ed, x1, x2, &fixedaddress, rex, NULL, 0, 0);
+                    FLD_D(v1, ed, fixedaddress);
+                } else {
+                    VXOR_V(v1, v1, v1);
+                    VINSGR2VR_D(v1, ed, 0);
```
ksco, line 359: This is still wrong, you can use FLD_S here.

src/dynarec/la64/dynarec_la64_f30f.c
```diff
@@ -331,6 +331,48 @@ uintptr_t dynarec64_F30F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+                }
+            }
+            CLEAR_FLAGS(x2);
```
ksco, line 362: CLEAR FLAGS does not make sense for lbt, this should be inside the else.

src/dynarec/la64/dynarec_la64_f30f.c
```diff
@@ -331,6 +331,48 @@ uintptr_t dynarec64_F30F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+            BNEZ_MARK(ed);
+            if (la64_lbt) {
+                ADDI_D(x3, xZR, 1 << F_ZF);
+                X64_SET_EFLAGS(x3, X_ZF);
```
ksco, line 366: C_ZF should be X_ALL, to set ZF but clear others.

src/dynarec/la64/dynarec_la64_f30f.c
```diff
@@ -331,6 +331,48 @@ uintptr_t dynarec64_F30F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+                    SMREAD();
+                    addr = geted(dyn, addr, ninst, nextop, &ed, x1, x2, &fixedaddress, rex, NULL, 0, 0);
+                    FLD_D(v1, ed, fixedaddress);
+                } else {
+                    VXOR_V(v1, v1, v1);
+                    VINSGR2VR_D(v1, ed, 0);
+                }
+            }
+            CLEAR_FLAGS(x2);
+            BNEZ_MARK(ed);
```
ksco, line 363: What this line is for? I suppose it’s for the flags calculation, but ed here is a memory address sometimes.

ksco (changes requested):

ksco (changes requested):
This is the fourth round of review and the implementation is not even closer to correct. You should consider giving up this one. I can do it later, then you'll see.

PR #2165, opened 2024-12-18
[GDBJIT] Refined RV64 and LA64 support

ksco:
Still not working on these two for some reason, but this is the progress for now.


src/tools/gdbjit.c
```diff
@@ -34,15 +34,15 @@ typedef struct gdbjit_descriptor_s {
 } gdbjit_descriptor_t;
 
 /* GDB puts a breakpoint in this function. This can't be optimized out. */
-void __attribute__((noinline)) __jit_debug_register_code()
+EXPORT void __attribute__((noinline)) __jit_debug_register_code()
```
ptitSeb, line 37: the noinline attribute is probably no longer needed with the `EXPORT` attribute.

PR #2222, opened 2024-12-28
Use uname -m to get CPU architecture

4Darmygeometry:
Use uname -m to get CPU architecture is simplier than lscpu

### DIFF
--- a/src/tools/my_cpuid.c
+++ b/src/tools/my_cpuid.c
@@ -152,7 +152,7 @@ const char* getCpuName()
         }
     }
     // failled, try to get architecture at least
-    f = popen("LC_ALL=C lscpu | grep -i \"architecture:\" | head -n 1 | sed -r 's/(architecture:)\\s{1,}//gi'", "r");
+    f = popen("uname -m", "r");
     if(f) {
         char tmp[200] = "";
         ssize_t s = fread(tmp, 1, 200, f);


src/tools/my_cpuid.c
```diff
@@ -152,7 +152,7 @@ const char* getCpuName()
         }
     }
     // failled, try to get architecture at least
-    f = popen("LC_ALL=C lscpu | grep -i \"architecture:\" | head -n 1 | sed -r 's/(architecture:)\\s{1,}//gi'", "r");
+    f = popen("uname -m", "r");
```
ptitSeb, line 155: historicaly, there was a hack using a fake `uname` to fool linux installer into thinking the arch was x86_64, and that was making this impossible to use. The hack is not usefull now, as box64 detect when a program to `uname -m` and redirect that to `box64 -m` to show "x86_64".

ksco: Hmmm, but I'm still using x86 bash from time to time, so I don't like this change.

ptitSeb: what do you mean @ksco ?

ksco: Sorry, I misunderstood.

ksco:
And why this is closed? I'll reopen it.

src/tools/my_cpuid.c
4Darmygeometry: I think Box64 can refer to the source code of fastfetch to obtain CPU architecture

4Darmygeometry: > historicaly, there was a hack using a fake `uname` to fool linux installer into thinking the machine arch was x86_64, and that was making this impossible to use. The hack is not usefull now, as box64 detect when a program to `uname -m` and redirect that to `box64 -m` to show "x86_64".

use file /usr/bin/uname to check if uname is x86_64 or not

use /usr/bin/uname -m instead of uname -m

PR #2233, opened 2025-01-05
Fix crash with incorrect argv[0]: rewrite it only if "box64" is present

ValdikSS:


src/core.c
```diff
@@ -2142,8 +2142,10 @@ int initialize(int argc, const char **argv, char** env, x64emu_t** emulator, elf
     {
         my_context->box86path = box_strdup(my_context->box64path);
         #ifndef BOX32
-        char* p = strrchr(my_context->box86path, '6');  // get the 6 of box64
-        p[0] = '8'; p[1] = '6'; // change 64 to 86
+        if (strstr(my_context->box86path, "box64")) {
+            char* p = strrchr(my_context->box86path, '6');  // get the 6 of box64
+            p[0] = '8'; p[1] = '6'; // change 64 to 86
+        }
```
ptitSeb, line 2148: I think there should be an "else" statement to free and set the value to NULL (with a null test in the following test).

ptitSeb: sometink more complete like:
```diff
+        if (strstr(my_context->box86path, "box64")) {
+            char* p = strrchr(my_context->box86path, '6');  // get the 6 of box64
+            p[0] = '8'; p[1] = '6'; // change 64 to 86
+        } else {
+            box_free(my_context->box86path);
+            my_context->box86path = ResolveFileSoft("box86", my_context->box64_path);
+        }
+        if(my_context->box86path && !FileExist(my_context->box86path, IS_FILE)) {
```

PR #2235, opened 2025-01-06
[WRAPPED] Added xdr_string and xdr_free

xiangzhai:
Hi,

Please review my patch.

Thanks,
Leslie Zhai


src/wrapped/wrappedlibc.c
```diff
@@ -3811,6 +3811,13 @@ EXPORT int my_xdr_string(x64emu_t* emu, void* arg1, char** arg2, unsigned int ar
 #endif
 }
 
+EXPORT void my_xdr_free(x64emu_t* emu, void* arg1, char* arg2)
+{
+#ifdef HAVE_RPC_XDR_H
```
ptitSeb, line 3816: what is this macro? were does it comes from?
my don't you use `my->xdr_free(arg1, arg2` instread, which will comes from tyhe lib openned.

ptitSeb: what happens if you don't have xdr stuffs and the programs use it?

ptitSeb:
I still don't understand this commit. Why using `GOM` and not `GO` for those 2 functions? What happened?

xiangzhai:
> Why using GOM and not GO for those 2 functions?

I added `HAVE_RPC_XDR_H` macro and `check_include_files` to test whether or not the OS provided `xdr`. And I also added `#ifdef HAVE_RPC_XDR_H` in the `my_xdr_string` for provided `xdr`. So I chose [GOM](https://github.com/ptitSeb/box64/blob/main/src/wrapped/wrappedlib_init.h#L41) to map for my_symbol. Perhaps I made simple question more complicated :rofl:  I should try `GO` instead.

Thanks,
Leslie Zhai

xiangzhai:
Thanks for your patient teaching :beers:

PR #2261, opened 2025-01-14
[CORE] Simplified xhead* detection and standardized extension display

ksco:


src/rv64detect.c
```diff
@@ -92,59 +92,18 @@ void RV64_Detect_Function()
...
-        TH_MULA(A0, A0, A1);
-        BR(xRA);
-        rv64_xtheadmac = Check(my_block);
-
-        // Test XTheadFmv with TH_FMV_X_HW
-        block = (uint32_t*)my_block;
-        TH_FMV_X_HW(A0, A1);
-        BR(xRA);
-        rv64_xtheadfmv = Check(my_block);
+        rv64_xtheadba
```
ptitSeb, line 98: No need to check individual instruction?if you have 1 extension, you have all of them?

ksco: Yes, always.

PR #2266, opened 2025-01-16
[ARM64_DYNAREC] Removed an unused arg in SET_DFNONE

ksco:


ptitSeb:
Thanks, That bother me each time I saw this (but not enough to actualy do it) :)

ksco:
> Thanks, That bother me each time I saw this (but not enough to actualy do it) :)

lmao

PR #2276, opened 2025-01-20
[RCFILE] Add MiSide Profile

KreitinnSoftware:


ptitSeb:
It would be intereting to check why avx slow things down: there might be some avx opcode that needs to be added in the dynarec probably.

PR #2281, opened 2025-01-22
Fixed various issues introduced in the env infra refactor

ksco:


src/core.c
```diff
@@ -538,23 +537,6 @@ void PrintHelp() {
     printf("    '-h'|'--help' to print this and quit\n");
 }
 
-void addNewEnvVar(const char* s)
```
ptitSeb, line 541: this is not used anymore? it was used by `BOX64_ENV` and `BOX64_ENV1` ... type of env. var. before

ksco: yeah, `BOX64_ENV*` are now handled at https://github.com/ptitSeb/box64/blob/main/src/tools/env.c#L145-L150

ksco: wait, that part seems badly wrong, lol

ksco: Will fix soon

PR #2291, opened 2025-01-24
[RV64_DYNAREC] Removed some TABLE64 usage

ksco:


ptitSeb:
Note that TABLE64 were reduce on ARM64 because, for an address in a 48bits address space, it takes up to 3 opcode (so 96 bytes) to directly load the adress, making the use of TABLE64 rarely usefull (unless the same address is used multiple time). Not sure the same reasoning happens here, especialy on RV64 where it might use more than 3 opcodes.

ksco:
Well, there is not that much of a difference in terms of code size, for a 48-bit address, it's 6 opcodes at max, and for a 39-bit address, it's 4 opcodes at max. But I decided to port this because `TABLE64` in `GETIP` seems not to be working well with bigblock>0, for some unknown reasons.

ksco:
Also, for the GETIP part, with an extra scratch register, it’s possible to use AUIPC+ADDI for any 32bit offset, that will come later ;)

ptitSeb:
> Also, for the GETIP part, with an extra scratch register, it’s possible to use AUIPC+ADDI for any 32bit offset, that will come later ;)

You will probably not be able to use that, I think. First because it's a 32bits offset between x86 and rv64 PC, which is unlikely (well, on 39bits address space, it's less unlikly of couse). Second, and more importan:  because the RV64 address is only known in PASS3, making it difficult to use this in PASS2 without creating difference between those 2 passes.

ksco:
Sorry, I meant LUI, not AUIPC

PR #2297, opened 2025-01-27
[RV64_DYNAREC] Fixed scalar CVTPD2DQ

xctan:


ptitSeb:
How is that a fix? it looks like the same, just unrolled loop?

xctan:
Sorry, the narrowing conversion is feasible with a loop for in place conversions, so the old code should work.

PR #2307, opened 2025-02-01
[ENV] Fixed crash when rcfile does not exist (for #2295)

ksco:
Fixed #2295

PR #2309, opened 2025-02-01
[ENV] More tweaks on recoding env mapping

ksco:


ksco:
Related to https://github.com/ptitSeb/box64/pull/2307. Actually, this matters because `box64env_entries` and `box64env_entries_gen` both need to be initialized.

src/tools/env.c
```diff
@@ -548,6 +548,7 @@ static kh_mapping_entry_t* mapping_entries = NULL;
 
 void RecordEnvMappings(uintptr_t addr, size_t length, int fd)
 {
+    if (!box64env_entries) return;
```
ptitSeb, line 551: Oh now, not like that. We don't want to remove the file mapped trcking for this reason. Because file map will have much more than just profile later on!

ksco: You mean the aot cache? okay...

ptitSeb: Yeah, that. But before that, the dynarec memory...

ksco: is it okay now?

ptitSeb: yep. just waiting for the CI.

ptitSeb (approved):

PR #2311, opened 2025-02-02
Added auto-generated x64Int3 printer

ksco:


rebuild_wrappers.py
```diff
@@ -620,7 +630,7 @@ def add_symbol_name(funname: Union[str, None], funsname: Dict[ClausesStr, List[s
 		print("(Also, the program WILL crash later if you proceed.)")
 		sys.exit(2) # Check what you did, not proceeding
 	
-	return gbl, redirects, filespec
+	return gbl, redirects, filespec, alltypedefs
```
rajdakin, line 633: You forgot to update the return signature.
(Also, the file should be accepted by `mypy` if you comment the imports you don't use at the beginning of the file, and this error should have been caught by this tool.)

rebuild_wrappers.py
```diff
@@ -815,8 +825,9 @@ def main(root: str, files: Iterable[Filename], ver: str):
 	gbl_tmp: JumbledGlobals
 	red_tmp: JumbledRedirects
 	fsp_tmp: JumbledFilesSpecific
+	alltypedefs: JumbledTypedefs
 	
-	gbl_tmp, red_tmp, fsp_tmp = readFiles(files)
+	gbl_tmp, red_tmp, fsp_tmp, alltypedefs = readFiles(files)
```
rajdakin, line 830: Please also sort the typedef structure (in the `sortArrays` function, similar to the first element of the values of `fsp_tmp`), so that the order in which the content of `x64printer.c` is generated stays constant from one run to another.

rebuild_wrappers.py
```diff
@@ -1730,7 +1774,25 @@ def function_writer(f, N: FunctionType, W: str) -> None:
...
+		file.write(files_header["x64printer.c"].format(lbr="{", rbr="}", version=ver))
+		
+		signature: RedirectType = ""
+		for signature in alltypedefs:
+			if any(letter in signature[2:] for letter in "OSNxXYb"):
+				continue
+			formats = signature[2:].replace("G", "UU")
+			params = '' if formats == "v" else ", " + function_args_systemV(formats)[:-2]
+			formats = '' if formats == "v" else ', '.join("%\" PRI{x} \"".format(x=formats_map[signature.get_convention().values.index(t)]) for t in formats)
+			file.write("    {rbr} else if ({condition}) {lbr}\n".format(condition=" || ".join(["!strcmp(func, \"" + func + "\")" for func in alltypedefs[signature]]), lbr="{", rbr="}"))
```
rajdakin, line 1791: Instead of comparing the function name, maybe try to compare the function called? So you compare `w` with `&iFv` instead of `s` (which you called `fnc` here) against every function name which takes no argument and return an int. That would also reduce the amount of instruction it adds (50M is a lot! I'd think a lot of this is simply storing the mapping as there are a _lot_ of functions.)

ksco: yeah, that's why I converted this into a draft, working on it.

ptitSeb (approved):

rajdakin (approved):
Looks good to me now.

ksco:
There are som type errors reported by mypy, I'll push a fix soon.

PR #2316, opened 2025-02-06
[RCFILE] Don't apply per-file settings to the main binary

ksco:


ptitSeb:
not sure I understand. The name deduced with the file descriptor should have a `/` in front, and so always reference a dll/mainbin.
But the actual main binary resolution and setting application is done elsewhere.
And we can also imagine a case were both `[game.exe]` and `[/game.exe]` exist for some twisted reason.

ksco:
This only applies to Linux binaries, the mainbin will be `wine` for Windows applications. I found it confusing to have both `[/factorio]` and `[factorio]` exist, but sure, there is no harm in keeping the possibility.

PR #2342, opened 2025-02-11
fix VPMASKMOV and VMASKMOVP

ye-yeshun:
Just like 0x8c, I've made changes to the relevant code to avoid accessing out-of-bounds memory when a portion of the mask is 0.
I added the test case but commented it out, I don't think I should tamper with ref30.txt and the binary file test30.

### DIFF
--- a/src/dynarec/arm64/dynarec_arm64_avx_66_0f38.c
+++ b/src/dynarec/arm64/dynarec_arm64_avx_66_0f38.c
@@ -1278,31 +1278,65 @@ uintptr_t dynarec64_AVX_66_0F38(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip
             nextop = F8;
             q0 = fpu_get_scratch(dyn, ninst);
             for(int l=0; l<1+vex.l; ++l) {
-                if(!l) {
-                    GETGX(v0, 0); GETVX(v2, 0);
-                    if(MODREG) {
+                if(MODREG) {
+                    if(!l) {
+                        GETGX(v0, 0); GETVX(v2, 0);
                         s0 = (nextop&7)+(rex.b<<3);
                         v1 = sse_get_reg_empty(dyn, ninst, x1, s0);
                     } else {
+                        GETGY(v0, 0, vex.v, s0, -1); v2 = ymm_get_reg(dyn, ninst, x1, vex.v, 0, gd, s0, -1);
+                        v1 = ymm_get_reg_empty(dyn, ninst, x1, s0, gd, vex.v, -1);
+                    }
+                    if(rex.w)
+                        VSSHRQ_64(q0, v2, 63);
+                    else
+                        VSSHRQ_32(q0, v2, 31);
+                    VBITQ(v1, v0, q0);
+                } else {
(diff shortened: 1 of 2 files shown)


src/dynarec/arm64/dynarec_arm64_avx_66_0f38.c
```diff
@@ -1141,8 +1257,15 @@ uintptr_t dynarec64_AVX_66_0F38(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip
...
                         CBZx(x4, 4+1*4);
                         VLD1_64(v1, 1, ed);
-                        if(!l && vex.l)
-                            ADDx_U12(ed, ed, 8);
+                        if(vex.l)
+                        {
+                            if (!l)
+                                ADDx_U12(ed, ed, 8);
+                            else
+                                SUBx_U12(ed, ed, 24);
```
ptitSeb, line 1265: The SUB should be useless, "ed" is never an x86 register as it's copied to `x3` if needed just after the `geted(...)`

ptitSeb: Note that the version you added don't have this check after geted, you might do that instead of "restoring" ed, else, it would make signal handling more complex.

ptitSeb (changes requested):
You need to check if `ed` is `x3` after the `geted(...)` so you don't need to restore "ed" to it's initial value. The `SUB(ed, ed, 24)`  and `SUB(ed, ed, 8)` can also be removed after that.

ptitSeb:
Side question: did you encounter program/games that needs this?

src/dynarec/arm64/dynarec_arm64_avx_66_0f38.c
ye-yeshun: I removed the SUB of "ed", and fixed a bug on clearing "v0".

ye-yeshun:
> Side question: did you encounter program/games that needs this?

Sure. Opening the email page of WXWork software and switching emails several times will cause an exception.
After debugging and analysing, I found that it was the MASKMOV instruction accessing an unallocated memory page at a memory page boundary.

ye-yeshun:
> You need to check if `ed` is `x3` after the `geted(...)` so you don't need to restore "ed" to it's initial value. The `SUB(ed, ed, 24)` and `SUB(ed, ed, 8)` can also be removed after that.

Due to the number of changes, I used a forced push, please re-review

ptitSeb:
It seems this PR has broken "The Talos Principle 2" (an UE5+ game). I'll debug tomorrow.

ye-yeshun:
> It seems this PR has broken "The Talos Principle 2" (an UE5+ game). I'll debug tomorrow.

OK, thank you. If you get anything after debugging, please comment here.

ptitSeb:
> > It seems this PR has broken "The Talos Principle 2" (an UE5+ game). I'll debug tomorrow.
> 
> OK, thank you. If you get anything after debugging, please comment here.

Yes, the issue was that "ed" was not check for being "x3". It's fixed.

PR #2344, opened 2025-02-11
[ARM64_DYNAREC] Fixed wrong COUNT operand in VPSLL{W,D,Q}

wannacu:


ptitSeb:
Ah, indeed, count is always a 128bits reg, even on 256bits operation! Nice catch.

PR #2347, opened 2025-02-12
[ARM64_DYNAREC] Added some BMI.0F38 opcodes

wannacu:


src/dynarec/arm64/dynarec_arm64_avx_0f38.c
```diff
@@ -137,6 +137,83 @@ uintptr_t dynarec64_AVX_0F38(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, i
...
+                        TSTxw_REG(ed, ed);
+                        CSETMw(x3, cEQ);
+                        BFIw(xFlags, x3, F_CF, 1);
+                    }
+                    SUBxw_U12(x3, ed, 1);
+                    need_tst = 0;
+                    IFX(X_ZF) need_tst = 1;
+                    IFXNATIVE(X_SF, NF_SF) need_tst = 1;
+                    IFXNATIVE(X_OF, NF_VF) need_tst = 1;
+                    if(need_tst)
```
ptitSeb, line 156: EOR does not update flags. So thi smakes no sence.

Are using AI assisted tools to create those PR?

src/dynarec/arm64/dynarec_arm64_avx_0f38.c
```diff
@@ -150,13 +227,14 @@ uintptr_t dynarec64_AVX_0F38(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, i
             GETED(0);
             GETVD;
             UXTBw(x1, vd);
-            CMPSw_U12(x1, rex.w?64:32);
-            CSETxw(x2, cPL);
+            CMPSw_U12(x1, rex.w?63:31);
+            CSETxw(x2, cGT);
```
ptitSeb, line 231: are you sure?

src/dynarec/arm64/dynarec_arm64_avx_0f38.c
```diff
@@ -150,13 +227,14 @@ uintptr_t dynarec64_AVX_0F38(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, i
...
             IFX(F_CF) {
                 BFIw(xFlags, x2, F_CF, 1);
             }
-            MVNxw_REG(x2, x2); //prepare mask
-            B_MARK(cPL);
+            MOV64xw(x2, 0); //prepare mask
+            B_MARK(cGT);
+            MOV64xw(x2, -1);
```
ptitSeb, line 237: why this?

src/dynarec/arm64/dynarec_arm64_avx_0f38.c
```diff
@@ -184,6 +262,45 @@ uintptr_t dynarec64_AVX_0F38(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, i
...
+            B_MARK(cEQ);
+            LSRxw_REG(x1, ed, x2);
+            CMPSw_U12(x3, rex.w?64:32);
+            B_MARK(cGE);
+            MOV32w(x2, rex.w?64:32);
+            SUBw_REG(x2, x2, x3);
+            LSLxw_REG(x1, x1, x2);
+            LSRxw_REG(x1, x1, x2);
+            need_tst = 0;
+            IFXNATIVE(X_SF, NF_SF) need_tst = 1;
```
ptitSeb, line 285: there is no need to use need_test if the condition is that simple.

src/dynarec/arm64/dynarec_arm64_avx_0f38.c
ptitSeb: The simple way to fix that is to simple do the EOR without condition, and if `need_test` add a conditionnal `TSTxw` after that, to compute the flags.

src/dynarec/arm64/dynarec_arm64_avx_0f38.c
ptitSeb: I was testing `x1>=64` (or 32), and you changed it to `x1>63` (or 31). That's strictly the same thing, so I don't see why it should be changed.

src/dynarec/arm64/dynarec_arm64_avx_0f38.c
ptitSeb: The code that was there is stricly equivalent to the one you put, but shorter.

x2 is `-1` if x1>=64, while it's `0` else, because of the CSET before...
after the MVN it become `0` if x1>=64, and `-1` else
then it jump to MARK if x1>=64

src/dynarec/arm64/dynarec_arm64_avx_0f38.c
ptitSeb: What I mean is this
```C
IFXNATIVE(X_SF, NF_SF) 
    TSTxw_REG(x1, x1);
```
is enough, no need to use need_test here.

src/dynarec/arm64/dynarec_arm64_avx_0f38.c
ptitSeb: Ah, the mystake is that it should be `CSETM` instead of `CSET` for this to work. (CSETM set all bits to 1, while CSET set the value to 1)

src/dynarec/arm64/dynarec_arm64_avx_0f38.c
```diff
@@ -184,6 +260,42 @@ uintptr_t dynarec64_AVX_0F38(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, i
...
+            TSTw_REG(x3, x3);
+            B_MARK(cEQ);
+            LSRxw_REG(x1, ed, x2);
+            CMPSw_U12(x3, rex.w?64:32);
+            B_MARK(cGE);
+            MOV32w(x2, rex.w?64:32);
+            SUBw_REG(x2, x2, x3);
+            LSLxw_REG(x1, x1, x2);
+            LSRxw_REG(x1, x1, x2);
+            IFXNATIVE(X_SF, NF_SF)
```
ptitSeb, line 282: According to Intel doc, SF is undefined. Only ZF is set according to result.

ptitSeb (approved):
For BEXTR opcode, ZF needs to be computed. SF is undefeined, but computing real SF is fine I suppose.

ptitSeb (changes requested):
For BEXTR opcode, Intel doc says ZF needs to be computed from result. SF is undefined (but computing it is fine)

ptitSeb:
> > For BEXTR opcode, ZF needs to be computed. SF is undefeined, but computing real SF is fine I suppose.
> 
> ```diff
> -            IFXNATIVE(X_SF, NF_SF)
> -                TSTxw_REG(x1, x1);
> +            TSTxw_REG(x1, x1);
>              MARK;
>              MOVxw_REG(gd, x1);
> -            IFX(X_SF) {
> -                IFNATIVE(NF_SF) {} else {
> -                    LSRxw(x3, gd, rex.w?63:31);
> -                    BFIw(xFlags, x3, F_SF, 1);
> +            IFX(X_ZF) {
> +                IFNATIVE(NF_EQ) {} else {
> +                    CSETw(x3, cEQ);
> +                    BFIw(xFlags, x3, F_ZF, 1);
>                  }
>              }
>              IFX(X_AF) BFCw(xFlags, F_AF, 1);
>              IFX(X_PF) BFCw(xFlags, F_PF, 1);
>              IFX(X_CF) BFCw(xFlags, F_CF, 1);
>              IFX(X_OF) IFNATIVE(NF_VF) {} else {BFCw(xFlags, F_OF, 1);}
> -            IFX(X_ZF) IFNATIVE(NF_EQ) {} else {BFCw(xFlags, F_ZF, 1);}
> +            IFX(X_SF) IFNATIVE(NF_SF) {} else {BFCw(xFlags, F_SF, 1);}
> ```
> 
> should change like this? and I have another question: Does the eflag of undef need to be removed in SETFLAGS to avoid performing calculations?

Yes, that would be OK.

Undefined means they are computed / altered. So yes, this opcode will change ALL eflags, even if some are undefined.

src/dynarec/arm64/arm64_emitter.h
```diff
@@ -582,7 +582,7 @@ int convert_bitmask(uint64_t bitmask);
 #define CSINV_gen(sf, Rm, cond, Rn, Rd)     ((sf)<<31 | 1<<30 | 0b11010100<<21 | (Rm)<<16 | (cond)<<12 | (Rn)<<5 | (Rd))
 #define CSINVx(Rd, Rn, Rm, cond)            EMIT(CSINV_gen(1, Rm, cond, Rn, Rd))
 #define CSINVw(Rd, Rn, Rm, cond)            EMIT(CSINV_gen(0, Rm, cond, Rn, Rd))
-#define CSINVxw(Rd, Rn, Rm, cond)           EMIT(CSINV_gen(rex.w?, Rm, cond, Rn, Rd))
+#define CSINVxw(Rd, Rn, Rm, cond)           EMIT(CSINV_gen(rex.w?1:0, Rm, cond, Rn, Rd))
```
ptitSeb, line 585: sidenot, just `rex.w` is enough, it's already 1 bit 1 or 0.

ptitSeb (approved):

PR #2357, opened 2025-02-13
[COSIM] Print decoded instruction mnemonics if trace enabled

ksco:


ksco:
Oops, I accidentally deleted the branch.

PR #2358, opened 2025-02-14
[ARM64_DYNAREC] Add some opcodes

wannacu:


src/dynarec/arm64/dynarec_arm64_66f30f.c
```diff
@@ -55,6 +55,58 @@ uintptr_t dynarec64_66F30F(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int
...
     switch(opcode) {
+        case 0xB8:
+            INST_NAME("POPCNT Gw, Ew");
+            SETFLAGS(X_ALL, SF_SET);
+            SET_DFNONE();
+            nextop = F8;
+            GETGW(x2);
+            GETEW(x1, 0);
+            v1 = fpu_get_scratch(dyn, ninst);
+            VMOVQDfrom(v1, 0, ed);
```
ptitSeb, line 66: you might need to VEOR v1 first, as this just replace 1 element on the vector, leving the other untouched.

src/dynarec/arm64/dynarec_arm64_66f30f.c
```diff
@@ -63,19 +115,21 @@ uintptr_t dynarec64_66F30F(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int
...
             GETGW(x2);
+            TSTxw_REG(ed, ed);
             IFX(X_CF) {
-                TSTxw_REG(ed, ed);
                 CSETw(x3, cEQ);
                 BFIw(xFlags, x3, F_CF, 1);  // CF = is source 0?
             }
             LSLw_IMM(ed, ed, 16);
             CLZw(gd, ed);
+            MOV32w(x3, 16);
```
ptitSeb, line 125: another solution would be the set bit 15 as 1, with an OR_mask, but his solution works too and fixes the issue

ptitSeb: Or just not LSL and substract 16 to the result. might be even simpler.

src/dynarec/arm64/dynarec_arm64_avx_f2_0f38.c
```diff
@@ -60,7 +60,36 @@ uintptr_t dynarec64_AVX_F2_0F38(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip
...
+            // x3 = mask of mask, loop while not 0
+            MOV32w(gd, 0);
+            MOV64x(x2, 1);
+            MOV64x(x3, 1);
+            MARK;
+            TSTxw_REG(ed, x3);
+            B_MARK2(cEQ);
+            TSTxw_REG(vd, x2);
+            B_MARK3(cEQ);
+            ORRxw_REG(gd, gd, x3);
```
ptitSeb, line 84: should be x2 there instead of x3

src/dynarec/arm64/dynarec_arm64_avx_f2_0f38.c
```diff
@@ -60,7 +60,36 @@ uintptr_t dynarec64_AVX_F2_0F38(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip
...
+            MARK3;
+            LSLxw_IMM(x2, x2, 1);
```
ptitSeb, line 86: the LSL x2 should be before the MARK3

src/dynarec/arm64/dynarec_arm64_avx_f2_0f38.c
```diff
@@ -60,7 +60,36 @@ uintptr_t dynarec64_AVX_F2_0F38(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip
...
+                gb1 = 0;
+            }
+            // x3 = mask of mask, loop while not 0
+            MOV32w(gd, 0);
+            MOV64x(x2, 1);
+            MOV64x(x3, 1);
+            MARK;
+            TSTxw_REG(ed, x3);
+            B_MARK2(cEQ);
+            TSTxw_REG(vd, x2);
```
ptitSeb, line 82: this should be x3 instead of x2

ptitSeb: Oh, my bad, I mixed up with another opcode

PR #2366, opened 2025-02-16
[LA64_DYNAREC] Added more opcodes

ksco:


ptitSeb:
Something is wrong with the sse 4.2 test. The CRC32 part probably?

ksco:
I tested locally and it passed. Maybe QEMU does not support that opcode. I’ll have a look later.

PR #2372, opened 2025-02-17
[ARM64_DYNAREC] Added more AVX opcodes

wannacu:


ptitSeb:
Test29 "avx_intrinsics" is failling. Some newly added AVX is not correct.

src/dynarec/arm64/dynarec_arm64_67_avx.c
```diff
@@ -126,6 +126,20 @@ uintptr_t dynarec64_67_AVX(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int
                 }
                 break;
 
+            case 0x6E:
+                INST_NAME("VMOVD Gx, Ed");
+                nextop = F8;
+                GETGX_empty(v0);
+                GETED(0);
```
ptitSeb, line 133: it should be `GETED32(0)` here, as this is a `67` prefixed opcode.

src/dynarec/arm64/dynarec_arm64_67_avx.c
```diff
@@ -126,6 +126,20 @@ uintptr_t dynarec64_67_AVX(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int
...
+                if(rex.w) {
+                    FMOVDx(v0, ed);
+                } else {
+                    VEORQ(v0, v0, v0); // RAZ vector
```
ptitSeb, line 137: Why not `FMOVSw(...)` here if `FMOVDx(...)` is good enough for `rex.w` case?

ptitSeb: You need the `VEORQ(...)` on both branch of the if actually, because you are setting `Gx` here and not `Gm`.

I suggest you use FMOV or VMOVQ on both branch also. Mayby favor `VMOVQ` has it's a `Gx` here...

PR #2380, opened 2025-02-17
[BUNDLE] Add script to bundle x86 libraries

LukeShortCloud:
This manages both x86_64 and x86 libraries. It helps with improving emulation in all of the Box projects.


LukeShortCloud:
I've been using these libraries primarily to test Box32 but they can help with Box64 and Box86, too. Games I've been testing include: Amnesia: The Dark Descent, SOMA, SuperTuxKart (0.9.2, which still used OpenGL), and a few others I can't remember right now.

If/when this is merged, next steps include:
- Add more x86[_64] libraries based on user reports from GitHub Issues
- Add CMake integration
- Add the `box64-bundle-x86-libs.tar.gz` archive to the next Box64 release on GitHub.

LukeShortCloud:
Also, these are not just random libraries that I started with. At least one game or another requires everything I have bundled right now.

ksco:
Can we add the .so files in x64lib and x86lib to the bundle too? So we don't have to host those files directly in the repo.

LukeShortCloud:
Yes, I'll make sure to add those in my next follow-up PR! :-)

PR #2381, opened 2025-02-18
[RV64_DYNAREC] Optimized unaligned path for REP MOVSB

ksco:


src/dynarec/rv64/dynarec_rv64_00_2.c
```diff
@@ -657,18 +657,42 @@ uintptr_t dynarec64_00_2(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
                 CBZ_NEXT(xRCX);
                 ANDI(x1, xFlags, 1 << F_DF);
                 BNEZ_MARK2(x1);
-                IF_ALIGNED (ip) {
+                IF_UNALIGNED (ip) {
```
ptitSeb, line 660: I don't understand. The aligned and unaligned path are the same? What is the point of unaligned?

ksco: it's not the same, unaligned path checks RSI and RDI before entering the hot loop.

ksco: bah, I removed the unaligned path. If it's proved beneficial, I'll add it back.

ptitSeb: unaligned on RiscV was only usefull on the lastest Sophgo CPU iirc, so it might be tricky to test it anyway.

ksco: (it's useful on the VisionFive 2 with an external GPU too.)

ptitSeb: well, on arm64, the unaligned "fast" path also only move 4-bytes at a time instead of 8-bytes on aligned path

ksco: yeah, I noticed that, but for what?

ptitSeb: because this code only exist when a SIGBUS is trigger, so unligned happens. Using a less aggressive "fast" path makes sense so it's actually taken more often, even if a bit slower.

Remember Unaligned is only generated when a SIGBUS happenned for this opcode address at least once.

PR #2386, opened 2025-02-18
[CORE] Display Box64 version info in mangohud

ksco:
image


ptitSeb:
What happen if you have custom config with `~/.config/MangoHud/MangoHud.conf` for example?

ksco:
So we will also need to check if those config files exist.

ptitSeb:
well, it's never to late for another PR....

PR #2397, opened 2025-02-20
Made some tweaks to banner and help message

ksco:


ptitSeb:
All red on the CI?

ksco:
yeah because I broke nobanner, fixed.

ptitSeb:
Ah. nobanner is very important, for when script are run and redirected for example, to avoid box64 interefering with actual program output.

PR #2419, opened 2025-03-05
Bundle more libs

LukeShortCloud:


LukeShortCloud:
The total size of x86_64 and x86 libraries combined is now 79 MiB. Still much better than 1 GiB for a complete rootfs!

LukeShortCloud:
I haven't modified a CMake project in a long time. Seb, maybe you can help me out with that. I'd like to have a follow-up to this that'll automatically `make install` these libraries if the tarball `box64-bundle-x86-libs.tar.gz` is found.

box64-bundle-x86-libs.csv
```diff
@@ -1,15 +1,37 @@
...
 https://repo.almalinux.org/almalinux/9.5/BaseOS/x86_64/os/Packages/bluez-libs-5.72-2.el9.i686.rpm,af1aba8e6fa5ccf33032d58a3acc59368c07d532293396db205fb9a4bc1241ae
 https://repo.almalinux.org/almalinux/9.5/BaseOS/x86_64/os/Packages/bluez-libs-5.72-2.el9.x86_64.rpm,49c273813b4616240acdae8dae893192658c7541482f9bbb2c46dfe19b61877d
 https://repo.almalinux.org/almalinux/9.5/AppStream/x86_64/os/Packages/fribidi-1.0.10-6.el9.2.i686.rpm,75f808f0f5f93aceb4e477aa5f302edbed07a605a0644f0ea929d4ea609e36a0
 https://repo.almalinux.org/almalinux/9.5/AppStream/x86_64/os/Packages/fribidi-1.0.10-6.el9.2.x86_64.rpm,b4763ca410f8479b2775efb55ba09e6722948af7526180a56309e3e090c27010
+https://repo.almalinux.org/almalinux/9.5/BaseOS/x86_64/os/Packages/glib2-2.68.4-14.el9_4.1.i686.rpm,e651d60f1906415554e34e21ce414b7eea9f8bd0316ca8a320a6d9da98b2a44d
+https://repo.almalinux.org/almalinux/9.5/BaseOS/x86_64/os/Packages/glib2-2.68.4-14.el9_4.1.x86_64.rpm,2309263c8fb1e3b732eb643bd243cbd7bbe73ed6a58851e8661f19ef41ba7128
+https://repo.almalinux.org/almalinux/9.5/BaseOS/x86_64/os/Packages/glibc-2.34-125.el9_5.1.alma.2.i686.rpm,72c2b7fcf768a0c09badc8453686b8ecdacdf9f5fb0ddc71a4892150b6093ec4
```
ptitSeb, line 9: emulated glibc library will be ignored, no mater the settings you use on box64.

box64-bundle-x86-libs.csv
```diff
@@ -20,19 +42,43 @@ https://repo.almalinux.org/almalinux/9.5/AppStream/x86_64/os/Packages/libogg-1.3
...
+https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/34/Everything/x86_64/os/Packages/l/libunwind-1.4.0-5.fc34.i686.rpm,eb518e9d073c36fd29862c7afd00a8a7cb186074a0e162983f18fb6c9dc71d2c
+https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/34/Everything/x86_64/os/Packages/l/libunwind-1.4.0-5.fc34.x86_64.rpm,0f44e37e4eb472508e0dba8505ebbb04c2c1390c6a43970a6af80aad886037e3
 https://repo.almalinux.org/almalinux/9.5/AppStream/x86_64/os/Packages/libvorbis-1.3.7-5.el9.i686.rpm,e38c5185fea611330ea253db020315dd564507bf95c48a81dbe7db186ce4214b
 https://repo.almalinux.org/almalinux/9.5/AppStream/x86_64/os/Packages/libvorbis-1.3.7-5.el9.x86_64.rpm,3c340337a073975636d198a282503f32a54bf85fb5345522a2c22e530b8d6c9e
+https://repo.almalinux.org/almalinux/9.5/AppStream/x86_64/os/Packages/libxcb-1.13.1-9.el9.i686.rpm,a984b3ea130f77d739dc763bcc37c008573df9fc069ae457421c59b8ad692f8b
+https://repo.almalinux.org/almalinux/9.5/AppStream/x86_64/os/Packages/libxcb-1.13.1-9.el9.x86_64.rpm,840ea71fb8beb52093e9f6d23dd0bc7de86f9d6da8c48888d766b0915cf1e1bc
```
ptitSeb, line 64: emulated libxcb (and libX11-xcb) will be ignored by box64, no matter the settings.

ptitSeb:
You can look at `library.c` line 427 for the list of library that will never be emulated: https://github.com/ptitSeb/box64/blob/main/src/librarian/library.c#L427

LukeShortCloud:
Thanks for all of the tips here and in the related GitHub Issue. I wasn't aware of the libraries that won't be used or the "all or nothing" libraries. I'll work on an update later today.

box64-bundle-x86-libs.csv
LukeShortCloud: Does this also apply to all Xorg and Wayland libraries? Or just XCB from Xorg?

ptitSeb: it's about coherency... if a lib is used as a dependancy on an emulated lib and a wrapped lib, and it's used emulated on box64, that means the 2 version (native and emulated) will be loaded and that will not work well. For this reason, it's better to avoid emulated xcb libs. Xorg one... well, I prefer to use 100% wrapped but it's more of a gray area here.
For Wayland, I'm unsure... but I would expect the libs to present in the system for user that want to use them.

LukeShortCloud:
I updated the documentation with all of the tips you shared with me. I also updated the CSV database. Attached is a list of all of the library files that are currently bundled. Could you double check this to make sure there are no more problematic libraries that I'm bundling here?

[box64-bundle-more-libs-2419.txt](https://github.com/user-attachments/files/19111974/box64-bundle-more-libs-2419.txt)

LukeShortCloud:
I removed one more lingering XCB package.

Here's the latest list of libraries bundled with this PR:

[box64-bundle-more-libs-2419-20250312.txt](https://github.com/user-attachments/files/19222342/box64-bundle-more-libs-2419-20250312.txt)

@ptitSeb Everything should be good now unless you have any other last minute concerns.

PR #2426, opened 2025-03-07
[RV64_DYNAREC] Rollback some falsy optimization in the xtheadvector path

ksco:


OFFTKP:
It doesn't work on XTheadVector?

ksco:
> It doesn't work on XTheadVector?

It can work, but the register used in `vmslt.vx` must be aligned to `LMUL2` or `LMUL8` because of the design difference of the mask register. We can use a scratch and `vmv.vv` when it's not aligned, but that makes no difference from the original code regarding instructions numbers.

PR #2450, opened 2025-03-19
[RCFILE] Added profile for HSR Game

shenmo7192:


system/box64.box64rc
```diff
@@ -758,6 +758,9 @@ BOX64_DYNAREC_STRONGMEM=1
 BOX64_DYNAREC_SAFEFLAGS=2
 BOX64_DYNAREC_CALLRET=0
 
+[StarRail.exe]
```
ksco, line 761: Would be better to keep alphabetical.

ksco: Ah looks like the alphabetical order has been disrupted already, never mind.

shenmo7192: Both StarRail and Zenless Zone Zero is an mihoyo game. I thought it was categorized by game producers. I'll change it

shenmo7192: > Ah looks like the alphabetical order has been disrupted already, never mind.

Okay then i will keep it here ~ XD

PR #2476, opened 2025-03-28
[LA64_DYNAREC] Added preliminary MMX support

ksco:


src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -820,6 +820,12 @@ int sse_setround(dynarec_la64_t* dyn, int ninst, int s1, int s2)
     return s2;
 }
 
+
+void x87_purgecache(dynarec_la64_t* dyn, int ninst, int next, int s1, int s2, int s3)
+{
+    // TODO
```
ptitSeb, line 826: :O don't that break games?

ksco: well, there is no x87 support, so nothing to do for now, no?

ptitSeb: Ah yeah, fair point.

PR #2489, opened 2025-04-01
[LA64_DYNAREC] Fix f0.80.1 LOCK OR, LoongArch  ANDI is zero-extended.

phorcys:


ksco (approved):
Thank you for the fix!

PR #2494, opened 2025-04-02
Moved some emit functions to os.h

ksco:


src/dynarec/dynarec_native_functions.c
```diff
@@ -185,42 +186,37 @@ void native_ud(x64emu_t* emu)
...
+    EmitSignal(emu, SIGSEGV, (void*)R_RIP, 0xbad0);
 }
 
 void native_int(x64emu_t* emu, int num)
 {
     emu->test.test = 0;
-    emit_interruption(emu, num, (void*)R_RIP);
-}
-
-void native_singlestep(x64emu_t* emu)
```
ptitSeb, line 210: this is not used?

ksco: yeah it’s used before but not anymore

ptitSeb: ok

PR #2497, opened 2025-04-03
Moved more functions to os.h

ksco:
Also deleted some redundant inclusion of `emu/x64run_private.h`

### DIFF
--- a/src/dynarec/arm64/dynarec_arm64_00.c
+++ b/src/dynarec/arm64/dynarec_arm64_00.c
@@ -2578,7 +2578,7 @@ uintptr_t dynarec64_00(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
                         STORE_XEMU_CALL(xRIP);
                         ADDx_U12(x3, xRIP, 8+8);    // expected return address
                         ADDx_U12(x1, xEmu, (uint32_t)offsetof(x64emu_t, ip)); // setup addr as &emu->ip
-                        CALL_(NativeCall, -1, x3);
+                        CALL_(EmuInt3, -1, x3);
                         SMWRITE2();
                         LOAD_XEMU_CALL(xRIP);
                         addr+=8+8;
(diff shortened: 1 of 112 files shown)


src/dynarec/arm64/dynarec_arm64_00.c
```diff
@@ -2578,7 +2578,7 @@ uintptr_t dynarec64_00(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
                         STORE_XEMU_CALL(xRIP);
                         ADDx_U12(x3, xRIP, 8+8);    // expected return address
                         ADDx_U12(x1, xEmu, (uint32_t)offsetof(x64emu_t, ip)); // setup addr as &emu->ip
-                        CALL_(x64Int3, -1, x3);
```
ptitSeb, line 2581: I am neatpeeking here, but x64Int3 is not only for Native Call, but also for guenuine Int 3 `CC` opcodes.

src/include/os.h
```diff
@@ -1,12 +1,23 @@
...
 void EmitDiv0(void* emu, void* addr, int code);
 void EmitInterruption(void* emu, int num, void* addr);
 
+// These functions only applies to Linux
+int IsBridgeSignature(char s, char c);
+int IsNativeCall(uintptr_t addr, int is32bits, uintptr_t* calladdress, uint16_t* retn);
+void NativeCall(void* emu, void* addr);
+void* EmuFork(void* emu, int forktype);
+
+void EmuX64Syscall(void* emu);
```
ptitSeb, line 18: I am sure  Syscall also exist on Windows.

ptitSeb: And Wine handle it, at least on the 64bits side. I think it's ignored on 32bits

src/dynarec/arm64/dynarec_arm64_00.c
ksco: indeed, how about changing it to `EmuInt3`?

src/include/os.h
ksco: yeah, that's why there is a newline from the above functions, but I should clarify the scope of the comment.

src/dynarec/arm64/dynarec_arm64_00.c
ptitSeb: yeah, sounds good to me

PR #2501, opened 2025-04-05
Corrected arguments of my_epoll_pwait2

kpyrkosz:
Box64 segfaults on this call. I updated it in accordance with the specification in the header file:
```
GOM(epoll_ctl, iFEiiip)
GOM(epoll_pwait, iFEipiip)
GOM(epoll_pwait2, iFEipipp)
GOM(epoll_wait, iFEipii)
```
With this change an example program ran successfully.
On a side note, why wasn't this typo detected by the script validating the wrappers?

### DIFF
--- a/src/wrapped/wrappedlibc.c
+++ b/src/wrapped/wrappedlibc.c
@@ -2201,7 +2201,7 @@ EXPORT int32_t my_epoll_pwait(x64emu_t* emu, int32_t epfd, void* events, int32_t
         UnalignEpollEvent(events, _events, ret);
     return ret;
 }
-EXPORT int my_epoll_pwait2(int epfd, void* events, int maxevents, struct timespec *timeout, sigset_t * sigmask)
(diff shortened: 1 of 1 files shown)


ptitSeb:
Not sure yes. I think the control has been disabled at soe point: @rajdakin can you check?

rajdakin:
That's because the script validating the wrappers only reads the `_private.h`s, not the `.c`s. Since the missing stuff was in the `.c`, no issue was reported.
(Also, in `box64` the `E` argument is not mandatory for `GOM`s, so no checks for that parameter exist.)

I could maybe hack something with the wrapperhelper to also check the signatures in the `.c`.
Another idea would be to add checks in the `.c`s to say "if `xxx` is a `GOM`, then `my_xxx` must have signature x". The macros would generate something like `_Static_assert(_Generic(&my_epoll_pwait2, iFEipipp_t*: 1, default: 0));` for each `GOM` in the `_private.h`. Maybe that's not possible and/or practical though...

ptitSeb:
Ah yeah, of course, no .c check... That would be nice, but is really just a "nice to have".

PR #2507, opened 2025-04-06
[CMAKE]: Fix arm64 build

cjy880927:
Log: Ensure successful compilation on devices that do not support
     cortex-a77.cortex-a55, such as those with CPU Model Cortex-A55.


CMakeLists.txt
```diff
@@ -230,8 +230,8 @@ elseif(ARM64)
     #add_definitions(-pipe -march=native)
 endif()
 if(ARM_DYNAREC)
-    #putting a random cpu that have 8.2 architecture so assembly can be build. the use of the feature are conditionnal anyway
-    set(ASMFLAGS  -pipe -mcpu=cortex-a76)
+    set(CFLAGS -pipe -march=armv8.2-a+crc+simd+crypto)
```
ptitSeb, line 233: No. This will conflict with what is setup for other profile.

CMakeLists.txt
```diff
@@ -230,8 +230,8 @@ elseif(ARM64)
...
+    set(ASMFLAGS  -pipe -march=armv8.2-a+crc+simd+crypto)
```
ptitSeb, line 234: this flags is for compiling asm file, but it doesn't necessarily need to be able to run. You have a compiler that doesn't support "cortex-a76" but support "armv8.2"?

ptitSeb (changes requested):

CMakeLists.txt
cjy880927: Moreover, upgrading GCC to version 9 or higher in my environment is either difficult or cumbersome. Therefore, the compilation is modified to use only the ARMv8.2-A base architecture.

CMakeLists.txt
ptitSeb: What about setting those flags only if GCC is older than 9.0?

cjy880927: > What about setting those flags only if GCC is older than 9.0?

How about changing to the following code?

ptitSeb (approved):
It's still not the correct place for the CFLAGS one

ptitSeb (changes requested):
Actually, no. CFLAGS must not be set here. That would broke non-arm8.2 build on older gcc. Please remove that line.

cjy880927:
> Actually, no. CFLAGS must not be set here. That would broke non-arm8.2 build on older gcc. Please remove that line.


Thank you for the reminder! The CFLAGS have been removed.

PR #2513, opened 2025-04-08
[WOW64] Add wow64 PE build scaffolding

ksco:
Also moved more functions to os.h.

### DIFF
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -2,16 +2,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <dlfcn.h>
-#include <signal.h>
 #include <pthread.h>
 #include <errno.h>
-#include <syscall.h>
-#include <sys/personality.h>
 
 #include "os.h"
 #include "box64context.h"
-#include "elfloader.h"
 #include "debug.h"
 #include "x64trace.h"
 #include "x64emu.h"
@@ -19,7 +14,6 @@
 #include "bridge.h"
 #include "library.h"
 #include "callback.h"
-#include "wrapper.h"
 #include "threads.h"
 #include "x64trace.h"
 #include "signals.h"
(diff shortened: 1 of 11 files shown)


src/custommem.c
```diff
@@ -2085,18 +2077,23 @@ void relockCustommemMutex(int locks)
 
 static void init_mutexes(void)
 {
-    #ifdef USE_CUSTOM_MUTEX
+#ifdef _WIN32 // TODO: better wow64 support?
+    InitializeCriticalSection(&mutex_blocks);
+    InitializeCriticalSection(&mutex_prot);
+#else
+#ifdef USE_CUSTOM_MUTEX
```
ptitSeb, line 2084: shouldn't USE_CUSTOM_MUTEX be the top test and then WIN32/LINUX in the fallback?

src/os/os_linux.c
```diff
@@ -62,3 +63,45 @@ void EmuX86Syscall(void* emu)
 {
     x86Syscall((x64emu_t*)emu);
 }
+
+void PersonalityAddrLimit32Bit(void)
+{
+    personality(ADDR_LIMIT_32BIT);
+}
+
+void ShowNativeBackTrace(int loglevel_min)
```
ptitSeb, line 72: can we don't have 2 functions calls for showBT (as they will appears on the backtrace)

ptitSeb:
There is a separate CMakeList.txt? can't it be merged with the main one?

ptitSeb:
> > There is a separate CMakeList.txt? can't it be merged with the main one?
> 
> Yeah, eventually it would make sense to have it merged, or at least shared. For now I just try to keep this simple so I don’t have to bother the build system.

Ah ok, fair enough.

src/os/os_linux.c
ksco: thanks for pointing this out, I converted it to a macro.

ksco:
I don't understand why static build is still failed of build.

PR #2532, opened 2025-04-14
[WOW64] Added non-functional PE build

ksco:
More moving around and placeholders, and now it builds. That's some progress I guess.

```
[100%] Linking C shared library libwowbox64.dll
[100%] Built target wowbox64
```


ptitSeb:
So many things moved around :'(

PR #2550, opened 2025-04-19
Merge `mmapmem` into `mapallmem`

### LINKED ISSUE #2546: Proposal to Merge `mmapmem` into `mapallmem`: Need Clarification on `data == 2`
I'm planning to merge `mmapmem` into `mapallmem`  (see [commit 93ebb0c](https://github.com/devarajabc/box64/commit/93ebb0c58a434bb6c0f39503b88a8b9ee4047139)), by using the `data` field to indicate whether a memory fragment is allocated via `mmap`. Since all memory mapped in `mmapmem` is also mapped in `mapallmem`, maintaining both trees may cause memory redundancy.

In the proposed scheme:

- `data == 1`: Memory fragment is allocated.
- `data == 2`: Memory fragment is allocated by `mmap`.

However, I noticed that the `data` field is already being assigned the value `2` in the `reverveHigMem32()` function
```c
//src/custommem.c
void reverveHigMem32(void) {
    // -- skipped --
    rb_set(mapallmem, (uintptr_t)cur, (uintptr_t)cur + cur_size, 2);
    // -- skipped --
}
```
I couldn't find any documentation or code that clearly distinguishes this usage of `data == 2` in `mapallmem.`

So my question is: what does `data == 2` represent in this context?
Alternatively, would it be acceptable if I use `data == 3` in `mapallmem` to indicate memory fragments allocated by `mmap`?

To test this approach, I ran  `chess.exe` with `box64` ([commit 93ebb0c](https://github.com/devarajabc/box64/commit/93ebb0c58a434bb6c0f39503b88a8b9ee4047139)) and `wine`  (same config as in [issue #2511](https://github.com/ptitSeb/box64/issues/2511)). 
Below is the result after merging `mmapmem` into `mapallmem`:

Image
Compare to Box64 (v0.3.5 73dfe869):
Image

As shown above, the total number of red-black tree nodes was reduced by approximately 81.
Since each node occupies 56 bytes, this translates to a memory saving of roughly 4.4 KB, which can become more significant when running larger programs with extensive memory mappings.

devarajabc:
Remove the separate `mmapmem` red-black tree and merges its into `mapallmem`, reducing overall memory usage.

To distinguish memory regions that were previously tracked separately in `mmapmem`,
we now use a `mem_flag_t` bitmask enum in the data field of `mapallmem`:
```
    MEM_ALLOCATED = 1   // allocated
    MEM_RESERVED  = 2   // reserved 
    MEM_MMAP      = 3   // allocated via mmap()
```

In tests with `chess.exe` (Box64 + Wine), this reduced the total node count by ~81, saving ~4.4 KB of memory. 
The impact is expected to be greater in mmap-heavy programs.

Resolves #2546

**EDIT:** The total node count was reduced by approximately 189 nodes (the entire `mmapmem` tree).
It's not 81 as I previously mentioned.

### DIFF
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -59,9 +59,13 @@ pthread_mutex_t     mutex_blocks;
 rbtree_t* memprot = NULL;
 int have48bits = 0;
 static int inited = 0;
-
+typedef enum {
+    MEM_UNUSED = 0,
+    MEM_ALLOCATED = 1,
+    MEM_RESERVED = 2,
+    MEM_MMAP = 3
+} mem_flag_t;
 rbtree_t*  mapallmem = NULL;
-static rbtree_t*  mmapmem = NULL;
 static rbtree_t*  blockstree = NULL;
 
 #define BTYPE_MAP   1
@@ -980,10 +984,10 @@ void* box32_dynarec_mmap(size_t size)
     uintptr_t bend = 0;
     while(bend<0x800000000000LL) {
         if(rb_get_end(mapallmem, cur, &flag, &bend)) {
-            if(flag==2 && bend-cur>=size) {
+            if(flag == MEM_RESERVED && bend-cur>=size) {
                 void* ret = InternalMmap((void*)cur, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_FIXED | MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
                 if(ret!=MAP_FAILED)
-                    rb_set(mapallmem, cur, cur+size, 1);    // mark as allocated
+                    rb_set(mapallmem, cur, cur+size, MEM_ALLOCATED);    // mark as allocated
                 else
                     printf_log(LOG_INFO, "BOX32: Error allocating Dynarec memory: %s\n", strerror(errno));
                 cur = cur+size;
@@ -1757,7 +1761,7 @@ void updateProtection(uintptr_t addr, size_t size, uint32_t prot)
     LOCK_PROT();
     uintptr_t cur = addr & ~(box64_pagesize-1);
     uintptr_t end = ALIGN(cur+size);
-    rb_set(mapallmem, cur, cur+size, 1);
+    rb_set(mapallmem, cur, cur+size, MEM_ALLOCATED);
     while (cur < end) {
         uintptr_t bend;
         uint32_t oprot;
@@ -1789,7 +1793,7 @@ void setProtection(uintptr_t addr, size_t size, uint32_t prot)
     ++setting_prot;
     uintptr_t cur = addr & ~(box64_pagesize-1);
     uintptr_t end = ALIGN(cur+size);
-    rb_set(mapallmem, cur, end, 1);
+    rb_set(mapallmem, cur, end, MEM_ALLOCATED);
     rb_set(memprot, cur, end, prot);
     --setting_prot;
     UNLOCK_PROT();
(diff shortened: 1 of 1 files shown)


ptitSeb:
are you sure the merging of this is a good thing? because the more fragmented the rbtree is, the slower access are. 4.4KB is really a low value. Also, side note, but those allocation (rbtree node) are all <128bytes and will be done by box64 custom allocator, and specificaly the 128bytes map allocator...

devarajabc:
I conducted a benchmark using [AnsiBench](https://github.com/nfinit/ansibench), comparing Box64 with and without `mmapmem`. The result shows no measurable slowdown:

Iterations_sec  (NORMALIZED)

I'm still trying to understand where fragmentation might increase by removing `mmapmem`.
From what I see, `mmapmem` is a subset of `mapallmem` — every region inserted into `mmapmem` is also inserted into `mapallmem` (`setProtection_mmap()`).

From that perspective, merging `mmapmem` into `mapallmem` just tags existing entries with extra metadata.
Unless I’m missing something, this shouldn't increase fragmentation — but please let me know if there's a scenario I'm not seeing.

Experimental Environment:
Box64 Version: v0.3.5(Commit 549b57e)
Host System: Raspberry Pi 400 running Linux Debian 6.12.3-arm64 #1 SMP Debian 6.12.3-1 (2024-12-07) aarch64 GNU/Linux
Build Configuration:
```
cmake \
    -G Ninja \
    -D RPI4ARM64=1 -D ARM_DYNAREC=ON \
    -D CMAKE_BUILD_TYPE=RelWithDebInfo \
    -D CMAKE_C_COMPILER=aarch64-none-linux-gnu-gcc \
    ..
ninja
```

devarajabc:
You're right — mmapmem likely serves as a cache or fast-access layer for mmap-specific lookups.
Removing it might impact lookup efficiency depending on access patterns and how frequently mmap-related queries are made.

I ran a benchmark using [AnsiBench](https://github.com/nfinit/ansibench), comparing Box64 with and without mmapmem.
The results show no measurable slowdown:
Iterations_sec  (NORMALIZED)

That said, I plan to gather more `statistics` on the runtime usage patterns of `mmapmem` — specifically how often it's queried and whether it significantly reduces tree traversal overhead in practice.

Experimental Environment:
Box64 Version: v0.3.5(build on Commit 549b57e)
Host System: Raspberry Pi 400 running Linux Debian 6.12.3-arm64 #1 SMP Debian 6.12.3-1 (2024-12-07) aarch64 GNU/Linux
Build Configuration:
```
cmake \
    -G Ninja \
    -D RPI4ARM64=1 -D ARM_DYNAREC=ON \
    -D CMAKE_BUILD_TYPE=RelWithDebInfo \
    -D CMAKE_C_COMPILER=aarch64-none-linux-gnu-gcc \
    ..
ninja
```

src/custommem.c
```diff
@@ -1801,15 +1805,22 @@ void setProtection_mmap(uintptr_t addr, size_t size, uint32_t prot)
...
-    LOCK_PROT();
-    rb_set(mmapmem, addr, addr+size, 1);
     if(!prot) {
-        rb_set(mapallmem, addr, addr+size, 1);
+        LOCK_PROT();
+        rb_set(mapallmem, addr, addr+size, MEM_MMAP);
         rb_unset(memprot, addr, addr+size);
+        UNLOCK_PROT();
+    }
+    else{//SetProtection
```
ptitSeb, line 1814: Why removing the call to setProtection to inline the code? the compiler will do that anyway, without the need to duplicate code.

devarajabc: Because `setProtection` assigns the `MAP_ALLOCATED` flag to the memory range, but in this case, we need to mark it with `MAP_MMAP` instead.

ptitSeb: ok.

Issue #2511, opened 2025-04-08
Clarification on the Usage and Purpose of the `dynamap` Red-Black Tree in Box64


devarajabc:
While examining Box64's memory management system, particularly the usage of `rbtree`, I observed unusual behavior involving the `dynamap` tree. By running Box64 combined with Wine64 on a Windows 64-bit chess game, I noticed that the `dynamap` tree significantly outnumbers other trees in node count. However, despite the high node usage, nodes from `dynamap` were rarely (if ever) actively accessed during execution (`rb_get`, `rb_get_64`, `rb_get_end`, `rb_get_end_64`, `rb_get_right` or `rb_get_left` calls).

![Image](https://github.com/user-attachments/assets/ca2db267-dd9d-48e3-8ced-8b0d6b1d6220)

Additionally, I collected detailed statistics on node creation and usage:
A total of 54,306 rbtree nodes were created during execution.
Only 17,886 nodes were released before the process terminated.

Among the released nodes, the breakdown of unused nodes (i.e., created but never accessed) is as follows:
1. `blockstree`: 0 unused nodes
2. `db_sizes`: 52 unused nodes
3. `dynamap`: 16,697 unused nodes (100% of released `dynamap` nodes were unused)
4. `envmap`: 1 unused node
5. `mapallmem`: 158 unused nodes
6. `mapmem`: 105 unused nodes
7. `memprot`: 873 unused nodes
8. `rbt_dynmem`: 0 unused nodes

To further verify this observation, I experimentally disabled the usage of dynamap within the Box64 codebase and repeated my tests. Surprisingly, the system functioned perfectly without noticeable issues.
```c
int rb_set_64(rbtree_t *tree, uintptr_t start, uintptr_t end, uint64_t data) {
    if(!strcmp(“dynamap”, tree->name))
        return 0;
    //--skip--//
}
```

![Image](https://github.com/user-attachments/assets/1f5b9b48-e7db-4275-b62e-ec2315112006)

Given these observations, I seek clarification on the following points:

1. **Intended Purpose**: What was the initial purpose behind implementing the `dynamap` tree within Box64's memory management system?

2. **Node Utilization**: Why does `dynamap` contain so many nodes if these nodes are rarely (or never) accessed during regular operations?

3. **Impact of Removal**: Considering that disabling `dynamap` had no observable negative impact in my tests, could removing or optimizing `dynamap` improve memory efficiency or overall performance?

Experimental Environment:
[...]

PR #2557, opened 2025-04-21
[RBTREE] Cache boundary nodes and remove `add_range()` 

devarajabc:
### 1. Cache leftmost and rightmost node 
By caching boundary nodes, we eliminate the need for repeated traversal.
Cached nodes are updated only during insertions or deletion, via simple assignments.

To test this approach, I ran chess.exe with box64 and wine (same config as in https://github.com/ptitSeb/box64/issues/2511).
The result: 397 full tree traversals were avoided, thanks to just 607 lightweight cache updates during approximately 3,500 tims insertions.
The performance impact is expected to be even greater in larger or more complex workloads.

### 2. Remove redundant `add_range()` wrapper
The function `add_range()` was only called when `tree->root == NULL`.  
In such cases, the internal while-loop never executes, and the call simply forwards to:

```c
add_range_next_to(tree, NULL, start, end, data);
```

### DIFF
--- a/src/tools/rbtree.c
+++ b/src/tools/rbtree.c
@@ -246,17 +246,6 @@ static int add_range_next_to(rbtree_t *tree, rbnode *prev, uintptr_t start, uint
     return -1; // unreachable
 }
 
-static int add_range(rbtree_t *tree, uintptr_t start, uintptr_t end, uint64_t data) {
-// printf("add_range\n");
-    rbnode *cur = tree->root, *prev = NULL;
-    while (cur) {
-        prev = cur;
-        if (cur->start < start) cur = cur->right;
-        else cur = cur->left;
-    }
-    return add_range_next_to(tree, prev, start, end, data);
-}
-
 static rbnode *find_addr(rbtree_t *tree, uintptr_t addr) {
     rbnode *node = tree->root;
     while (node) {
@@ -643,7 +632,7 @@ int rb_set_64(rbtree_t *tree, uintptr_t start, uintptr_t end, uint64_t data) {
 // printf("rb_set( "); rbtree_print(tree); printf(" , 0x%lx, 0x%lx, %hhu);\n", start, end, data); fflush(stdout);
 dynarec_log(LOG_DEBUG, "set %s: 0x%lx, 0x%lx, 0x%x\n", tree->name, start, end, data);
     if (!tree->root) {
-        return add_range(tree, start, end, data);
+        return add_range_next_to(tree, NULL, start, end, data);
     }
     
     rbnode *node = tree->root, *prev = NULL, *last = NULL;
@@ -840,7 +829,7 @@ uint64_t rb_inc(rbtree_t *tree, uintptr_t start, uintptr_t end) {
     // printf("rb_inc( "); rbtree_print(tree); printf(" , 0x%lx, 0x%lx);\n", start, end); fflush(stdout);
     dynarec_log(LOG_DEBUG, "inc %s: 0x%lx, 0x%lx\n", tree->name, start, end);
     if (!tree->root) {
-        add_range(tree, start, end, 1);
+        add_range_next_to(tree, NULL, start, end, 1);
         return 1;
     }
     
--- a/src/tools/rbtree.c
+++ b/src/tools/rbtree.c
@@ -31,6 +31,10 @@ struct rbtree {
     rbnode *root;
     const char* name;
     bool is_unstable;
+    // Cache 
+    rbnode *righter; 
+    rbnode *lefter;
+    // TODO: Refine the naming scheme
 };
 
 rbtree_t* rbtree_init(const char* name) {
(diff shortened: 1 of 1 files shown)


src/tools/rbtree.c
```diff
@@ -31,13 +31,18 @@ struct rbtree {
     rbnode *root;
     const char* name;
     bool is_unstable;
+    //Cache
```
ptitSeb, line 34: be aware that `sizeof(rbtree_t)` (and `sizeof(rbnode)` too) should not exceed 128 bytes, or bad things will happens!

devarajabc: `sizeof(struct rbtree)` is 40 bytes (on a 64-bit system with standard alignment).

ptitSeb: Yeah, I know it's ok for now. It's just something to keep in mind

src/tools/rbtree.c
```diff
@@ -1199,6 +1194,27 @@ static void print_rbnode(const rbnode *node, unsigned depth, uintptr_t minstart,
...
+        printf_log(LOG_NONE, "<invalid rightmost node> ");
+    rbnode *left_node = tree->root;
+    while (left_node) {
+        if(!left_node->left)
+            break;
+        left_node = left_node->left;
+    }
+    if(tree->lefter != left_node)
+        printf_log(LOG_NONE, "<invalid leftmost node> ");
+    return 1;
```
rajdakin, line 1216: This function does not return whether the leftmost and rightmost nodes are valid.

rajdakin (changes requested):
Other than some minor formatting issues and with the new function `cache_check` which should be static, I see no problem with this PR.

jserv:
>  Cache leftmost and rightmost node
> By caching boundary nodes, we eliminate the need for repeated traversal.
> Cached nodes are updated only during insertions or deletion, via simple assignments.

The idea of maintaining cached pointers to the leftmost and rightmost nodes in a red-black tree implementation appears to have been influenced by the Linux kernel's approach. According to the Linux kernel documentation, this optimization converts potentially expensive O(log N) operations for finding the smallest element (via `rb_first()`) into simple pointer dereferences, providing significant performance benefits for common tree operations like traversals. When implementing this optimization, it would be valuable to reference the Linux kernel's implementation as precedent and include experimental performance data in your commit messages to clearly demonstrate both the advantages and any potential memory overhead tradeoffs.

See [Red-black Trees (rbtree) in Linux](https://www.kernel.org/doc/html/latest/core-api/rbtree.html)

src/tools/rbtree.c
```diff
@@ -31,13 +31,18 @@ struct rbtree {
     rbnode *root;
     const char* name;
     bool is_unstable;
+    // Cache 
+    rbnode *righter; 
+    rbnode *lefter;
```
jserv, line 36: Refine the naming scheme to reflect leftmost and rightmost. Also, improve the comments.

src/tools/rbtree.c
```diff
@@ -1216,6 +1234,7 @@ void rbtree_print(const rbtree_t *tree) {
         return;
     }
     unsigned bdepth = 0;
+    // cache_check(tree);
```
jserv, line 1237: Don't leave the dead code.

src/tools/rbtree.c
```diff
@@ -1199,13 +1194,38 @@ static void print_rbnode(const rbnode *node, unsigned depth, uintptr_t minstart,
     }
     printf_log(LOG_NONE, ")");
 }
+static void cache_check(const rbtree_t *tree) {
```
jserv, line 1197: Add a blank line prior to this line. Always follow the consistent style.

src/tools/rbtree.c
jserv: Place TODO comments for enforcing the consistent naming.

jserv (changes requested):
Enhance the git commit messages for the red-black tree implementation by explaining the motivation behind each change, quantifying performance improvements or other benefits, and including references to relevant documentation or issues.

jserv (changes requested):
Replace "Ref: Linux kernel `include/linux/rbtree.h`" with the hyperlink to the official Linux kernel documentation instead of the file name.

src/tools/rbtree.c
```diff
@@ -1216,6 +1238,7 @@ void rbtree_print(const rbtree_t *tree) {
         return;
     }
     unsigned bdepth = 0;
+    cache_check(tree);
```
jserv, line 1241: Why did you place the call to `cache_check` here? Move it prior to `unsigned bdepth` declaration.

src/tools/rbtree.c
```diff
@@ -819,7 +819,7 @@ uint64_t rb_inc(rbtree_t *tree, uintptr_t start, uintptr_t end) {
     // printf("rb_inc( "); rbtree_print(tree); printf(" , 0x%lx, 0x%lx);\n", start, end); fflush(stdout);
     dynarec_log(LOG_DEBUG, "inc %s: 0x%lx, 0x%lx\n", tree->name, start, end);
     if (!tree->root) {
-        add_range(tree, start, end, 1);
+        add_range_next_to(tree, NULL ,start, end, 1);
```
jserv: Should be `add_range_next_to(tree, NULL, start, end, 1);`.
Be aware of the position of commas.

jserv (approved):

rajdakin (approved):
Looks good to me now.

PR #2610, opened 2025-05-07
[DYNAREC] Use PE volatile metadata in dynarec

ksco:
Only the dlls for now, main exe will come later...


ptitSeb:
When .exe are handled (should be trivial?), it would be good to evolve the `UnityPlayer.dll` to not setup strongmem=1 but that new volatile_metadata instead.

Issue #2579, opened 2025-04-28
[NON4KPAGE] Some more adjustments for non-4Kpagesize system regression


xiangzhai:
Hi,

[PltResolver64 is not able to store offs](https://github.com/ptitSeb/box64/blob/main/src/elfs/elfloader.c#L1911) to `R_X86_64_JUMP_SLOT` `0x3f00003fb0` because `0x3f00003fb0` is read only:

```
[BOX64] PltResolver64: Addr=0xff88005130, Slot=5 Return=0x3f000011a6(/yourpath/jdk/build/linux-x86_64-server-release/images/test/hotspot/jtreg/native/libTest35859.so/Java_Test35859_inc + 0x26): elf is /yourpath/jdk/build/linux-x86_64-server-release/images/test/hotspot/jtreg/native/libTest35859.so (VerSym=0x678, deepbind=0, local_maplib=0xff8800de30) func param: 0xba, 0xffc95ea018...
[BOX64]             Apply STB_GLOBAL R_X86_64_JUMP_SLOT 0x3f00003fb0 with sym=syscall(optver 2: syscall@GLIBC_2.2.5) (0x3f00001086 -> 0x30020740 / box64)
...
=> [BOX64] 3f00000000-3f00004000 r--p 00000000 00:00 0 
                                 ^--- read only
[BOX64] 3f00004000-3f00008000 rw-p 00000000 00:00 0
...
#8  0x000000ff8de852b0 in PltResolver64 (emu=0xff8800b470) at /yourpath/box64/src/elfs/elfloader.c:1912
#9  0x000000ff8de92cb0 in x64Int3 (emu=0xff8800b470, addr=<optimized out>) at /yourpath/box64/src/emu/x64int3.c:330
#10 0x000000ffc8d602fc in ?? ()
...
(gdb) frame 8
#8  0x000000ff8de852b0 in PltResolver64 (emu=0xff8800b470) at /yourpath/box64/src/elfs/elfloader.c:1912
1912	            *p = offs;
(gdb) x p
0x3f00003fb0:	0x00001086
(gdb) x/22i $pc-44
   0xff8de85284 <PltResolver64+1076>:	st.d	$r27,$r3,8(0x8)
   0xff8de85288 <PltResolver64+1080>:	st.d	$r16,$r3,0
   0xff8de8528c <PltResolver64+1084>:	move	$r10,$r25
   0xff8de85290 <PltResolver64+1088>:	move	$r8,$r11
[...]

PR #2611, opened 2025-05-08
[NON4KPAGE] Check PltResolver64 Elf64_Rela r_offset address is writable (#2579)

xiangzhai:
Hi,

`make test` Passed.
[jtreg](https://github.com/openjdk/jtreg) and [jna](https://github.com/java-native-access/jna) testcase Passed.
[btbench](https://github.com/xiangzhai/btbench) Passed.

Please review my patch.

Thanks,
Leslie Zhai

### DIFF
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -1940,6 +1940,29 @@ void loadProtectionFromMap()
 #endif
 }
 
+int isWritableFromMap(uintptr_t addr)
+{
+#ifndef _WIN32 // TODO: Should this be implemented on Win32?
+    char buf[500];
+    FILE* f = fopen("/proc/self/maps", "r");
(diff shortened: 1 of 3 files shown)


ptitSeb:
I don' t understand what is happening here.

First, why do you browse /proc/sef/maps for each relocation, that will be very very slow. BWhy not use internal getProtection() instead.

Second, are you skipping the relocation if it's write only? The relocation is not optionnal.

The root cause, I think, here, is that the page should not be read-only, but read-write, and is R/O only because the page are too large and an initialy R/W page got R/O after the next block is loaded. The page should be R/W, the elfloader should not write procted exiting page and combine the rights. 
I thought this was fixed in the past.

PR #2619, opened 2025-05-11
[WOW64] Initial implementation patches

AndreRH:
I first need to clarify some copyright topics for the rest.
Please check if my compiler.h approach is ok for now.

### DIFF
--- a/src/include/env.h
+++ b/src/include/env.h
@@ -21,8 +21,14 @@
     STRING(NAME, name)
 */
 
+#ifdef _WIN32
+#define DEFAULT_LOG_LEVEL (LOG_INFO)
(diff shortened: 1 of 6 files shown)


ptitSeb:
That looks good to me. I don't see any issue with the compiler.h. It's in wow64 subdir, so fine with me.

ptitSeb (approved):

ksco (approved):

AndreRH:
Would it be okay for you to not squash my patches in future PRs? I'm doing atomic patches and that effort is lost when squashed

ptitSeb:
> Would it be okay for you to not squash my patches in future PRs? I'm doing atomic patches and that effort is lost when squashed

I'll try, but I tend to forget sometimes.

PR #2620, opened 2025-05-11
Fix some steam crashing issues

woutermans:
This seems to fix some steam issues for me, using box64 and box32 on a Snapdragon X1 Elite device.

### DIFF
--- a/install_steam.sh
+++ b/install_steam.sh
@@ -22,6 +22,12 @@ echo '#!/bin/bash
 export STEAMOS=1
 export STEAM_RUNTIME=1
 export DBUS_FATAL_WARNINGS=0
+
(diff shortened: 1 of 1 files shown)


woutermans:
I'm sorry for creating this PR so quickly without a proper implementation. Would a check for whether the system is using the freeadreno driver (for example checking if the ICD exists) be okay in your opinion?

For me the random crashes still occur, but normally the client would more often just hang, which seems not to happen with these ENV vars.

ptitSeb:
Box32 is still a bit unstable yes. Steam still crashes/freeze from time to time. But I think removing the libvulkan_gfxstream.so does the trick on my side.

libfreedreno is always installed as soon as you install mesa on most distro, testing it's presence is not discriminant enough.

PR #2624, opened 2025-05-12
wow64 part2 (dont squash)

AndreRH:


wow64/wowbox64.c
```diff
@@ -121,7 +176,22 @@ void WINAPI BTCpuSimulate(void)
...
 {
-    // NYI
+    WOW64_CONTEXT *ctx;
+    x64emu_t *emu = RtlAllocateHeap( GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(*emu) );
+    __wine_dbg_output("[BOX64] BTCpuThreadInit\n");
+
+    RtlWow64GetCurrentCpuArea( NULL, (void **)&ctx, NULL );
+    emu->context = &box64_context;
+
+    // setup cpu helpers
```
ptitSeb, line 186: is this supposed to init the whole `x64emu_t` structure? why not using the regular init function for that?

ptitSeb: We can discuss that later anyway, but I think using only 1 function to initialize a x64emu_t structure is better.

PR #2631, opened 2025-05-13
[DOCS] improve consistency between documentation and code

sehraf:
The generated usage documentation is not guaranteed to match the code, therefore I hacked a little python script to perform some consistency checks. The intention is to have an easy way the check if both files (`usage.json` and `env.h`) are (mostly) in line. 

The script checks the following:
- same default value
- only one default value in `usage.json`
- same min/max value
- that an option is included in both files

Limitations:
- All options that take arbitrary values are not checked for min/max matching. (documented using `XXXX` in `usage.json`)
  - you can set `INCLUDE_BLANK = True` to ignore this
- `STRING` and `ADDRESS` entries are just checked for existing documentation.
- The script is very simple :upside_down_face: 

---

I've fixed the only usage of `XXX` (everywhere else `XXXX` is used), though some issues remain:
```
BOX64_DUMP                    : max value mismatch: env.h=2, usage.json=1
BOX64_AVX                     : missing usage.json entry
BOX64_DYNAREC                 : missing usage.json entry
BOX64_FUTEX_WAITV             : missing usage.json entry
BOX64_MMAP32                  : missing usage.json entry
BOX64_LD_PRELOAD              : missing env.h entry
```


ptitSeb:
Interesting, thanks!

sehraf:
> This is nice, but I suggest to move the code into gen.py and run the check every time we regenerate the documents.

can do

PR #2633, opened 2025-05-14
[DOCS] Align usage.json and env.h

ksco:
Thanks @sehraf for the checker (#2631), I added and improved most of the missing things. Now the checker outputs:
```
BOX64_LD_PRELOAD              : missing env.h entry
```

... which is fine as this one is special and not managed by the env module.

Important changes:

1. `BOX64_MMAP32` is now enabled by default.
2. `BOX64_FUTEX_WAITV` is removed, only configurable at compilation time.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -171,6 +171,14 @@ Add --no-sandbox argument to the guest program.
 
 ## Compatibility
 
+### BOX64_AVX
+
+Expose AVX extension to CPUID and cpuinfo file.
+
+ * 0: Do not expose AVX capabilities. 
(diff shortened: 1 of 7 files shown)


docs/USAGE.md
```diff
@@ -171,6 +171,14 @@ Add --no-sandbox argument to the guest program.
 
 ## Compatibility
 
+### BOX64_AVX
+
+Expose AVX extension to CPUID and cpuinfo file.
```
ptitSeb, line 176: because it's dafault value depends on the architecture, maybe it should be noted in the description that it's eanbled at 2 for arm build and disable by default on other architecture?

ksco: Indeed, I forgot that.

PR #2647, opened 2025-05-19
wow64 part3 (dont squash)

AndreRH:


src/emu/x64emu_private.h
```diff
@@ -124,6 +124,7 @@ typedef struct x64emu_s {
     uintptr_t   old_savedsp;
     #endif
 
+    uint64_t    win64_teb;
```
ptitSeb, line 127: can we `#ifdef _WIN32` this one?

AndreRH: good point

PR #2661, opened 2025-05-23
Added -k option to kill all box64 instances

ksco:


src/core.c
```diff
@@ -578,6 +580,46 @@ void PrintHelp() {
...
+void KillAllInstances()
+{
+
+    struct dirent* entry;
+    ssize_t len;
+    char proc_path[PATH_MAX];
+    char exe_path[PATH_MAX];
+    char exe_target[PATH_MAX];
+
+    DIR* proc_dir = opendir("/proc");
```
jserv, line 595: At the end of this function, you should release `/proc` directory stream. That is, `closedir(proc_dir)`.

ksco: Thanks, fixed. Also changed the process name from hard-coded `"box64"` to the name of self process.

ksco:
Please change to squash before merging this ;) @ptitSeb

PR #2667, opened 2025-05-25
wow64 part5(dont squash) 

AndreRH:


src/os/my_cpuid_wine.c
```diff
@@ -1,15 +1,548 @@
...
+}
 
 const char* getBoxCpuName()
 {
-    return NULL;
+    static char branding[] = "libwowbox64.dll";
+    return branding;
 }
 
 void my_cpuid(x64emu_t* emu, uint32_t tmp32u)
```
ptitSeb, line 33: why duplicating this code?

ptitSeb: It's exactly the same code as the linux version right?

AndreRH: true, would you rather have a my_cpuid_common.c for that function or only one file with ifdefs?

AndreRH: Which suggestion? : D

PR #2682, opened 2025-05-29
[WOW64] Added more missing pieces and the interpreter works

ksco:


ksco:
@AndreRH I ported some more missing part from hangover, please have a look if you find the time. :^)

AndreRH:
First of all: I won't always be starting a review as quickly as today :)

Second: I know it's very common here to have squashing and non-atomic commits, but I really don't understand why saving on the commit count would have any benefit over easy reviews, comprehensibility and easy git bisects. So having all those topics in one commit really is a mess to me: various formatting changes, syscall cleanup, signal handling, file moving, adding debug messages. I'd have liked to see at least 5 commits here.

Anyway, I'll leave some comments on the code

ksco (commented):
Thanks for the quick review! I've addressed the suggestions, please check again!

PR #2694, opened 2025-05-31
[RBTREE] Update comments on red-black tree usage

devarajabc:
I used `grep -rnw . -e 'rbtree_init('` to discover that there are seven red-black trees, and the information in `rbtree.h` is out of date. This PR:
- Remove outdated information
- Document the seven red-black trees that Box64 currently uses:
  1. memprot
  2. mapallmem
  3. blockstree
  4. db_sizes
  5. envmap
  6. rbt_dynmem
  7. volatileRanges
- Clarify each tree's role
- Notes the trade-offs of using the bitmap allocator (`map128_customMalloc`), which allocates 128-byte slabs and results in internal fragmentation

Note: I'm not sure I fully understand the rules and behavior of `envmap`, so feedback or corrections on that part would be appreciated.

### DIFF
--- a/src/include/rbtree.h
+++ b/src/include/rbtree.h
@@ -1,25 +1,58 @@
 /*
- * The primary function of red-black trees in Box64 is to provide a fast and efficient method (O(log(n))) for managing memory mappings. 
- * Each red-black tree node includes two additional pointers, "start" and "end," to denote a specific memory range. 
- * When a range of memory is mapped from the system, it is recorded in red-black trees based on the type and characteristics of that memory range.
+ * The primary function of red-black trees in Box64 is to provide an O(log n) method for managing memory mappings.
+ * This provides a mechanism that allows retrieving the memory range or associated data based on an address.
+ * Unlike standard red-black trees, each node here includes two additional fields, "start" and "end," to denote a specific memory range.
  *
- * Box64 utilizes four red-black trees, each serving distinct purposes:
- * 1. memprot: Manages memory protection settings where each node's data field represents the permissions of the memory range. 
- *    The "rb_set" function within this tree can set or change permissions for a specified range.
- * 2. mapallmem: Tracks all memory mappings. Its nodes' data field can indicate whether the memory is included in this mapping.
- * 3. mmapmem: Similar to "mapallmem" but specifically for memory mapped through the mmap system call. 
- *    The nodes' data fields here differentiate whether the memory is exclusively mapped by mmap.
- * 4. blockstree: Specifically contains memory ranges for blocks, where each node represents a different block. 
- *    The "data" field in this tree represents the index of the block in the "blocklist_s" array "p_blocks".
- *
- * Currently, there is an overlap between "mapallmem" and "mmapmem", where if memory is mapped in "mmapmem", it is also considered mapped in "mapallmem". 
- * If "mapallmem" and "mmapmem" were to be merged in the future, the "data" field could then take values indicating:
- * - 0: Memory is not mapped.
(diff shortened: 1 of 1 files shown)


devarajabc:
@ksco  The original intention was to help future contributors quickly understand the purpose of each rbtree, especially since their roles aren't always obvious from the variable names alone.

This context may also be useful for anyone looking to propose a faster lookup mechanism to replace rbtrees in the future, as understanding the current usage patterns is a necessary first step.

That said, I agree that it might not be ideal to place this in rbtree.h—I'm also not entirely sure where the best location would be.

devarajabc:
@ptitSeb  You're right that O(log N) is hard to beat in general.  
But depending on the data distribution and access patterns , there might be room for optimization.
Different structures may offer better constants or cache locality in some cases. Of course, that would require profiling and benchmarks to justify.

devarajabc:
@ptitSeb  I'm a bit concerned that it could clutter the code, but since it's more of a formatting issue, I'm happy to follow your preferred approach.

ptitSeb:
> @ptitSeb I'm a bit concerned that it could clutter the code, but since it's more of a formatting issue, I'm happy to follow your preferred approach.

I leave it up to you. But keeping it in rbtree.h is guarantied obsolescence next time an rbtree is removed / changed. That can still be acceptable anyway (not much changes planned here for now).

src/include/rbtree.h
```diff
@@ -1,25 +1,58 @@
 /*
- * The primary function of red-black trees in Box64 is to provide a fast and efficient method (O(log(n))) for managing memory mappings. 
- * Each red-black tree node includes two additional pointers, "start" and "end," to denote a specific memory range. 
- * When a range of memory is mapped from the system, it is recorded in red-black trees based on the type and characteristics of that memory range.
+ * The primary function of red-black trees in Box64 is to provide an O(log(n)) method for managing memory mappings.
```
jserv: Change `O(log(n))` to `O(log n)` as the standard mathematical notation for Big O complexity does not use parentheses around the variable.

src/include/rbtree.h
```diff
@@ -1,25 +1,58 @@
...
+ * This provides a mechanism that allows retrieving the memory range or associated data based on an address.
+ * Each red-black tree node includes two additional fields, "start" and "end," to denote a specific memory range.
```
jserv: Consider the following:
> Unlike standard red-black trees, each node here includes two additional fields, "start" and "end," to denote a specific memory range.

devarajabc:
@rajdakin  Agreed — `db_sizes` isn’t range-based, yet it relies on rbnode, which introduces unnecessary memory overhead. As seen in issue #2511 .  
I’ll prototype a replacement and benchmark its impact.

jserv (approved):

ptitSeb:
> @rajdakin Agreed — `db_sizes` isn’t range-based, yet it relies on rbnode, which introduces unnecessary memory overhead. As seen in issue #2511 . I’ll prototype a replacement and benchmark its impact.

db_sizes still use many of the rb_tree properties. especialy, when the current max is removed, the "righter" is looked for, which is still `O(log n)` on rbtree, while it would be `O(n)` on a hash table. Also, db_size will be changed later when dynarec memory is splited per mmap file (but will still be an rb_tree).

The objective of db_sizes is to have the maximum size of a dynablock always up-to-date, even when dynablock are removed. the maximum dynablock size is use when invalidating dynablock because of memory write, and it should be as small as possilble, so db_size keep this value accurate.

PR #2723, opened 2025-06-10
[WOW64] Added a minimal Windows runtime

ksco:


ptitSeb:
So this is to avoid using Kernel32.dll?

ksco:
> So this is to avoid using Kernel32.dll?

Exactly, now it has no dependency on any DLLs other than wow64.dll and ntdll.dll

PR #2730, opened 2025-06-12
[ARM64_DYNAREC] Added 64/65 67 89/8B 64bits opcodes

ksco:


src/dynarec/arm64/dynarec_arm64_6764.c
```diff
@@ -0,0 +1,96 @@
...
+        case 0x89:
+            INST_NAME("MOV Seg:Ed, Gd");
+            nextop = F8;
+            GETGD;
+            if (MODREG) { // reg <= reg
+                MOVxw_REG(TO_NAT((nextop & 7) + (rex.b << 3)), gd);
+            } else {      // mem <= reg
+                grab_segdata(dyn, addr, ninst, x4, seg, (MODREG));
+                addr = geted(dyn, addr, ninst, nextop, &ed, x2, &fixedaddress, NULL, 0, 0, rex, NULL, 0, 0);
+                ADDxw_REG(x4, x4, ed);
```
ptitSeb: I don't think this is correct. x4 should be a 64bits address, on "w" operation, the high 32bits will be losts, wich is not correct.

src/dynarec/arm64/dynarec_arm64_6764.c
```diff
@@ -0,0 +1,96 @@
...
+        case 0x8B:
+            INST_NAME("MOV Gd, Seg:Ed");
+            nextop = F8;
+            GETGD;
+            if (MODREG) { // reg => reg
+                MOVxw_REG(gd, TO_NAT((nextop & 7) + (rex.b << 3)));
+            } else {      // mem => reg
+                grab_segdata(dyn, addr, ninst, x4, seg, (MODREG));
+                addr = geted(dyn, addr, ninst, nextop, &ed, x2, &fixedaddress, NULL, 0, 0, rex, NULL, 0, 0);
+                ADDxw_REG(x4, x4, ed);
```
ptitSeb: same here

src/dynarec/arm64/dynarec_arm64_6764.c
ksco: `STRxw_REG`?

ptitSeb: Yes. like on other 64 prefix opcodes. `STRxw_REG(gd, x4, ed)`

ptitSeb:
Just to note: `STRxw_REGz` is not needed here because we know it's 64bits only. Else, "z" would be important to make sure the 32bits offset (that could be negative, and is mostly is) is 1st expanded to 64bits with "SXTW" in the memory operation.

ksco:
> Just to note: `STRxw_REGz` is not needed here because we know it's 64bits only. Else, "z" would be important to make sure the 32bits offset (that could be negative, and is mostly is) is 1st expanded to 64bits with "SXTW" in the memory operation.

yeah yeah I'm aware of that, as the 32-bit version has been split out

PR #2737, opened 2025-06-12
[WRAPPER] Wrapped a few more functions in gio2 and libresolv

ksco:


ptitSeb:
Need a version of the wrapper for static lib too. A `#if` needs an `#else`....

PR #2738, opened 2025-06-12
Bundle fix permissions

LukeShortCloud:


LukeShortCloud:
This started off as a single bug fix but then I found a few more issues.

PR #2745, opened 2025-06-14
[LA64_DYNAREC]Add basic avx support for la64.

phorcys:
basic infra for la64 avx
     some basic ops for avx 
       VMOVDQU/VMOVDQA/VMOVUPS/VMOVAPS/VMOVUPD/VMOVAPD 
       VZEROUPPER/VZEROALL 
       VMOVD/VMOVSD/VMOVSS 
       VINSERTF128/VINSERTI128/VEXTRACTF128/VEXTRACTI128 
       VBROADCASTSS/VBROADCASTSD/VBROADCASTF128

This pullrequest add basic AVX support for la64.

LoongArch LASX is 256bits SIMD, this patch use native 256-width reg/inst to imp avx insts.
Box64 interpreter has split YMM storage(x64emu_t.ymm[] only keep upper 128bits of YMM). 
In this patch, 256bits AVX reg  push/pop can't used xvld/xvst, 
but use  vld,vld+xvpermi.q  or vst,xvpermi.q+vst instead.

unlike arm64's ymm zero trace method, this patch use avxcache_t to store avx reg width and upper 128bits zero-fill states.

reg conv:
| conv  | additional transform |
| ------------- | ------------- |
|SSE->AVX | clear ssecache|
|AVX->SSE|  purge upper 128bits, clear avxcache|
|AVX128 -> AVX256 | land upper 128bits zero-fill if needed|
|AVX256 -> AVX128 | no special transform.|

Relay on x64 compiler's auto vzeroup/vzeroall generate. No AVX-SSE Transition Penalties imp yet.


ksco:
:+1:  Thank you for working on this. At first glance, it looks promising. But it's still a big PR, so I need to find a couple of hours to review, which will be early next week.

ksco:
I should've started the AVX support on RV64 too...

phorcys:
feel free for review this .
the state sync in this patch is a little complicated,I'm not sure that I wrote everything right.
especially when BOX64_DYNAREC_TEST=1.

ptitSeb:
That's an impressive PR! Thank you for your involvement in box64 developpement.

src/dynarec/la64/dynarec_la64_avx.c
```diff
@@ -0,0 +1,82 @@
...
+    // else if( (vex.m==VEX_M_0F38) && (vex.p==VEX_P_F3))
+    //     addr = dynarec64_AVX_F3_0F38(dyn, addr, ip, ninst, vex, ok, need_epilog);
+    else {
+        DEFAULT;
+    }
+
+
+    if ((*ok == -1) && (BOX64ENV(dynarec_log) >= LOG_INFO || BOX64DRENV(dynarec_dump) || BOX64ENV(dynarec_missing) == 1)) {
+        dynarec_log(LOG_NONE, "Dynarec unimplemented AVX opcode size %d prefix %s map %s opcode %02X ", 128 << vex.l, avx_prefix_string(vex.p), avx_map_string(vex.m), opcode);
+    }
```
ksco: This needs to be updated, check the latest arm64 code.

src/dynarec/la64/dynarec_la64_avx_0f.c
```diff
@@ -0,0 +1,186 @@
...
+    MAYUSE(cacheupd);
+#if STEP > 1
+    static const int8_t mask_shift8[] = { -7, -6, -5, -4, -3, -2, -1, 0 };
+#endif
+
+    rex_t rex = vex.rex;
+
+    switch (opcode) {
+        case 0x10:
+            INST_NAME("VMOVUPS Gx,Ex");
```
ksco: Please add a space after `,` for all the instruction names, i.e. `VMOVUPS Gx, Ex`.

src/dynarec/la64/dynarec_la64_avx_0f.c
```diff
@@ -0,0 +1,186 @@
...
+                INST_NAME("VZEROUPPER");
+                if (vex.v != 0) {
+                    UDF();
+                } else {
+                    q2 = fpu_get_scratch(dyn);
+                    XVXOR_V(q2, q2, q2);
+                    for (int i = 0; i < (rex.is32bits ? 8 : 16); ++i) {
+                        q1 = avx_get_reg(dyn, ninst, x1, i, 1, LSX_AVX_WIDTH_128);
+
+                        if (dyn->lsx.avxcache[i].v != -1 && dyn->lsx.avxcache[i].width == LSX_AVX_WIDTH_256) {
```
ksco: I don't understand this -- how is the width possible to be 256-bit when q1 is already 128-bit in line 160?

src/dynarec/la64/dynarec_la64_avx_0f.c
```diff
@@ -0,0 +1,186 @@
...
+                    }
+                }
+            } else {
+                INST_NAME("VZEROALL");
+                if (vex.v != 0) {
+                    UDF();
+                } else {
+                    for (int i = 0; i < (rex.is32bits ? 8 : 16); ++i) {
+                        q0 = avx_get_reg_empty(dyn, ninst, x1, i, LSX_AVX_WIDTH_256);
+                        XVXOR_V(q0, q0, q0);
```
ksco, line 178: Don't we need to set `zero_upper = 0` here?

src/dynarec/la64/dynarec_la64_avx_66_0f.c
```diff
@@ -0,0 +1,231 @@
...
+            if (MODREG) {
+                GETEYxy(q1, 0, 0);
+                if (vex.l) {
+                    XVOR_V(q0, q1, q1);
+                } else {
+                    VOR_V(q0, q1, q1);
+                }
+            } else {
+                SMREAD();
+                addr = geted(dyn, addr, ninst, nextop, &ed, x2, x1, &fixedaddress, rex, NULL, 0, 0);
```
ksco: better to be `geted(dyn, ..., 1, 0);`?

src/dynarec/la64/dynarec_la64_avx_66_0f.c
```diff
@@ -0,0 +1,231 @@
...
+            nextop = F8;
+            GETGYx_empty(q0);
+            GETED(0);
+            VXOR_V(q0, q0, q0);
+            if (rex.w) {
+                VINSGR2VR_D(q0, ed, 0);
+            } else {
+                VINSGR2VR_W(q0, ed, 0);
+            }
+            YMM0(gd);
```
ksco: `YMM0` here seems unnecessary.

src/dynarec/la64/dynarec_la64_avx_66_0f.c
```diff
@@ -0,0 +1,231 @@
...
+            if (MODREG) {
+                GETEYxy(q1, 0, 0);
+                if (vex.l) {
+                    XVOR_V(q0, q1, q1);
+                } else {
+                    VOR_V(q0, q1, q1);
+                }
+            } else {
+                SMREAD();
+                addr = geted(dyn, addr, ninst, nextop, &ed, x2, x1, &fixedaddress, rex, NULL, 0, 0);
```
ksco: better to be `geted(dyn, ..., 1, 0);`?

src/dynarec/la64/dynarec_la64_avx_0f.c
```diff
@@ -0,0 +1,186 @@
...
+    MAYUSE(gb2);
+    MAYUSE(q0);
+    MAYUSE(q1);
+    MAYUSE(d0);
+    MAYUSE(d1);
+    MAYUSE(s0);
+    MAYUSE(j64);
+    MAYUSE(cacheupd);
+#if STEP > 1
+    static const int8_t mask_shift8[] = { -7, -6, -5, -4, -3, -2, -1, 0 };
```
ksco: Unused.

src/dynarec/la64/dynarec_la64_avx_66_0f38.c
```diff
@@ -0,0 +1,98 @@
...
+    MAYUSE(gb2);
+    MAYUSE(q0);
+    MAYUSE(q1);
+    MAYUSE(d0);
+    MAYUSE(d1);
+    MAYUSE(s0);
+    MAYUSE(j64);
+    MAYUSE(cacheupd);
+#if STEP > 1
+    static const int8_t round_round[] = { 0, 2, 1, 3 };
```
ksco: Unused.

src/dynarec/la64/dynarec_la64_avx_66_0f3a.c
```diff
@@ -0,0 +1,105 @@
...
+    MAYUSE(gb2);
+    MAYUSE(q0);
+    MAYUSE(q1);
+    MAYUSE(d0);
+    MAYUSE(d1);
+    MAYUSE(s0);
+    MAYUSE(j64);
+    MAYUSE(cacheupd);
+#if STEP > 1
+    static const int8_t round_round[] = { 0, 2, 1, 3 };
```
ksco: Unused.

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
```diff
@@ -0,0 +1,106 @@
...
+            INST_NAME("VMOVSD Gx, [Vx,] Ex");
+            nextop = F8;
+            GETGYx_empty(q0);
+            GETEYSD(q2, 0, 0);
+            if (MODREG) {
+                GETVYx(q1, 0);
+                if (gd == vex.v) {
+                    VEXTRINS_D(q0, q2, 0);
+                } else {
+                    GETVYx(q1, 0);
```
ksco: redundant GETVYx.

ksco: Can be:
```
if (gd != vex.v) VOR_V(q0, q1, q1);
VEXTRINS_D(q0, q2, 0);
```

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
```diff
@@ -0,0 +1,106 @@
...
+                    VOR_V(q0, q1, q1);
+                    VEXTRINS_D(q0, q2, 0);
+                }
+            } else {
+                VXOR_V(q0, q0, q0);
```
ksco: Use `XVXOR_V` and `VEXTRINS_D` so we can remove the `YMM0` in `else` path?

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
```diff
@@ -0,0 +1,106 @@
...
+                if (ed == vex.v) {
+                    VEXTRINS_D(q0, q2, 0);
+                } else {
+                    GETVYx(q1, 0);
+                    VOR_V(q0, q1, q1);
+                    VEXTRINS_D(q0, q2, 0);
+                }
+                YMM0(gd);
+            } else {
+                addr = geted(dyn, addr, ninst, nextop, &ed, x2, x1, &fixedaddress, rex, NULL, 0, 0);
```
ksco: better to be `geted(dyn, ..., 1, 0);`?

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
```diff
@@ -0,0 +1,106 @@
...
+            if (MODREG) {
+                GETEYSD(q0, 0, 0);
+                if (ed == vex.v) {
+                    VEXTRINS_D(q0, q2, 0);
+                } else {
+                    GETVYx(q1, 0);
+                    VOR_V(q0, q1, q1);
+                    VEXTRINS_D(q0, q2, 0);
+                }
+                YMM0(gd);
```
ksco: It's not gd need to be clear up.

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
```diff
@@ -0,0 +1,106 @@
...
+                    VEXTRINS_D(q0, q2, 0);
+                }
+            } else {
+                VXOR_V(q0, q0, q0);
+                VEXTRINS_D(q0, q2, 0);
+            }
+            YMM0(gd);
+            break;
+        case 0x11:
+            INST_NAME("VMOVSD Ex, [Vx,] Gx");
```
ksco, line 79: Vx is not optional.

src/dynarec/la64/dynarec_la64_avx_f3_0f.c
```diff
@@ -0,0 +1,136 @@
...
+            if (MODREG) {
+                GETEYSS(q0, 0, 0);
+                if (ed == vex.v) {
+                    VEXTRINS_W(q0, q2, 0);
+                } else {
+                    GETVYx(q1, 0);
+                    VOR_V(q0, q1, q1);
+                    VEXTRINS_W(q0, q2, 0);
+                }
+                YMM0(gd);
```
ksco: It's not gd need to be clear up.

src/dynarec/la64/dynarec_la64_avx_f3_0f.c
```diff
@@ -0,0 +1,136 @@
...
+                    VEXTRINS_W(q0, q2, 0);
+                }
+            } else {
+                VXOR_V(q0, q0, q0);
+                VEXTRINS_W(q0, q2, 0);
+            }
+            YMM0(gd);
+            break;
+        case 0x11:
+            INST_NAME("VMOVSS Ex, [Vx,] Gx");
```
ksco, line 65: Vx is not optional.

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -883,6 +883,11 @@ int sse_get_reg(dynarec_la64_t* dyn, int ninst, int s1, int a, int forwrite)
         }
         return dyn->lsx.ssecache[a].reg;
     }
+    // upgrade from sse to avx
```
ksco: Does not seems to be an "upgrade" here

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -976,47 +992,232 @@ static void sse_reflectcache(dynarec_la64_t* dyn, int ninst, int s1)
...
+
+void avx_purge_reg_upper128(dynarec_la64_t* dyn, int ninst, int a, int forwrite)
+{
+    if (dyn->lsx.avxcache[a].v == -1 || forwrite == 0)
+        return;
+    if (dyn->lsx.lsxcache[dyn->lsx.avxcache[a].reg].t == LSX_CACHE_YMMW) {
+        if (dyn->lsx.avxcache[a].zero_upper == 1) {
+            XVXOR_V(SCRATCH, SCRATCH, SCRATCH);
+            XVINSVE0_D(dyn->lsx.avxcache[a].reg, SCRATCH, 2);
+            XVINSVE0_D(dyn->lsx.avxcache[a].reg, SCRATCH, 3);
```
ksco: Why do we need these 2 lines if this is only used in avx->sse?

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -976,47 +992,232 @@ static void sse_reflectcache(dynarec_la64_t* dyn, int ninst, int s1)
...
+        dyn->lsx.ssecache[a].v = -1;
+    }
+    dyn->lsx.avxcache[a].reg = fpu_get_reg_ymm(dyn, LSX_CACHE_YMMW, a);
+    dyn->lsx.avxcache[a].write = 1;
+    dyn->lsx.avxcache[a].width = width;
+    if (width == LSX_AVX_WIDTH_256) dyn->lsx.avxcache[a].zero_upper = 0;
+    return dyn->lsx.avxcache[a].reg;
+}
+
+void avx_purge_reg_upper128(dynarec_la64_t* dyn, int ninst, int a, int forwrite)
```
ksco: IMO this is not a good name for a function only used in avx->sse.

src/dynarec/la64/la64_emitter.h
```diff
@@ -1946,6 +2062,158 @@ LSX instruction starts with V, LASX instruction starts with XV.
...
 
+#define XVLD(vd, rj, imm12)         EMIT(type_2RI12(0b0010110010, imm12, rj, vd))
+#define XVST(vd, rj, imm12)         EMIT(type_2RI12(0b0010110011, imm12, rj, vd))
+#define XVLDX(vd, vj, vk)           EMIT(type_3R(0b00111000010010000, vk, vj, vd))
+#define XVSTX(vd, vj, vk)           EMIT(type_3R(0b00111000010011000, vk, vj, vd))
+
+#define XVPERMIQ_Vdu 0b11
+#define XVPERMIQ_Vdd 0b10
+#define XVPERMIQ_Vju 0b01
+#define XVPERMIQ_Vjd 0b00
```
ksco: Not sure about the meanings of these macros, but use raw number seems to be more readable to me.

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -898,6 +903,11 @@ int sse_get_reg_empty(dynarec_la64_t* dyn, int ninst, int s1, int a)
         dyn->lsx.lsxcache[dyn->lsx.ssecache[a].reg].t = LSX_CACHE_XMMW;
         return dyn->lsx.ssecache[a].reg;
     }
+    // upgrade from sse to avx
```
ksco: Does not seems to be an "upgrade" here

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -929,16 +939,22 @@ void sse_purge07cache(dynarec_la64_t* dyn, int ninst, int s1)
...
     for (int i = 0; i < 8; ++i)
-        if (dyn->lsx.ssecache[i].v != -1) {
+        if (dyn->lsx.ssecache[i].v != -1 || dyn->lsx.avxcache[i].v != -1) {
             if (old == -1) {
                 MESSAGE(LOG_DUMP, "\tPurge XMM0..7 Cache ------\n");
                 ++old;
             }
-            if (dyn->lsx.lsxcache[dyn->lsx.ssecache[i].reg].t == LSX_CACHE_XMMW) {
+            if (dyn->lsx.lsxcache[dyn->lsx.avxcache[i].reg].t == LSX_CACHE_YMMW) {
+                XVPERMI_Q(SCRATCH, dyn->lsx.avxcache[i].reg, XVPERMIQ_Vju);
```
ksco: I think we need to check `zero_upper` here.

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -976,47 +992,232 @@ static void sse_reflectcache(dynarec_la64_t* dyn, int ninst, int s1)
...
+        }
+        if (width == LSX_AVX_WIDTH_128) {
+            dyn->lsx.avxcache[a].width = LSX_AVX_WIDTH_128;
+        } else {
+            // if width changed to 256, and vzeroup ==1, means need zero-fill upper 128bits now.
+            if (dyn->lsx.avxcache[a].zero_upper == 1) {
+                dyn->lsx.avxcache[a].zero_upper = 0;
+                XVXOR_V(SCRATCH, SCRATCH, SCRATCH);
+                XVINSVE0_D(dyn->lsx.avxcache[a].reg, SCRATCH, 2);
+                XVINSVE0_D(dyn->lsx.avxcache[a].reg, SCRATCH, 3);
```
ksco: Use `XVPERMI_Q`?

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -976,47 +992,232 @@ static void sse_reflectcache(dynarec_la64_t* dyn, int ninst, int s1)
...
+void avx_reflect_reg(dynarec_la64_t* dyn, int ninst, int a)
+{
+    if (dyn->lsx.avxcache[a].v == -1)
+        return;
+    if (dyn->lsx.lsxcache[dyn->lsx.avxcache[a].reg].t == LSX_CACHE_YMMW) {
+        VST(dyn->lsx.avxcache[a].reg, xEmu, offsetof(x64emu_t, xmm[a]));
+        if (dyn->lsx.avxcache[a].zero_upper == 1) {
+            XVXOR_V(SCRATCH, SCRATCH, SCRATCH);
+            XVINSVE0_D(dyn->lsx.avxcache[a].reg, SCRATCH, 2);
+            XVINSVE0_D(dyn->lsx.avxcache[a].reg, SCRATCH, 3);
```
ksco: Use `XVPERMI_Q`?

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -976,47 +992,232 @@ static void sse_reflectcache(dynarec_la64_t* dyn, int ninst, int s1)
...
+        } else {
+            XVPERMI_Q(SCRATCH, dyn->lsx.avxcache[a].reg, XVPERMIQ_Vju);
+        }
+        VST(SCRATCH, xEmu, offsetof(x64emu_t, ymm[a]));
+        dyn->lsx.lsxcache[dyn->lsx.avxcache[a].reg].t = LSX_CACHE_YMMR;
```
ksco: Why is this marked as YMMR?

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -976,47 +992,232 @@ static void sse_reflectcache(dynarec_la64_t* dyn, int ninst, int s1)
...
+    int old = -1;
+    for (int i = 0; i < 16; ++i)
+        if (dyn->lsx.avxcache[i].v != -1) {
+            if (dyn->lsx.avxcache[i].write) {
+                if (old == -1) {
+                    MESSAGE(LOG_DUMP, "\tPurge %sAVX Cache ------\n", next ? "locally " : "");
+                    ++old;
+                }
+                if (dyn->lsx.lsxcache[dyn->lsx.avxcache[i].reg].t == LSX_CACHE_YMMW) {
+                    avx_reflect_reg(dyn, ninst, dyn->lsx.avxcache[i].reg);
```
ksco: Why use `avx_reflect_reg` here?

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -976,47 +992,232 @@ static void sse_reflectcache(dynarec_la64_t* dyn, int ninst, int s1)
...
+        MESSAGE(LOG_DUMP, "\t------ Purge AVX Cache\n");
+    }
+}
+
+static void avx_reflectcache(dynarec_la64_t* dyn, int ninst, int s1)
+{
+    for (int i = 0; i < 16; ++i) {
+        if (dyn->lsx.avxcache[i].v != -1 && dyn->lsx.avxcache[i].write) {
+            if (dyn->lsx.lsxcache[dyn->lsx.avxcache[i].reg].t == LSX_CACHE_YMMW) {
+                avx_reflect_reg(dyn, ninst, i);
```
ksco, line 1168: Why use `avx_reflect_reg` here?

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -976,47 +992,232 @@ static void sse_reflectcache(dynarec_la64_t* dyn, int ninst, int s1)
...
+    }
+
+    if (n) {
+        MESSAGE(LOG_DUMP, "\tPush XMM/YMM Cache (%d)------\n", n);
+        for (int i = start; i < 16; ++i) {
+            if ((dyn->lsx.ssecache[i].v != -1) && (dyn->lsx.ssecache[i].write)) {
+                VST(dyn->lsx.ssecache[i].reg, xEmu, offsetof(x64emu_t, xmm[i]));
+            }
+            if ((dyn->lsx.avxcache[i].v != -1) && (dyn->lsx.avxcache[i].write)) {
+                avx_reflect_reg(dyn, ninst, i);
```
ksco: Why use avx_reflect_reg here?

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -1026,13 +1227,19 @@ void fpu_reflectcache(dynarec_la64_t* dyn, int ninst, int s1, int s2, int s3)
...
     sse_reflectcache(dyn, ninst, s1);
+    avx_reflectcache(dyn, ninst, s1);
 }
 
 void fpu_unreflectcache(dynarec_la64_t* dyn, int ninst, int s1, int s2, int s3)
 {
     // TODO
 }
 
+void ymm_mark_zero(dynarec_la64_t* dyn, int ninst, int a)
```
ksco: This is never used.

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -433,6 +433,188 @@
...
 
+#define YMM0(a)                                             \
+    do {                                                    \
+        if (dyn->lsx.avxcache[a].v != -1) {                 \
+            dyn->lsx.avxcache[a].zero_upper = 1;            \
+            dyn->lsx.avxcache[a].width = LSX_AVX_WIDTH_128; \
+        }                                                   \
+    } while (0)
+
+#define YMM0_UNMARK_ZERO(a)                      \
```
ksco: Never used.

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -433,6 +433,188 @@
         BSTRINS_D(wback, ed, wb2 + 7, wb2); \
     }
 
+#define YMM0(a)                                             \
+    do {                                                    \
+        if (dyn->lsx.avxcache[a].v != -1) {                 \
+            dyn->lsx.avxcache[a].zero_upper = 1;            \
+            dyn->lsx.avxcache[a].width = LSX_AVX_WIDTH_128; \
```
ksco: Is this line necessary?

src/dynarec/la64/la64_emitter.h
ksco, line 1: It would be good to know how this code is written. Is it handwritten, or do you have a generator?

src/dynarec/la64/la64_printer.c
ksco, line 1: It would be good to know how this code is written. Is it handwritten, or do you have a generator?

ksco (changes requested):
Overall, this looks good. I left some comments. Please do smaller PRs next time!

src/dynarec/la64/dynarec_la64_avx.c
phorcys: got it.

src/dynarec/la64/dynarec_la64_avx_0f.c
phorcys: I'll fix all of those.

ksco: tips: you can use the review function in the web ui to add pending comments and submit them once, so there won’t be 30 emails in my mailbox 😉

src/dynarec/la64/dynarec_la64_avx_0f.c
phorcys: Old code that I forgot to fix,  :-(
But it's working by accident.
I'll fix it.

src/dynarec/la64/dynarec_la64_avx_0f.c
phorcys: zero_upper is set in  avx_get_reg_empty().
If you get an empty reg, either 128 or 256,  zero_upper shoule be automatic cleared and write is set.
If you get an 256-width reg, 
avx_get_reg will do automatic zero fill upper 128bits (zero_upper==1 && pre-width==128)

src/dynarec/la64/dynarec_la64_avx_66_0f.c
phorcys: yes, you are right.

src/dynarec/la64/dynarec_la64_avx_66_0f.c
phorcys: VMOVD (VEX-Encoded Version when Destination is an XMM Register) 
DEST[31:0] := SRC[31:0]
DEST[MAXVL-1:32] := 0

Your question might be "Can we use XVXOR_V, so we can save one YMM0"
But YMM0 itself is zero cost. Only emit  XVXOR_V+XVPERMI_Q when  register purged/pushed.
mark YMM0 or XVXOR_V  make no different.

src/dynarec/la64/dynarec_la64_avx_66_0f.c
phorcys: I'll fix it.

src/dynarec/la64/dynarec_la64_avx_0f.c
phorcys: removed.

src/dynarec/la64/dynarec_la64_avx_66_0f38.c
phorcys: Fixed.

src/dynarec/la64/dynarec_la64_avx_66_0f3a.c
phorcys: Deleted.

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
phorcys: Fixed VMOVSD/VMOVSS

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
phorcys: same as above.

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
phorcys: Fixed.

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
phorcys: My mistake, fixed.

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
phorcys:
```
VEX.LIG.F2.0F.WIG 11 /r VMOVSD xmm1, xmm2, xmm3 	E 	V/V 	AVX 	Merge scalar double precision floating-point value from xmm2 and xmm3 registers to xmm1.
VEX.LIG.F2.0F.WIG 11 /r VMOVSD m64, xmm1 	C 	V/V 	AVX 	Store scalar double precision floating-point value from xmm1 register to m64.
```
Vx  is not used when dest operand  is memory.

src/dynarec/la64/dynarec_la64_avx_f3_0f.c
phorcys: Fixed.

src/dynarec/la64/dynarec_la64_avx_f3_0f.c
phorcys:
```
VEX.LIG.F3.0F.WIG 11 /r VMOVSS xmm1, xmm2, xmm3 	E 	V/V 	AVX 	Move scalar single precision floating-point value from xmm2 and xmm3 to xmm1 register.
VEX.LIG.F3.0F.WIG 11 /r VMOVSS m32, xmm1 	C 	V/V 	AVX 	Move scalar single precision floating-point value from xmm1 register to m32.
```
same as VMOVSD

src/dynarec/la64/dynarec_la64_helper.c
phorcys: You are right, should use migrate or changed.

src/dynarec/la64/dynarec_la64_helper.c
phorcys: Old code for  native LASX, should removed.

src/dynarec/la64/dynarec_la64_helper.c
phorcys: Yes, it's not a good name.
How about avx_reflect_reg_upper128?

src/dynarec/la64/dynarec_la64_helper.c
phorcys: changed to migrate

src/dynarec/la64/la64_emitter.h
phorcys: We can switch to raw numbers, or use some macro like VEXTRINS_IMM( high , low), eg: XVPERMIQ_IMM(high, low)
Both good for me, which one do you prefer?

src/dynarec/la64/dynarec_la64_helper.c
phorcys: Yes, you are right.
I'll fix it.

src/dynarec/la64/dynarec_la64_helper.c
phorcys: by https://jia.je/unofficial-loongarch-intrinsics-guide

xvpermi.q  latency 3 IPC 4
xvinsve0.d latency 1 IPC 4

Two XVINSVE0_D looks ugly,  but  a little faster...

src/dynarec/la64/dynarec_la64_helper.c
phorcys: same as above.

src/dynarec/la64/dynarec_la64_helper.c
phorcys: YMMW=>dirty ( un-synced with x64emu_t)
YMMR=>clean ( synced with x64emu_t)

Try reduce future store op emits (if this reg is not used as a dst reg)

src/dynarec/la64/dynarec_la64_helper.c
phorcys: My mistake.
fpu push/pop should skip ymm store if zero_upper==1

src/dynarec/la64/dynarec_la64_helper.c
rajdakin: Be careful about cache metadata poisoning though: if you have something like
```
  ; ymm is YMMW here
  jz some_later_code
  use ymm
some_later_code:
  use ymm
```
then both `use ymm` should still have `LSX_CACHE_YMMW`, so that regardless of the conditional jump the register is synced with the `x64emu_t` (I don't know if this is possible with your code, but it looks like it is pretty easy to accidentally create the issue later even if it is not right now). A similar bug did happen before, and it was pretty hard to find and debug.

src/dynarec/la64/dynarec_la64_helper.h
phorcys: I guess not.
Maybe assert here is better, or just use zero_upper=1 is enough.

src/dynarec/la64/dynarec_la64_helper.h
phorcys: removed

src/dynarec/la64/la64_emitter.h
phorcys: I use 2 python script to generate  jit/printer code from binutils loongarch-opc.c
printer: https://gist.github.com/phorcys/c7db46f11ff02745e2bdd8252db65bf4
emitter: https://gist.github.com/phorcys/05c6de04e8692ac4b806017cd0d1b79e

src/dynarec/la64/dynarec_la64_helper.c
phorcys: > Be careful about cache metadata poisoning though: if you have something like
> 
> ```
>   ; ymm is YMMW here
>   jz some_later_code
>   use ymm
> some_later_code:
>   use ymm
> ```
> 
> then both `use ymm` should still have `LSX_CACHE_YMMW`, so that regardless of the conditional jump the register is synced with the `x64emu_t` (I don't know if this is possible with your code, but it looks like it is pretty easy to accidentally create the issue later even if it is not right now). A similar bug did happen before, and it was pretty hard to find and debug.

Thank you. I think that I should remove YMMW override.

src/dynarec/la64/dynarec_la64_avx_66_0f.c
ksco: sorry I misread the `VXOR_V` as `XVXOR_V`, but marking YMM0 dirty is not zero cost, because it means we may need to clear it after.

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
ksco: Same as what?

src/dynarec/la64/dynarec_la64_helper.c
ksco: I doubt that it's faster than a single `XVPERMI_Q`, unless there are proofs. And if that is true, you should use it everywhere.

phorcys: I did a microbech, xvpermi.q is faster then 2 vxinsve0.d in this patch's scene(write same reg), slower when multi issue( write different reg). I'll fix this.

src/dynarec/la64/dynarec_la64_avx_66_0f.c
phorcys: I means   YMM0  vs  XVXOR_V  both set high 128bits to zero. both causing a pending store(store high 128) in current code. So they are same now.

XVXOR_V would be better when reg is reused in next in-block insts as 256-width.
And to use XVXOR_V , you need to mark reg as 256-width, or cleared high bits will be ignored in current code.
Maybe we can do this opt later when most avx code imp landed.

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
phorcys: VMOVD's  comment.

src/dynarec/la64/dynarec_la64_helper.c
phorcys: It's the only place avx_reflect_reg used.

phorcys (commented):
updated code.

ksco:
If this is ready for another round, please use the re-request review button.

phorcys:
> If this is ready for another round, please use the re-request review button.
图片

You didn't push "request change", so there is no "request review" button

src/dynarec/la64/dynarec_la64_avx_66_0f.c
ksco: Sorry I don't understand. My suggestion here is to change line 153 to `XVXOR_V` and change line 159 to `zero_upper = 0`, so we don't have to clear the upper 128 bits again later.

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
ksco: Could you just repeat the comment? I can't find that easily via the web UI...

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -976,47 +996,238 @@ static void sse_reflectcache(dynarec_la64_t* dyn, int ninst, int s1)
...
+        dyn->lsx.ssecache[a].v = -1;
+    }
+
+    // new reg
+    dyn->lsx.avxcache[a].reg = fpu_get_reg_ymm(dyn, forwrite ? LSX_CACHE_YMMW : LSX_CACHE_YMMR, a);
+    int ret = dyn->lsx.avxcache[a].reg;
+    dyn->lsx.avxcache[a].write = forwrite;
+    dyn->lsx.avxcache[a].width = width;
+    if (width == LSX_AVX_WIDTH_128) {
+        VLD(ret, xEmu, offsetof(x64emu_t, xmm[a]));
```
ksco, line 1038: Don't we need to set zero_upper = 1 here?

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -976,47 +996,238 @@ static void sse_reflectcache(dynarec_la64_t* dyn, int ninst, int s1)
...
+
+    if (n) {
+        MESSAGE(LOG_DUMP, "\tPush XMM/YMM Cache (%d)------\n", n);
+        for (int i = start; i < 16; ++i) {
+            if ((dyn->lsx.ssecache[i].v != -1) && (dyn->lsx.ssecache[i].write)) {
+                VST(dyn->lsx.ssecache[i].reg, xEmu, offsetof(x64emu_t, xmm[i]));
+            }
+            if ((dyn->lsx.avxcache[i].v != -1) && (dyn->lsx.avxcache[i].write)) {
+                    VST(dyn->lsx.avxcache[i].reg, xEmu, offsetof(x64emu_t, xmm[i]));
+                    if (dyn->lsx.avxcache[i].zero_upper == 0) {
```
ksco, line 1194: If `zero_upper == 1`, don't we need to VST zero to ymm[i] here?

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -976,47 +996,238 @@ static void sse_reflectcache(dynarec_la64_t* dyn, int ninst, int s1)
...
+
+    if (n) {
+        MESSAGE(LOG_DUMP, "\tPop XMM/YMM Cache (%d)------\n", n);
+        for (int i = start; i < 16; ++i) {
+            if (dyn->lsx.ssecache[i].v != -1) {
+                VLD(dyn->lsx.ssecache[i].reg, xEmu, offsetof(x64emu_t, xmm[i]));
+            }
+            if (dyn->lsx.avxcache[i].v != -1) {
+                VLD(dyn->lsx.avxcache[i].reg, xEmu, offsetof(x64emu_t, xmm[i]));
+                if (dyn->lsx.avxcache[i].zero_upper == 0) {
```
ksco, line 1222: If `zero_upper == 1`, don't we need to set ymm[i] to zero here?

ksco:
> You didn't push "request change", so there is no "request review" button

Sorry, I must have selected "Comment" instead of "Request changes".

src/dynarec/la64/dynarec_la64_avx_66_0f.c
phorcys: current reg is  128-width with YMM0, compare  YMM0 with XVXOR_V
|     | YMM0 | XVXOR_V |
| -------- | ------- |------- |
| state change needed:  | mark zero_upper=1    | change width to 256|
| next inst use this reg is 128-width avx |  mark ag, zero cost     |  mark ag zero cost |
| next inst use this reg is 256-width avx    | emit xvpermi.q    |  zero cost| 
| next inst use this reg is sse or push/purged|  emit xvpermi.q+ vst | emit xvpermi.q +vst|

XVXOR_V is better if insts in block reuse this reg as 256-width avx.
But you need to mark this reg as 256-width , or upper cleared zero will be ignored.
Which will cause a  VEX.128 inst   result a  VEX.256 reg.
This is why I didn't use XVXOR_V.

If you think this is not a problem, I'm ok to opt those insts with  XVXOR_V, since there is a chance to save 2 emit insts.

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
phorcys: current reg is  128-width with YMM0, compare  YMM0 with XVXOR_V
|     | YMM0 | XVXOR_V |
| -------- | ------- |------- |
| state change needed:  | mark zero_upper=1    | change width to 256|
| next inst use this reg is 128-width avx |  mark ag, zero cost     |  mark ag zero cost |
| next inst use this reg is 256-width avx    | emit xvpermi.q    |  zero cost| 
| next inst use this reg is sse or push/purged|  emit xvpermi.q+ vst | emit xvpermi.q +vst|

XVXOR_V is better if insts in block reuse this reg as 256-width avx.
But you need to mark this reg as 256-width , or upper cleared zero will be ignored.
Which will cause a  VEX.128 inst   result a  VEX.256 reg.
This is why I didn't use XVXOR_V.

If you think this is not a problem, I'm ok to opt those insts with  XVXOR_V, since there is a chance to save 2 emit insts.

src/dynarec/la64/dynarec_la64_helper.c
phorcys: No we don't，we set zero_upper=1 in insts code.
We can set zero_upper=1 here, to default all 128-width avx reg zero_upper=1.
Or we set them in insts code.
I choosed to use YMM0 in insts code, to make it more readable. 
And if we need to do some opt on this, have YMM0 in every 128-width insts would be easier.

src/dynarec/la64/dynarec_la64_helper.c
phorcys: This is fpu push/pop.
In my understanding:
avxcache is unchanged during fpu push/pop(call).
so zero_upper=1 is keeped, so we can skip  vst(upper_zero ，ymm) and vld(upper_zero, ymm)
Correct me if I'm wrong, I'll fix it then.

src/dynarec/la64/dynarec_la64_helper.c
phorcys: This is fpu push/pop.
In my understanding:
avxcache is unchanged during fpu push/pop(call).
so zero_upper=1 is keeped, so we can skip vst(upper_zero ，ymm) and vld(upper_zero, ymm)
Correct me if I'm wrong, I'll fix it then.

src/dynarec/la64/dynarec_la64_avx_66_0f.c
ksco: Why do we have to mark this reg as 256-bit to use `XVXOR_V`?

src/dynarec/la64/la64_emitter.h
ksco: Both are fine to me.

src/dynarec/la64/dynarec_la64_helper.c
ksco: Hmmm, but if this can be done in the infra automatically, I don't see the necessity of manually setting YMM0 in the opcode implementation code.

ksco: Nah I don't think this can be done automatically, nvm.

ksco:
So there are only a few pending comments left. After we resolved all that, I'll do a full review again. @phorcys

src/dynarec/la64/dynarec_la64_helper.c
phorcys: Yes, but I'm not sure about it.
So I keeped YMM0, at least it won't wrong.
VZEROUPPER and VZEROALL is  special, but they are not problems.
I'm OK with both of them.

src/dynarec/la64/dynarec_la64_avx_66_0f.c
phorcys: > Why do we have to mark this reg as 256-bit to use `XVXOR_V`?

yes, you are right. If vzero_upper=0, VEX.128 => VEX.256 reg no need to set width to 256.
This opt is worth adopt.
I'll fix it.

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
phorcys: yes, you are right. If vzero_upper=0, VEX.128 => VEX.256 reg no need to set width to 256.
This opt is worth adopt.
I'll fix it.

phorcys:
I've fixed all the fix required, please review them ag when you have free time. Thanks. @ksco

src/dynarec/la64/dynarec_la64_avx_0f.c
```diff
@@ -0,0 +1,188 @@
...
+        case 0x10:
+            INST_NAME("VMOVUPS Gx, Ex");
+            nextop = F8;
+            GETGYxy_empty(q0);
+            if (MODREG) {
+                GETEYxy(q1, 0, 0);
+                if (vex.l) {
+                    XVOR_V(q0, q1, q1);
+                } else {
+                    VOR_V(q0, q1, q1);
```
ptitSeb, line 68: I don't understand. this is the same opcode than the non-AVX one. But this one have to 0-out the upper 128bits of q0, while on thenon-AVX version MOVUPS the upper 128bits should be untouched. How will this works?

ptitSeb: Ah, it's all in the GETGYxy_empty actualy, the `avx_get_reg_empty` is the one doing the wipe if needed, I see...

src/dynarec/la64/dynarec_la64_avx_66_0f38.c
```diff
@@ -0,0 +1,93 @@
...
+            if(vex.l){
+                XVREPLVE0_D(q0, q2);
+            }else{
+                VREPLVE_D(q0, q2, 0);
+            }
+            break;
+        case 0x1A:
+            INST_NAME("VBROADCASTF128 Gx, Ex");
+            nextop = F8;
+            GETGYy_empty(q0);
```
ptitSeb: It is usualy non safe to gather a register as "empty" first, in case it's also one of the next regs...like if `VBROADCASTF128 ymm0, xmm0`

src/dynarec/la64/dynarec_la64_avx_66_0f.c
```diff
@@ -0,0 +1,225 @@
...
+    MAYUSE(j64);
+    MAYUSE(cacheupd);
+
+    rex_t rex = vex.rex;
+
+    switch (opcode) {
+        case 0x10:
+            INST_NAME("VMOVUPD Gx, Ex");
+            nextop = F8;
+            GETGYxy_empty(q0);
```
ptitSeb: _empty gathering should be last

src/dynarec/la64/dynarec_la64_avx_66_0f.c
```diff
@@ -0,0 +1,225 @@
...
+                } else {
+                    VST(q0, ed, fixedaddress);
+                }
+                SMWRITE2();
+            }
+            break;
+        case 0x28:
+            INST_NAME("VMOVAPD Gx, Ex");
+            nextop = F8;
+            GETGYxy_empty(q0);
```
ptitSeb: _empty gathering should be last

src/dynarec/la64/dynarec_la64_avx_66_0f.c
```diff
@@ -0,0 +1,225 @@
...
+                XVINSGR2VR_D(q0, ed, 0);
+            } else {
+                XVINSGR2VR_W(q0, ed, 0);
+            }
+            YMM_UNMARK_UPPER_ZERO(q0);
+            break;
+        case 0x6F:
+            INST_NAME("VMOVDQA Gx, Ex");
+            nextop = F8;
+            GETGYxy_empty(q0);
```
ptitSeb: _empty gathering should be last

src/dynarec/la64/dynarec_la64_avx_66_0f38.c
```diff
@@ -0,0 +1,93 @@
...
+    MAYUSE(j64);
+    MAYUSE(cacheupd);
+
+    rex_t rex = vex.rex;
+
+    switch (opcode) {
+        case 0x18:
+            INST_NAME("VBROADCASTSS Gx, Ex");
+            nextop = F8;
+            GETGYxy_empty(q0);
```
ptitSeb: _empty gathering should be last

src/dynarec/la64/dynarec_la64_avx_66_0f38.c
```diff
@@ -0,0 +1,93 @@
...
+            if(vex.l){
+                XVREPLVE0_W(q0, q2);
+            }else{
+                VREPLVE_W(q0, q2, 0);
+            }
+            break;
+        case 0x19:
+            INST_NAME("VBROADCASTSD Gx, Ex");
+            nextop = F8;
+            GETGYxy_empty(q0);
```
ptitSeb: _empty gathering should be last

src/dynarec/la64/dynarec_la64_avx_66_0f3a.c
```diff
@@ -0,0 +1,101 @@
...
+    switch (opcode) {
+        case 0x18:
+        case 0x38:
+            if (opcode == 0x18) {
+                INST_NAME("VINSERTF128 Gx, Vx, Ex, imm8");
+            } else {
+                INST_NAME("VINSERTI128 Gx, Vx, Ex, imm8");
+            }
+            nextop = F8;
+            GETGYy_empty(q0);
```
ptitSeb: _empty gathering should be last

src/dynarec/la64/dynarec_la64_avx_f2_0f.c
```diff
@@ -0,0 +1,100 @@
...
+    MAYUSE(cacheupd);
+
+    rex_t rex = vex.rex;
+
+    switch (opcode) {
+
+        case 0x10:
+            INST_NAME("VMOVSD Gx, [Vx,] Ex");
+            nextop = F8;
+            GETGYx_empty(q0);
```
ptitSeb: _empty gathering should be last

src/dynarec/la64/dynarec_la64_avx_f3_0f.c
```diff
@@ -0,0 +1,128 @@
...
+    int64_t fixedaddress, gdoffset, vxoffset;
+    int unscaled;
+
+    rex_t rex = vex.rex;
+
+    switch (opcode) {
+        case 0x10:
+            INST_NAME("VMOVSS Gx, [Vx,] Ex");
+            nextop = F8;
+            GETGYx_empty(q0);
```
ptitSeb: _empty gathering should be last

src/dynarec/la64/dynarec_la64_avx_f3_0f.c
```diff
@@ -0,0 +1,128 @@
...
+            } else {
+                addr = geted(dyn, addr, ninst, nextop, &ed, x2, x1, &fixedaddress, rex, NULL, 1, 0);
+                FST_S(q2, ed, fixedaddress);
+                SMWRITE2();
+            }
+            break;
+        case 0x6F:
+            INST_NAME("VMOVDQU Gx, Ex");
+            nextop = F8;
+            GETGYxy_empty(q0);
```
ptitSeb: _empty gathering should be last

ptitSeb (changes requested):
It is usualy non safe to gather a register as "empty" first, in case it's also one of the next regs...like if `VBROADCASTF128 ymm0, xmm0`

src/dynarec/la64/dynarec_la64_avx_66_0f38.c
phorcys: I got it.I'll fix all the *empty order.
Guess it would be better to use some macro eg: GETGXEMPTY_VX_EX to ensure this order?
@ksco

ptitSeb: Yeah, that's why I created those macro on ARM64, to avoid thinking of the order (but those macro are not always usefull 100% the time, and in some case you still need to use individual macro... Sometime you will need to duplicate the `GETGYxy_empty(...)` in the 2 part of the `if(MODREG)`...

phorcys:
@ptitSeb  @ksco 
Macro added.
All order fixed ,please review ag.

ptitSeb:
Merged! :tada:

Issue #2740, opened 2025-06-13
Why does Box64 use a fixed 128 bitmap for small allocations?


devarajabc:
In `src/custommem.c` the function:
```c
void* map128_customMalloc(size_t size, int is32bits) {
    size = 128;
    …
}

```
hard-codes every allocation to 128 bytes. 
However, In my profiling (box64 + wine just like #2511 )
most allocation requests are far smaller (e.g. ≤ 56 bytes), so this can increase internal fragmentation.

![Image](https://github.com/user-attachments/assets/1cab3010-826f-4204-983a-09e8d6aeea7b)

4 | 1030
-- | --
8 | 85
16 | 716
32 | 268
40 | 14
56 | 7998
64 | 34
96 | 81
120 | 1152
128 | 24

#### Questions: 
1. What was the original design rationale for fixing the block size at 128?

2. Is it for alignment, page-size granularity, mmap overhead, or something else?

3. Would it be beneficial to reduce this to 64?

I’m happy to draft a PR to change the default to 64 if that would help.

### Environment
1. Box64 version: Box64 riscv64 v0.3.7 926a5980 with Dynarec built on Jun 13 2025
2. Build flags:
  ```
cmake -G Ninja \
  -DBOX32=ON \
  -DRV64=1 \
  -DRV64_DYNAREC=ON \
  -DGDBJIT=ON \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
  -DCMAKE_C_COMPILER=gcc \
  ..
```
3. Platform: RISC-V 64, Debian (chroot)
4. CPU: 4× cores(Andes AX45MP), Little Endian
5. page size 4096
6. Kernel: 6.1.47+
7. gcc :(Debian 14.2.0-19) 14.2.0
8. glibc :(Debian GLIBC 2.41-7) 2.41

Thanks!

ptitSeb:
Rationnale was: simplicity vs speed.

Adding a 64bytes bitmap allocator is doable, and quite easy, but that's a lot of code for a limited benefit.

Even memory-wise, adding a 64bytes allocator would not save much. In your example, that would 64*10145 so around 634k... Not even a megabytes.

Feel free to add a 64bytes allocator if you want, via a PR...

PR #2764, opened 2025-06-23
Add `map64_customMalloc` to reduce internal fragmentation

devarajabc:
Introduce `map64_customMalloc` to reduce internal fragmentation

This commit adds `map64_customMalloc`, a 64-byte–chunk variant of `map128_customMalloc`, to address internal fragmentation when most allocations are around 56 bytes. The new allocator uses a `uint16_t` map array—each bit in a 16-bit entry tracks the in-use status of one 64-byte chunk (16 chunks per entry), with the bitmap region fixed at 256 bytes. Here, “map” is a fast-lookup structure distinct from the bitmap itself. All chunk-size, shift, and map-size calculations have been updated accordingly.

Relates to #2740

---
**TODO:** decide whether to replace existing calls to `map128_customMalloc` with `map64_customMalloc`, or to keep both allocators available side-by-side.

### DIFF
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -81,6 +81,7 @@ static rbtree_t*  blockstree = NULL;
 
 #define BTYPE_MAP   1
 #define BTYPE_LIST  0
+#define BTYPE_MAP64 2
 
 typedef struct blocklist_s {
     void*               block;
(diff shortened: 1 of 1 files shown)


src/custommem.c
```diff
@@ -787,21 +893,28 @@ void* internal_customRealloc(void* p, size_t size, int is32bits)
     blocklist_t* l = findBlock(addr);
     if(l) {
         size_t subsize;
-        if(l->type==BTYPE_LIST) {
+        if(l->type == BTYPE_LIST) {
```
ptitSeb, line 896: This is nit piking and not important, but it is strange to me that `==` is forced to have space before and after, but `<=`, as seen a few lines below, does not...

(I personaly doesn't like the spaces in if for `==`, `!=`and the like, and prefer space between individual test, so with `&&` or `||` instead. I do put space in affectation tho)

PR #2765, opened 2025-06-23
[WRAPPER] additional wrapped symbols

airidosas252:
They seem to quiet down recent dlsym errors I got with libiconv.so and libodbc.so too.

For some reason, they are required now? 

Anyway, box64 is happy and a native libiconv.so can be used now.

Tested only on Android (bionic).

I might be missing some stuff, so apologies for my lazyness.


src/wrapped/wrappediconv_private.h
```diff
@@ -2,3 +2,14 @@
 #error Meh...
 #endif
 
+GO(libiconv_open, pFpp)
+GO(locale_charset, pFv)
+GO(libiconv_close, iFp)
+GO(libiconvctl, iFpip)
+GO(libiconv_set_relocation_prefix, vFpp)
+GO(libiconv, iFpppp)
```
ptitSeb: for me, signature of this one is 
```C
size_t iconv (iconv_t icd, const char* * inbuf, size_t *inbytesleft, char* * outbuf, size_t *outbytesleft);
```
with `libiconv` being an alias to  `iconv` , so that would make the signature `LFppppp` Where did you found your signature?

airidosas252: Ok, I'm still mistaking those signatures yet again...

I was investigating those on the internet and I still got them wrong... I'm really sorry.

ptitSeb: will you fix the signature in this PR?

airidosas252: Will do, but a bit later. Currently at work.

ptitSeb: Tnaks! Take your time, there is no rush

PR #2777, opened 2025-06-26
fix: Ensure ed equals x3 in "VPMASKMOVD/Q Ex, Vx, Gx"

ye-yeshun:
I encountered a CEF-based application that failed to launch properly. After debugging, I confirmed the issue was related to VPMASKMOVD/Q instructions - it doesn't verify if ed equals x3.

### DIFF
--- a/src/dynarec/arm64/dynarec_arm64_avx_66_0f38.c
+++ b/src/dynarec/arm64/dynarec_arm64_avx_66_0f38.c
@@ -1297,6 +1297,10 @@ uintptr_t dynarec64_AVX_66_0F38(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip
                         s0 = -1;
                         v1 = fpu_get_scratch(dyn, ninst);
                         addr = geted(dyn, addr, ninst, nextop, &ed, x3, &fixedaddress, NULL, 0, 0, rex, NULL, 0, 0);
+                        if(ed!=x3) {
+                            MOVz_REG(x3, ed);
+                            ed = x3;
+                        }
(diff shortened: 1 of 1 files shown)


src/dynarec/arm64/dynarec_arm64_avx_66_0f38.c
```diff
@@ -1221,6 +1221,10 @@ uintptr_t dynarec64_AVX_66_0F38(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip
                     if(!l) {
                         GETGX_empty_VX(v0, v2);
                         addr = geted(dyn, addr, ninst, nextop, &ed, x3, &fixedaddress, NULL, 0, 0, rex, NULL, 0, 0);
+                        if(ed!=x3) {
```
ptitSeb, line 1224: But this case is handled bellow. line 1313-1314 x3 (or 1328-1329 for 32bits) is then used, and ed=x3 if it wasn't before, It was done liket that to avoid a move and still assure ed is not changed if it's a x64 register.

Are you sure this fixes your issue?

ye-yeshun: Oops, my bad.     
When I initially tested, my local branch wasn't synced to the latest code. While I did sync before submitting, I only checked the code after 'geted'.     
Please close this PR - I've just verified that lines 1313-1314 actually fix my issue.

ptitSeb: Ok. Nice debugging btw, it took me some time also to found this issue...

PR #2782, opened 2025-06-27
[LA64_DYNAREC] Removed some TABLE64 usage

ksco:


src/dynarec/la64/dynarec_la64_00.c
```diff
@@ -2384,7 +2384,7 @@ uintptr_t dynarec64_00(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                     if (dyn->last_ip && (addr - dyn->last_ip < 0x800)) {
                         ADDI_D(x2, xRIP, addr - dyn->last_ip);
                     } else {
-                        TABLE64(x2, addr);
+                        MOV64x(x2, addr);
```
ptitSeb, line 2387: just to note: this will need a table64 when relocation is done on this arg (unless you add relocation on MOV64x type, wich is doable, but anoying to do), because addr is relative the mmap, wich can be moved on another load.

src/dynarec/la64/dynarec_la64_00.c
```diff
@@ -2485,7 +2476,7 @@ uintptr_t dynarec64_00(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                         // jumps out of current dynablock...
                         MARK;
                         j64 = getJumpTableAddress64(addr);
-                        TABLE64(x4, j64);
+                        MOV64x(x4, j64);
```
ptitSeb, line 2479: same here, this will need a special reloc when it's implemented.

src/dynarec/la64/dynarec_la64_00.c
ksco: yeah yeah I’m aware, this is to align the implementation before finishing dynacache so I won’t miss anything.

PR #2799, opened 2025-07-05
[WRAPPER] Wrapper xmlMemGet and xmlMemSetup

xiangzhai:
Hi,

> [Scilab 6.0.1](https://www.scilab.org/download/6.0.1/scilab-6.0.1.bin.linux-x86_64.tar.gz) just moved forward, it reproduced other wrappedxml2 issues, I will fix them :)

Fix them now :)

Please review my patch.

Thanks,
Leslie Zhai

### DIFF
--- a/src/wrapped/wrappedxml2.c
+++ b/src/wrapped/wrappedxml2.c
@@ -1235,6 +1235,98 @@ static void* find_xmlTextReaderErrorFunc_Fct(void* fct) // this one have a VAArg
     return NULL;
 }
 
+// xmlFreeFunc ...
+#define GO(A)   \
+static uintptr_t my_xmlFreeFunc_fct_##A = 0;                      \
+static void my_xmlFreeFunc_##A(void* a)                          \
+{                                                                 \
+    return (void)RunFunctionFmt(my_xmlFreeFunc_fct_##A, "p", a); \
+}
+SUPER()
+#undef GO
+static void* find_xmlFreeFunc_Fct(void* fct)
+{
+    if(!fct) return fct;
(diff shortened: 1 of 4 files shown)


src/wrapped/wrappedxml2.c
```diff
@@ -1491,4 +1583,16 @@ EXPORT void* my_xmlCtxtReadIO(x64emu_t* emu, void* ctxt, void* ioread, void* ioc
     return my->xmlCtxtReadIO(ctxt, find_xmlInputReadCallback_Fct(ioread), find_xmlInputCloseCallback_Fct(ioclose), ioctx, url, encoding, options);
 }
 
+EXPORT int my_xmlMemGet(x64emu_t* emu, void* v1, void* v2, void* v3, void* v4)
+{
+    int ret = my->xmlMemGet(find_xmlFreeFunc_Fct(v1), find_xmlMallocFunc_Fct(v2), find_xmlReallocFunc_Fct(v3), find_xmlStrdupFunc_Fct(v4));
```
ptitSeb: No, this is not correct. The xmlMemGet gives back the functions used. So the function should be called, and then each pointer (that has been filled) should be run through a "reverse_"  finder of wrapped function.

ptitSeb (changes requested):
xmlMemGet is incorrect, please fix (it might work without any special wrapping by the way).

xiangzhai:
Thanks for pointing out my fault! Please teach me if I need to wrapper a new library, for example [libtk](https://www.tcl-lang.org/), in future :)

src/wrapped/wrappedxml2.c
Talambaye: ``Répondre

PR #2811, opened 2025-07-13
[LA64_DYNAREC] Fix la64 VMASKMOVPS,VMOVHPD.

phorcys:


src/dynarec/la64/dynarec_la64_avx_66_0f38.c
```diff
@@ -200,7 +200,7 @@ uintptr_t dynarec64_AVX_66_0F38(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t i
         case 0x2E:
             INST_NAME("VMASKMOVPS Ex, Gx, Vx");
```
ksco: Are you sure it's `Ex, Gx, Vx`, in this order? 

... nope, it follows this table:

| Op/En | Operand 1     | Operand 2    | Operand 3     | Operand 4 |
| ----- | ------------- | ------------ | ------------- | --------- |
| RVM   | ModRM:reg (w) | VEX.vvvv (r) | ModRM:r/m (r) | N/A       |
| MVR   | ModRM:r/m (w) | VEX.vvvv (r) | ModRM:reg (r) | N/A       |

Oh I saw the Arm64 side (@ptitSeb) has this weird order too, I think it's a mistake.

src/dynarec/la64/dynarec_la64_avx_66_0f38.c
```diff
@@ -215,7 +215,7 @@ uintptr_t dynarec64_AVX_66_0F38(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t i
         case 0x2F:
             INST_NAME("VMASKMOVPD Ex, Gx, Vx");
```
ksco: Ditto.

phorcys:
@ksco  fixed wrong operand order.

ksco:
You may want to fix Arm64 side too, I think it's okay you do it in this PR.

phorcys:
> You may want to fix Arm64 side too, I think it's okay you do it in this PR.

Done

ksco (approved):

src/dynarec/la64/dynarec_la64_avx_66_0f38.c
ptitSeb: Just to be clear, the comment / INST_NAME might be confusiong / incorrect, but the code is exact. Mask is VX, EX/GX are destination/source, so ARM64 code is correct.

Most game that use AVX use this opcode (like satisfactory for example), it's been tested quite a lot already.

PR #2815, opened 2025-07-14
Collapse `type` and `is32bits` into single `flags` byte

devarajabc:
Combine the two uint8_t fields `type` and `is32bits` in `blocklist_t` into a single uint8_t `flags` field, using bit 0 for `is32bits` and bits 1–7 for `type`. Although this change doesn’t reduce overall memory use today (the bitmap allocator still works in 64‑byte chunks), it cleans up the struct and lays the groundwork for future allocator updates where per‑byte savings will matter.


src/custommem.c
```diff
@@ -89,8 +89,7 @@ typedef struct blocklist_s {
     size_t              size;
     void*               first;
     uint32_t            lowest;
-    uint8_t             type;       // could use 7bits for type and 1bit fot is32bits,
-    uint8_t             is32bits;   // but that wont really change the size of structure anyway
```
ksco, line 93: The comments here literally explained why two uint8_t are preferred.

src/custommem.c
```diff
@@ -89,8 +89,7 @@ typedef struct blocklist_s {
...
+    uint8_t            flags; // flags = (type << 1) | (is32bits & 1);
```
ptitSeb, line 92: no, why using this complex thing now?!

ptitSeb:
For now, I don't see any valid reason to colapse type and is32bits into the same memory area.

devarajabc:
My intention was to lay the groundwork for a future allocator that might work in smaller chunks. Right now, this is more of a style tweak than a functional improvement.

I may have misread the comment’s intent—sorry about that! 
Feel free to  close this PR if you’d rather keep the fields separate until there’s a real need.

ptitSeb:
Yeah, I prefer to keep them separate for now, as it will just generate (a bit) more code any other way. Once there is a use to regroup, then no problem (but please, use bitfield and not shift magic unless really necessary, it maks things more readable).

PR #2825, opened 2025-07-17
box64-bundle-x86-libs.csv: fix libffi library

fboudra:
libffi-devel package is empty and contains only a symlink. Use the correct package to bundle the library binary.


fboudra:
@LukeShortCloud

PR #2829, opened 2025-07-19
[BUNDLE] Check for broken symlinks

LukeShortCloud:
to prevent issues such as #2825

### DIFF
--- a/box64-bundle-x86-libs.sh
+++ b/box64-bundle-x86-libs.sh
@@ -101,6 +101,13 @@ for lib in libc.so libpthread.so librt.so libGL.so libGL.so libX11.so \
     do find "${dir_tmp_local}" -name "${lib}"* -delete
 done
 
+if find "${dir_tmp_local}"/bundle-libs/ -type l ! -exec test -e {} \; -print | grep bundle-libs
+    then echo "Broken symlinks found."
+    cd -
+    echo rm -r -f "${dir_tmp_local}"
+    exit 1
+fi
+
 mv "${dir_tmp_local}"/*.deb "${dir_tmp_local}/bundle-pkgs/"
 mv "${dir_tmp_local}"/*.eopkg "${dir_tmp_local}/bundle-pkgs/"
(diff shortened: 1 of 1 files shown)


LukeShortCloud:
For testing purposes, I temporarily reverted https://github.com/ptitSeb/box64/commit/b1613be0f07ced52600f3dc40e2a04a6b8cb2287 and verified that this new code was triggered.

box64-bundle-x86-libs.sh
```diff
@@ -101,6 +101,13 @@ for lib in libc.so libpthread.so librt.so libGL.so libGL.so libX11.so \
     do find "${dir_tmp_local}" -name "${lib}"* -delete
 done
 
+if find "${dir_tmp_local}"/bundle-libs/ -type l ! -exec test -e {} \; -print | grep bundle-libs
+    then echo "Broken symlinks found."
+    cd -
+    echo rm -r -f "${dir_tmp_local}"
```
ptitSeb, line 107: I am not comfortable with this. why testing the presence of `"${dir_tmp_local}"/bundle-libs/` to then delete parent `"${dir_tmp_local}"`
What happens if `"${dir_tmp_local}"` is nothing or `/` for some reason? or even `/tmp`? That foldr should not be deleted.

LukeShortCloud: I already do the same kind of deletion at the very end:

https://github.com/ptitSeb/box64/blob/v0.3.6/box64-bundle-x86-libs.sh#L96

This test lists out every file that has a broken symlink. We're basically aborting early if there are.

We set the `${dir_tmp_local}` variable here:

https://github.com/ptitSeb/box64/blob/v0.3.6/box64-bundle-x86-libs.sh#L35

I suppose there could be an edge-case where `mktemp` is not available (although, the supported platforms for running this script - Arch Linux, Debian, and Fedora - all have this installed by default). When testing this PR, I did test with an empty `${dir_tmp_local}` and `rm -r -f` did nothing.

ptitSeb: fair enough.

PR #2836, opened 2025-07-21
[DYNACACHE]LA64] Enabled dynacache for LA64

ksco:


ptitSeb:
Isn't defining `ARCH_VERSION` in `env.c` missing to enable dynacache?

ksco:
sorry, forgot to git add that file

PR #2839, opened 2025-07-21
Box64 bundle x86 libs improvements

fboudra:


fboudra:
@LukeShortCloud this is the rewrite version. Though I have another branch less intrusive if you don't like the approach. Let me know.

LukeShortCloud:
@fboudra Thanks for the PR! I'm always glad to see more folks interested in the bundle. I'll give it a thorough review later today. At a glance, it looks good.

box64-bundle-x86-libs.csv
```diff
@@ -1,8 +1,32 @@
-https://vault.almalinux.org/9.5/BaseOS/x86_64/os/Packages/bluez-libs-5.72-2.el9.i686.rpm,af1aba8e6fa5ccf33032d58a3acc59368c07d532293396db205fb9a4bc1241ae
-https://vault.almalinux.org/9.5/BaseOS/x86_64/os/Packages/bluez-libs-5.72-2.el9.x86_64.rpm,49c273813b4616240acdae8dae893192658c7541482f9bbb2c46dfe19b61877d
+https://archive.debian.org/debian/pool/main/c/curl/libcurl3-gnutls_7.74.0-1.3+deb11u7~bpo11+1_amd64.deb,d5ae55e3a9838d12358530cd03218cc0d18127a84e9cfc73bdf79dacd40b25e4
```
LukeShortCloud, line 1: Was `sort` used on this file? I was organizing the rows alphabetically by package name. I'm open to this be restructured differently as long as it's something meaningful (ex., all RPMs, all packages from Debian, etc.).

box64-bundle-x86-libs.sh
```diff
@@ -1,121 +1,123 @@
-#!/bin/bash
+#!/bin/sh
```
LukeShortCloud, line 1: Is this script actually portable and working on other shells? I used to have a large project written in Zsh and there were lots of issues with compatibility between Bash that I could not solve. In this case, we can guarantee compatibility with Bash, at least.

box64-bundle-x86-libs.sh
```diff
@@ -1,121 +1,123 @@
...
 
-set -x
+set -ex
```
LukeShortCloud, line 3: Nice addition!

box64-bundle-x86-libs.sh
```diff
@@ -1,121 +1,123 @@
...
-    # Example x86_64 package name: libjpeg-turbo-3.0.1_1.x86_64.xbps
-    # Example x86 package name: libjpeg-turbo-32bit-3.0.1_1.x86_64.xbps
-    elif echo "${1}" | grep -q -P ".xbps$"; then
-        echo "XBPS (Void Linux) package detected."
-        tar --extract --file "$(basename "${1}")"
-    else
-        echo "Unsupported package type detected."
-        exit 1
-    fi
+trap cleanup_exit INT TERM EXIT
```
LukeShortCloud, line 5: I almost did this in one of my last PRs. Thanks for adding the `trap`!

box64-bundle-x86-libs.sh
```diff
@@ -1,121 +1,123 @@
...
+cleanup_exit()
+{
+    echo "Running cleanup_exit..."
+    rm -rf /tmp/box64-bundle.*
 }
 
-box64_dir=$(pwd)
-dir_tmp_local="$(mktemp --directory)"
-cd "${dir_tmp_local}"
+# we must have few tools
```
LukeShortCloud, line 13: I was going to open a PR for similar checks this week. You beat me to it. ;-)

box64-bundle-x86-libs.sh
```diff
@@ -1,121 +1,123 @@
...
+awk=$(which awk) || { echo "E: You must have awk" && exit 1; }
+curl=$(which curl) || { echo "E: You must have curl" && exit 1; }
+curl_cmd="${curl} --connect-timeout 5 --retry 5 --retry-delay 1 --create-dirs -fsSLo"
+rpm2cpio=$(which rpm2cpio) || { echo "E: You must have rpm2cpio" && exit 1; }
```
LukeShortCloud, line 18: As strange as it may sound, we also need a test for `cpio` itself. On the AlmaLinux 9 minimal container image, `rpm2cpio` is installed but `cpio` is not. I assume that it is part of some embedded library for DNF and/or RPM which is why the CLI tool is not installed.

box64-bundle-x86-libs.sh
```diff
@@ -1,121 +1,123 @@
...
+for file in "$@"; do
+    rm -f "${dir_tmp}"/bundle-libs/usr/lib/box64-*-linux-gnu/"${file}".so*
 done
 
-if find "${dir_tmp_local}"/bundle-libs/ -type l ! -exec test -e {} \; -print | grep bundle-libs
-    then echo "Broken symlinks found."
-    cd -
-    echo rm -r -f "${dir_tmp_local}"
+# check broken symlinks
+if [ -n "$(find "${dir_tmp}"/bundle-libs -type l ! -exec test -e {} \; -print)" ]; then
```
LukeShortCloud: The problem with this is that the `$()` hides the output so it's not clear what symlinks are broken.

box64-bundle-x86-libs.sh
```diff
@@ -1,121 +1,123 @@
...
-mv "${dir_tmp_local}"/*.deb "${dir_tmp_local}/bundle-pkgs/"
-mv "${dir_tmp_local}"/*.eopkg "${dir_tmp_local}/bundle-pkgs/"
-mv "${dir_tmp_local}"/*.rpm "${dir_tmp_local}/bundle-pkgs/"
-mv "${dir_tmp_local}"/*.xbps "${dir_tmp_local}/bundle-pkgs/"
-
-tar --directory "${dir_tmp_local}/bundle-libs" --create --file "${box64_dir}/box64-bundle-x86-libs.tar.gz" .
-tar --directory "${dir_tmp_local}/bundle-pkgs" --create --file "${box64_dir}/box64-bundle-x86-pkgs.tar.gz" .
+# create shared libraries (.so) symlinks
+# if multiple libraries of the same name exist, it will symlink the oldest version
+# e.g. libcurl-gnutls
```
LukeShortCloud: It would be nice to have an example to better showcase what is happening here. Something like:

```
# libcurl-gnutls.so -> libcurl-gnutls.so.3
```

LukeShortCloud (changes requested):
Really great overall. I appreciate the optimizations to memory, speed, error checking, and also ShellCheck now passing!

When I first wrote this, it was just a proof-of-concept. Thanks for giving it that extra polish. I've been too focused on finding libraries versus cleaning up this script.

I left some small questions and comments.

box64-bundle-x86-libs.csv
fboudra: yes, I'm using the following commands:
`LC_ALL=C sort -u filename | sponge filename`

box64-bundle-x86-libs.sh
fboudra: yes, actually I'm a zsh user. Beside, I tested with dash which I have on my system as well. So tested successfully with bash, dash and zsh.

box64-bundle-x86-libs.sh
fboudra: I had the same in mind. We can call it twice:
```
if [ -n "$(find "${dir_tmp}"/bundle-libs -type l ! -exec test -e {} \; -print)" ]; then
    echo "E: Broken symlinks found"
    # display the output
    find "${dir_tmp}"/bundle-libs -type l ! -exec test -e {} \; -print
    exit 1
fi
```

box64-bundle-x86-libs.sh
fboudra: ack, I'll update the comment.

box64-bundle-x86-libs.sh
fboudra: ack, I'll  fix

fboudra: done

box64-bundle-x86-libs.sh
fboudra: done

box64-bundle-x86-libs.sh
LukeShortCloud: The original code I had isn't the prettiest

```
if find "${dir_tmp_local}"/bundle-libs/ -type l ! -exec test -e {} \; -print | grep bundle-libs
```

but running the `find` command twice seems unnecessary.

LukeShortCloud (changes requested):
I'd don't love the `sort` on the CSV file but at least the URLs are generally consistent with what OS it comes from.

Before the final sign-off, I'd like for the "check broken symlinks" section to be reverted back. Otherwise you addressed all my comments so thank you for that!

LukeShortCloud (approved):
@fboudra Thanks for the updates. Approved! :-)

@ptitSeb FYI, if you wanted to merge now.

PR #2864, opened 2025-07-31
[WRAPPER] Wrapped some functions for gmodule2

ksco:
I have double checked the signature.


ptitSeb:
This probably needs better wrapping latter, to allow x86_64 module to be loaded with native version of gmodule...

PR #2870, opened 2025-08-01
[WRAPPED] Wrapped more functions for libc

ksco:


ptitSeb:
Static builds are broken. You also need to defines those function in `static_libc.c`

PR #2890, opened 2025-08-04
[LA64_DYNAREC] Added and optimized more fastround=0 cases

ksco:


src/dynarec/la64/dynarec_la64_660f.c
```diff
@@ -2410,8 +2410,20 @@ uintptr_t dynarec64_660F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
             GETGX_empty(v0);
-            // TODO: fastround
-            VFTINTRZ_W_D(v0, v1, v1);
+            if (!BOX64ENV(dynarec_fastround)) {
+                d0 = fpu_get_scratch(dyn);
+                q0 = fpu_get_scratch(dyn);
+                q1 = fpu_get_scratch(dyn);
+                VFTINTRZ_W_D(d0, v1, v1);
+                VLDI(q0, 0b1001110000000); // broadcast 32bit 0x80000000 to all
+                LU52I_D(x5, xZR, 0x41e);
```
ptitSeb, line 2419: There is no "invalid operation" flags anywhere? you need to detect NaN or something like that instead?

ksco: Well, at least no "per-element" flags. This is just the way we do this all over the place. However, I noticed this implementation is not 100% exact, need to come up with something better later.

cc @phorcys

ptitSeb: > Well, at least no "per-element" flags. This is just the way we do this all over the place. However, I noticed this implementation is not 100% exact, need to come up with something better later.
> 
> cc @phorcys

On ARM, there is no "per element" flag either, so each element is handled separatly in a (unrolled) loop (unless FRINTTS extensions is present were there is an opcode that does the x86 rounding special cases directly)

ptitSeb: Merging this one, as the proprer handling might come later in a different PR.

ksco: Yeah, same on RISC-V: we fallback to scalar version for fastround=0. That might be the solution for loongarch too, but until then, it should be fine to keep the current implementation.

phorcys: I'll check those CVT insts ag carefullly next week.
It's ok to check  V fp exception , unroll loop, and insert 0x8000000 if needed.
But current imp "should" be a little faster.

ptitSeb: well, the idea of fastround=0 is to be exact first, fast second. this method is faster yes, but I don't think it's exact.

phorcys: > well, the idea of fastround=0 is to be exact first, fast second. this method is faster yes, but I don't think it's exact.

It's intend to be exact. 
it begins from these tables:

Compare SSE's cvtps2pi with LoongArch's ftint.w.s/vftint.w.s
|   | cvtps2pi  | vftint.w.s |
| ------------- | ------------- | ------------- |
|normal|--|--|
|+/- nan|0x80000000 raise V|0x0  raise V|
|+overflow|0x80000000 raises V|0x7FFFFFFF raises V|
|-overflow|0x80000000 raises V|0x80000000 raises V|
|-underflow|0x0 raises I|0x0 raises I|
|+underflow|0x0 raises I|0x0 raises I|

only +/-nan  &  +overflow result is different.
lucky that _Float32 is not accurate:
| float value | hex value  | type |
| ------------- | ------------- | ------------- |
|-2147483648.0f | 0xcf000000 |(_Float32)|
|-2147483520.0f | 0xceffffff |(_Float32)|
|2147483648.0f | 0x4f000000 |(_Float32)|
|2147483520.0f | 0x4effffff |(_Float32)|

ptitSeb:
It's the same on the 3 arch: nan & +overflow are just wrong. you are using a methog similar to interp I guess? `if NAN || val>=0x8000000 => value is 0x80000000`

phorcys:
> It's the same on the 3 arch: nan & +overflow are just wrong. you are using a methog similar to interp I guess? `if NAN || val>=0x8000000 => value is 0x80000000`

Yes,  use FCMP/VFCMP cond  cULE/cULT  can do NAN || val>=0x8000000 check in one inst.

As for  Double <-> Integer convert, no single value because of roudning.
But could use an additonal  int64 convert with rouding,  then using same method.

ptitSeb:
Mmmm, actually, I might see if I canimplement this way on arm64 too, it's probablly much more efficient, espcialy for float and avx 256bits rounding.

ksco:
It’s not exact. I saw some failures on -inf cases, but haven’t looked into that.

PR #2892, opened 2025-08-04
[LA64_DYNAREC] Fixed F3 0F 53 RCPSS opcode

ksco:
x86 spec indeed states that it's "approximate reciprocal", but it seems that we can't use the "approximate reciprocal" version in LoongArch to emulate it. I don't know why, tho.

cc @phorcys


ptitSeb:
Well, intel doc state that `|Relative Error| ≤ 1.5 ∗ 2^−12` so it might be that native approximate is not precise enough...

ptitSeb:
Should I merge it or should I wait?

ksco:
wait for what?

ksco:
oh for figuring out why? well, it’s undocumented so nowhere to look into…

ptitSeb:
ok, no wait :)

PR #2901, opened 2025-08-05
Add O(1) early-out to eusing cached bounds in `rb_get()` and `rb_get_64()`

devarajabc:
Add a fast-path in `rb_get()` and `rb_get_64()` that checks the tree’s cached `leftmost->start` and `rightmost->end` bounds before calling `find_addr()`. If the query address lies outside the address bounds, we return 0 immediately, avoiding an unnecessary tree walk.

This mirrors the idea used in [jserv/rbtree](https://github.com/jserv/rbtree/blob/d3e6d6d6b6b1d297642606a5e199108ba7fc19f0/rbtree.c#L881)’s `rb_cached_contains()`, using cached min/max to short-circuit lookups.

Behavior is unchanged (both functions still return 0 when not found); only negative lookups become cheaper.

### DIFF
--- a/src/tools/rbtree.c
+++ b/src/tools/rbtree.c
@@ -569,15 +569,19 @@ static rbnode *succ_node(rbnode *node) {
 }
 
 uint32_t rb_get(rbtree_t *tree, uintptr_t addr) {
+    if (tree->leftmost && addr < tree->leftmost->start) return 0;
+    if (tree->rightmost && addr > tree->rightmost->end) return 0;
     rbnode *node = find_addr(tree, addr);
     if (node) return node->data;
-    else return 0;
+    return 0;
 }
 
 uint64_t rb_get_64(rbtree_t *tree, uintptr_t addr) {
+    if (tree->leftmost && addr < tree->leftmost->start) return 0;
+    if (tree->rightmost && addr > tree->rightmost->end) return 0;
     rbnode *node = find_addr(tree, addr);
     if (node) return node->data;
-    else return 0;
+    return 0;
 }
 
 int rb_get_end(rbtree_t* tree, uintptr_t addr, uint32_t* val, uintptr_t* end) {


src/tools/rbtree.c
```diff
@@ -604,6 +608,7 @@ int rb_get_end(rbtree_t* tree, uintptr_t addr, uint32_t* val, uintptr_t* end) {
     return 0;
 }
 
+
```
jserv: Minimize the required changes. Don't add newlines unless necessary.

src/tools/rbtree.c
```diff
@@ -569,12 +569,16 @@ static rbnode *succ_node(rbnode *node) {
 }
 
 uint32_t rb_get(rbtree_t *tree, uintptr_t addr) {
+    if(tree->leftmost && addr < tree->leftmost->start) return 0;
+    if(tree->rightmost && addr > tree->rightmost->end) return 0;
     rbnode *node = find_addr(tree, addr);
     if (node) return node->data;
     else return 0;
```
jserv: The 'else' is not necessary.

src/tools/rbtree.c
```diff
@@ -569,12 +569,16 @@ static rbnode *succ_node(rbnode *node) {
 }
 
 uint32_t rb_get(rbtree_t *tree, uintptr_t addr) {
+    if(tree->leftmost && addr < tree->leftmost->start) return 0;
+    if(tree->rightmost && addr > tree->rightmost->end) return 0;
```
jserv: Be aware of the consistent code style.

src/tools/rbtree.c
rajdakin: It may be unnecessary, but it is a logical signal: the structure of the code is "if a node is found, do this, otherwise do that" (where "do" is actually "return"). It allows one to easily see at the first glance that the function will not always return `0`, and is similar to other functions that have a similar pattern. (Also, notice that it was not added by him, since there is no `+` on the leftmost column and it is not on a green background.)

Notice that without the `else`, the structure is:
```c
if (node) do_something;
return 0;
```
which is misleading at first glance.

jserv (approved):

PR #2930, opened 2025-08-12
[DYNAREC] Rearranged arch-specific AVX infra code

ksco:
Based on the current status and plans, the AVX infrastructure originally designed for Arm (with the expectation of compatibility with RV64 and LA64) will no longer be applicable. Maintaining shared code in this situation would likely create complications.

This PR moves all the "common code" to Arm DynaRec and removes the existing AVX infrastructure code from RV64 to provide a clean slate, as the development plan has changed as follows:

After this PR is merged, I'll begin implementing AVX support for RISC-V. The implementation will follow this sequence:

1. scalar version
2. vlen=128 version (both xtheadvector and v1.0)
3. vlen>=256 version

### DIFF
--- a/src/dynarec/arm64/dynarec_arm64_functions.c
+++ b/src/dynarec/arm64/dynarec_arm64_functions.c
@@ -1148,6 +1148,7 @@ void updateNativeFlags(dynarec_native_t* dyn)
 void rasNativeState(dynarec_arm_t* dyn, int ninst)
 {
     dyn->insts[ninst].nat_flags_op = dyn->insts[ninst].set_nat_flags = dyn->insts[ninst].use_nat_flags = dyn->insts[ninst].need_nat_flags = 0;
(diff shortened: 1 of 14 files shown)


phorcys:
Is making  x64emu_t's xmm/ymm into single 256bits array for la64 lasx/ rv64 vlen>=256 /arm64 SVE on roadmap?
Currently xmm[16]/ymm[16] memory layout causing avx2 256bits load/store performance degration.

ksco:
I think it’s a good idea, with the cost of another version of interpreter avx implementation or some state conversion between interp and dynarec. We’ll see ;)

xiangzhai:
> vlen>=256 version
> arm64 SVE

My Mac mini M2 only supported Neon :( Perhaps there are not many hardwares supported SVE? So test the code on qemu?

Thanks,
Leslie Zhai

phorcys:
> > vlen>=256 version
> > arm64 SVE
> 
> My Mac mini M2 only supported Neon :( Perhaps there are not many hardwares supported SVE? So test the code on qemu?
> 
> Thanks, Leslie Zhai

Snapdragon X Elite（win on arm version)
Apple M4
Those 2 have sve support 
New Snapdragon 8 Elite 2 coming in September , have sve support ( from leaked geekbench info)

PR #2943, opened 2025-08-15
[LA64_DYNAREC] Added more scalar AVX opcodes

ksco:


ksco:
Oops, it’s for rv64, not la64….

ptitSeb:
Ah yeah, so much AVX on rv64 was a bit strange...

PR #2954, opened 2025-08-21
[RV64_DYNAREC] Added more scalar avx 66 0F38 opcodes

ksco:


ptitSeb:
Some AVX and SSE 32bits tests are failing on RISC-V.

ksco:
Oops, it's the non-zbb path, fixed.

PR #2958, opened 2025-08-21
[CORE] Better trace_file logging

ksco:


src/core.c
```diff
@@ -1050,7 +1050,6 @@ int initialize(int argc, const char **argv, char** env, x64emu_t** emulator, elf
         ApplyEnvFileEntry(box64_wine_guest_name);
         box64_wine_guest_name = NULL;
     }
-    openFTrace(0);
```
ptitSeb, line 1053: is it still working when using fork?

ksco: Actually this line is here to open trace file again after applying the rcfile. But I still don't like the redundant "Trace redirected to ...".

ptitSeb: > Actually this line is here to open trace file again after applying the rcfile. But I still don't like the redundant "Trace redirected to ...".

Well, I don't like this repetation either. but  is the rcfile trace_file naming still working?

ksco: No, it stopped working. I'll come up with something better.

ksco:
Refined `openFTrace` to remove the repetition, also removed some code that has not been used for a long time.

PR #2966, opened 2025-08-22
[RV64_DYNAREC] Fixed scalar avx VCMPPD non-ymm case

ksco:


src/dynarec/rv64/dynarec_rv64_avx_66_0f.c
```diff
@@ -812,6 +812,10 @@ uintptr_t dynarec64_AVX_66_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip,
                     SD(x3, gback, gyoffset + 8 * (i - 2));
                 }
             }
+            if (!vex.l) {
+                SD(xZR, gback, gyoffset + 0);
+                SD(xZR, gback, gyoffset + 8);
+            }
```
ptitSeb, line 818: Oh, there is no ymm0 handling on rv64/scalar :O ?!!!

ksco: Yeah, I didn’t do that. Probably will do it later (or never), it’s not that important for a scalar implementation in my opinion.

ptitSeb: Well, it spare many useless 16 bytes writes, and it wil also be usefull for 128bits only vectors hardware... And it's pretty cheap.

PR #2973, opened 2025-08-26
[RV64_DYNAREC] Added more scalar avx opcodes

ksco:


ptitSeb:
Something is wrong, RiscV tests are red.

Issue #2986, opened 2025-08-31
compilation error on ubuntu bionic gcc-8


theofficialgman:
```
 [ 69%] Building C object CMakeFiles/mainobj.dir/src/wrapped/wrappedlibcrypt.c.o
/github/workspace/box64/src/wrapped/wrappedlibc.c: In function 'my_mallinfo2':
/github/workspace/box64/src/wrapped/wrappedlibc.c:3203:9: error: dereferencing pointer to incomplete type 'struct mallinfo2'
         *(struct mallinfo2*)p=f();
         ^~~~~~~~~~~~~~~~~~~~~
/github/workspace/box64/src/wrapped/wrappedlibc.c:3203:31: error: invalid use of undefined type 'struct mallinfo2'
         *(struct mallinfo2*)p=f();
                               ^
/github/workspace/box64/src/wrapped/wrappedlibc.c:3205:29: error: invalid application of 'sizeof' to incomplete type 'struct mallinfo2'
         memset(p, 0, sizeof(struct mallinfo2));
                             ^~~~~~
CMakeFiles/mainobj.dir/build.make:2036: recipe for target 'CMakeFiles/mainobj.dir/src/wrapped/wrappedlibc.c.o' failed
make[2]: *** [CMakeFiles/mainobj.dir/src/wrapped/wrappedlibc.c.o] Error 1
```

https://github.com/Pi-Apps-Coders/box64-debs/actions/runs/17353481120

theofficialgman:
closed with https://github.com/ptitSeb/box64/pull/2987

Issue #2484, opened 2025-03-31
-DRPI3ARM64=1 builds now fail


theofficialgman:
-DRPI3ARM64=1 builds now fail. other build options are not failing.

https://github.com/Pi-Apps-Coders/box64-debs/actions/runs/14165361624/job/39677513972#step:5:14788

```
 [  5%] Building ASM object CMakeFiles/dynarec_native.dir/src/dynarec/arm64/arm64_lock.S.o
/github/workspace/box64/src/dynarec/arm64/arm64_lock.S: Assembler messages:
/github/workspace/box64/src/dynarec/arm64/arm64_lock.S:124: Error: selected processor does not support `swpal x1,x0,[x0]'
/github/workspace/box64/src/dynarec/arm64/arm64_lock.S:144: Error: selected processor does not support `swpal w1,w0,[x0]'
/github/workspace/box64/src/dynarec/arm64/arm64_lock.S:164: Error: selected processor does not support `swpalh w1,w0,[x0]'
/github/workspace/box64/src/dynarec/arm64/arm64_lock.S:184: Error: selected processor does not support `swpalb w1,w0,[x0]'
/github/workspace/box64/src/dynarec/arm64/arm64_lock.S:207: Error: selected processor does not support `casal x2,x1,[x0]'
/github/workspace/box64/src/dynarec/arm64/arm64_lock.S:233: Error: selected processor does not support `casal w2,w1,[x0]'
/github/workspace/box64/src/dynarec/arm64/arm64_lock.S:260: Error: selected processor does not support `casal x2,x1,[x0]'
/github/workspace/box64/src/dynarec/arm64/arm64_lock.S:291: Error: selected processor does not support `casal w2,w1,[x0]'
/github/workspace/box64/src/dynarec/arm64/arm64_lock.S:319: Error: selected processor does not support `casal w2,w1,[x0]'
/github/workspace/box64/src/dynarec/arm64/arm64_lock.S:373: Error: selected processor does not support `swpal w1,wzr,[x0]'
make[2]: *** [CMakeFiles/dynarec_native.dir/src/dynarec/arm64/arm64_lock.S.o] Error 1
CMakeFiles/dynarec_native.dir/build.make:239: recipe for target 'CMakeFiles/dynarec_native.dir/src/dynarec/arm64/arm64_lock.S.o' failed
make[1]: *** [CMakeFiles/dynarec_native.dir/all] Error 2
[...]

Issue #2425, opened 2025-03-06
Build error


theofficialgman:
```
 [100%] Linking C executable box64
CMakeFiles/box64.dir/src/wrapped/wrappedlibc.c.o: In function `my_epoll_pwait2':
/github/workspace/box64/src/wrapped/wrappedlibc.c:2216: undefined reference to `epoll_pwait2'
collect2: error: ld returned 1 exit status
make[2]: *** [box64] Error 1
```

https://github.com/Pi-Apps-Coders/box64-debs/actions/runs/13669787081/job/38217680899#step:5:5212

ptitSeb:
Should be fixed now.

Issue #2190, opened 2024-12-23
Compilation failure `github/workspace/box64/src/custommem.c:2215:34: error: 'MAP_FIXED_NOREPLACE' undeclared (first use in this function)`


theofficialgman:
`github/workspace/box64/src/custommem.c:2215:34: error: 'MAP_FIXED_NOREPLACE' undeclared (first use in this function)`

https://github.com/Pi-Apps-Coders/box64-debs/actions/runs/12451874246/job/34760245889#step:5:2838

PR #2987, opened 2025-08-31
[CI] Added ARM64-GCC-8 target (Ubuntu Bionic)

theofficialgman:
uses taiki-e/checkout-action@v1 instead of actions/checkout@v4 due to ubuntu bionic container not having new enough nodejs or glibc to run most github actions that use javascript

does not build build Trace target due to ubuntu bionic not having the libzydis-dev package

adds cmake ppa for a newer cmake version that box64 requires

uses gcc-8 specifically (instead of bionic's default gcc-7) due to  historically better compatibility with box64 (and it has been used for years now by installation scripts and other ci such as https://github.com/Pi-Apps-Coders/box64-debs)

This is being added primarily because we use bionic with gcc-8 for platform compatibility at https://github.com/Pi-Apps-Coders/box64-debs (at least for some targets) and occasionally changes at box64 cause builds to fail and don't get caught until we reach out later (current issue https://github.com/ptitSeb/box64/issues/2986 , historical issues https://github.com/ptitSeb/box64/issues/2484 https://github.com/ptitSeb/box64/issues/2425 https://github.com/ptitSeb/box64/issues/2190 https://github.com/ptitSeb/box64/issues/1986 https://github.com/ptitSeb/box64/issues/1586 https://github.com/ptitSeb/box64/issues/1033 https://github.com/ptitSeb/box64/issues/342 https://github.com/ptitSeb/box64/issues/278 https://github.com/ptitSeb/box64/issues/249) . This CI adds building with bionic gcc-8 in a non-invasive way similar to other targets.

The CI fails due to the above linked issue but should pass once that is fixed.

I will add a similar PR to box86 for similar reasons once this is merged.


ksco:
Oh, it's because older glibc does not have `mallinfo2`. This patch should fix the issue:

```diff
diff --git a/src/wrapped/wrappedlibc.c b/src/wrapped/wrappedlibc.c
index 742ae677..967d867f 100644
--- a/src/wrapped/wrappedlibc.c
+++ b/src/wrapped/wrappedlibc.c
@@ -3190,7 +3190,20 @@ EXPORT void* my_mallinfo(x64emu_t* emu, void* p)
     return p;
 }
 
-typedef struct mallinfo2 (*mallinfo2_fnc)(void);
+struct my_mallinfo2_s {
+    size_t arena;
+    size_t ordblks;
+    size_t smblks;
+    size_t hblks;
+    size_t hblkhd;
+    size_t usmblks;
+    size_t fsmblks;
+    size_t uordblks;
+    size_t fordblks;
+    size_t keepcost;
+};
+
+typedef struct my_mallinfo2_s (*mallinfo2_fnc)(void);
 EXPORT void* my_mallinfo2(x64emu_t* emu, void* p)
 {
     static mallinfo2_fnc f = NULL;
@@ -3200,9 +3213,9 @@ EXPORT void* my_mallinfo2(x64emu_t* emu, void* p)
         f = (mallinfo2_fnc)dlsym(my_lib->w.lib, "mallinfo2");
     }
     if(f)
-        *(struct mallinfo2*)p=f();
+        *(struct my_mallinfo2_s*)p = f();
     else
-        memset(p, 0, sizeof(struct mallinfo2));
+        memset(p, 0, sizeof(struct my_mallinfo2_s));
     return p;
 }
```

Feel free to push this, or you can enable the "Allow maintainers to edit" option on the bottom-right corner of this webpage, so I can push the fix to this PR directly if you're lazy. ;)

theofficialgman:
oh sorry about that @ksco go ahead and push if you have it handy

PR #2998, opened 2025-09-02
[BASH] Use custom PS1 prefix to make box64-bash more distinguishable

ksco:


src/core.c
```diff
@@ -1031,7 +1034,17 @@ int initialize(int argc, const char **argv, char** env, x64emu_t** emulator, elf
     for(int i=1; i<my_context->argc; ++i) {
         my_context->argv[i] = box_strdup(argv[i+nextarg]);
         printf_log(LOG_INFO, "argv[%i]=\"%s\"\n", i, my_context->argv[i]);
+
+        if (setup_bash_rcfile && (!strcmp(my_context->argv[i], "--norc") || !strcmp(my_context->argv[i], "--rcfile") || !strcmp(my_context->argv[i], "--init-file"))) {
```
ptitSeb: I think you should check that there is no arguments in the command line for bash before inserting the custom ones, else that might break some scipts.

PR #2999, opened 2025-09-03
Fix undefined type ‘struct mallinfo2’ for LA64 ABI 1.0

xiangzhai:
Hi,

glibc >= 2.33 [Add mallinfo2 function that support sizes >= 4GB](https://sourceware.org/pipermail/glibc-cvs/2020q3/070629.html), but LA64 ABI 1.0 glibc is 2.28, so `mallinfo2` is not available.

Please review my patch.

Thanks,
Leslie Zhai


ksco:
This issue is being addressed here: https://github.com/ptitSeb/box64/pull/2987#issuecomment-3240949384

PR #3010, opened 2025-09-09
[RV64_DYNAREC] Added more scalar avx opcodes

ksco:


src/dynarec/rv64/dynarec_rv64_avx_f2_0f.c
```diff
@@ -78,6 +78,194 @@ uintptr_t dynarec64_AVX_F2_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip,
                 YMM0(ed);
             }
             break;
+        case 0x58:
+            INST_NAME("VADDSD Gx, Vx, Ex");
+            nextop = F8;
+            GETEX(x1, 0, 1);
```
ptitSeb, line 84: Why not using GETEXSD and friends?

ksco: I’m worried about the infra is not ready yet.

ptitSeb: but that's used on SSE opcodes right? Well the fact that there is 3 regs can make things a bit complicated of courses, but the speed increase for not reading / writting the double must be important.

ksco: Agree. Still I’m not sure if the infra of those macros is ready for avx (didn’t check at all), so I tends to leave it for later infra work if that makes sense sense.

ptitSeb: Ok. For later then.

Issue #3011, opened 2025-09-09
Emulator running WeChat Mini-Program in Arm64 is Ok


Alion-King:
Hello, ptitSeb
&emsp; &emsp; Now, By modifying Box64(commit:6f144aa429e67721582bd8f02170ee90fe3fafa9, date:20250908) , **WeChat MiniProgram  built for the x86_64 architecture** can be run on the Raspberry Pi 5.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/694b2bc3-9057-44cf-8ca2-63e14717da95" />

      
---
> BACKGROUND:   WeChat is an instant-messaging application with hundreds of millions of users worldwide.  Mini-Program is a built-in feature of WeChat, like browser plug-ins.  At present, Box64 can run the main process just for sending/receiving messages, but it cannot yet run Mini-Program.

&emsp; &emsp;Mini-Program is built on a Web-technology stack, which poses a new challenge for Box64  

&emsp; &emsp;Initially, our debugging efforts were blocked by a cryptic error:  `cfree(): invalid pointer`.

&emsp; &emsp;Further investigation revealed the root cause: Mini-Program ship  own `malloc`/`free` function family to manage memory. Consequently, directly mapping these functions to the host libc by the library-wrapping mechanism, causes the error. 

&emsp; &emsp;Fortunately,  Box64 provides the **malloc-hook** mechanism, which can handle the problem.  

&emsp; &emsp;However, under the special execution environment of Mini-Program,  we have uncovered several issues listed below, and designed a solution:

1. **Activation strategy of malloc-Hook for the x86_64 ELF**  
2. **Memory management** —the malloc-hoock mechainism relies on memory attributes
4. **Translated-code-blocks management**  
&emsp; Here we share our current achievements.  Hope to exchange ideas with you furtherly, making Box64 more prefect. 
Thanks.
Aqua.

ptitSeb:
So, setting `BOX64_MALLOC_HACK=2` for WeChat or sub-program is not enough?

Alion-King:
setting BOX64_MALLOC_HACK=2 for Mini-Program is not enough. 

except  setting BOX64_MALLOC_HACK=2 , I modify the source code.  This is the patch for commit: 6f144aa429e67721582bd8f02170ee90fe3fafa9
[box64_wechat-miniProgram.patch](https://github.com/user-attachments/files/22231777/box64_wechat-miniProgram.patch)

I understand:
[...]

PR #3013, opened 2025-09-11
Correct address check internalDBGetBlock

Aqua-Peak:
In internalDBGetBlock, input paramterss has "addr" and "filladdr". hasAlternate should check x64pc named addr, not hostpc named filladdr.
relate: "Emulator running WeChat Mini-Program in Arm64 is Ok #3011"

### DIFF
--- a/src/dynarec/dynablock.c
+++ b/src/dynarec/dynablock.c
@@ -227,7 +227,7 @@ void cancelFillBlock()
 */
 static dynablock_t* internalDBGetBlock(x64emu_t* emu, uintptr_t addr, uintptr_t filladdr, int create, int need_lock, int is32bits, int is_new)
 {
-    if (hasAlternate((void*)filladdr))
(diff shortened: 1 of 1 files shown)


ksco:
Well, those two addresses are for the alternate stuff; they have nothing to do with "hostpc" or "x64pc". Also, how is this related to #3011, does it fix something?

Alion-King:
"addr" is the x64pc of malloc/free funciton in x86_64 elf file.  "filladdr" is the hostpc of generated block by Box64. 
 the hash table in hasAlternate records the x64pc as key.

ptitSeb:
Mmmm, with this patch, you will disable dynarec for everything that has an alternate. I don't think that's the correct way to fix your issue.

ksco:
Closed for a long period of inactivity, feel free to open a new one if necessary.

PR #3014, opened 2025-09-11
Update x64syscall.c, add syscall:425

Aqua-Peak:
add syscall:425 in x64syscall. relate: Issue 3011


ptitSeb:
If you can add the syscall to the big sparse array on top, you don't need to add a `case` for it If you need to transform some of the parameters, then you need to comment the line in the array.

ksco:
Closed for a long period of inactivity, feel free to open a new one if necessary.

PR #3035, opened 2025-09-27
Explicitly link standard libs through CMake

definitelyuncertain:
I initially got undefined reference to `round` and other standard library functions at the linking step for `box64`. From closer inspection, I noticed in the build output (`VERBOSE=1`) that the linking step was adding the libraries _before_ the object files like so:

```bash
riscv64-unknown-linux-gnu-gcc --sysroot=path/to/sysroot -O3 -DNDEBUG -Wl,--export-dynamic -rdynamic -Wl,--no-as-needed
-lc -lm -ldl -lrt -lpthread -lresolv -lutil
-Wl,--as-needed -pthread  -Wl,-Ttext-segment,0x34800000 -Wl,--dependency-file=CMakeFiles/box64.dir/link.d CMakeFiles/mainobj.dir/src/os/backtrace.c.o CMakeFiles/mainobj.dir/src/os/emit_signals_linux.c.o CMakeFiles/mainobj.dir/src/os/hostext_common.c.o CMakeFiles/mainobj.dir/src/os/hostext_linux.c.o CMakeFiles/mainobj.dir/src/os/freq_linux.c.o CMakeFiles/mainobj.dir/src/os/os_linux.c.o
.
.
.
```

Explicitly linking the standard libraries through CMake (`target_link_libraries`) fixed it.
Presumably this is also a more portable solution.

Building for RV64 on Manjaro with a custom toolchain.


ptitSeb:
There are some Red build on the CI

definitelyuncertain:
I've restored the original link flags also, maybe that will work for the failed builds too?

ptitSeb:
Yeah, CI is happy and green now.

PR #3041, opened 2025-10-04
[DYNAREC] Optimized scalar AES impl in DynaRec

xctan:


ptitSeb:
Wow, lookup tables and hand-optimized assembly routines.. fancy!

ptitSeb:
AES tests are red, something is wrong.

xctan:
My current RVV impl isn't compatible with xtheadvector. I'll update a new version later.

ptitSeb:
So the whole assembly functins are disabled for now?

xctan:
While the xtheadvector detection works in my local xuantie qemu and sg2042 boards, I'm unsure why it failed in the CI. The assembly functions works well with RVV 1.0 and also passes RVV 1.0 CI, but produces erroneous results with xtheadvector. The assembly version only provides ~1.4x speedup compared to the new scalar version on sg2044. Should I move the assembly part to another future PR?

ptitSeb:
As you wish. But if you leave the files in this PR, make sure you bring a fix at some point.

PR #3076, opened 2025-10-17
[ENV] Added a new BOX64_ARCH option

ksco:


src/tools/env.c
```diff
@@ -612,12 +646,12 @@ void LoadEnvVariables()
...
     }
-#define STRING(NAME, name, wine)             \
-    p = GETENV(#NAME, wine);                 \
-    if (p) {                                 \
-        box64env.name = strdup(p);           \
-        box64env.is_##name##_overridden = 1; \
-        box64env.is_any_overridden = 1;      \
+#define STRING(NAME, name, wine)                \
+    p = GETENV(#NAME, wine);                    \
+    if (p && strcasecmp(#NAME, "BOX64_ARCH")) { \
```
ptitSeb, line 651: mm, what's this?

ksco: To exclude this if it's in the env variable.

ptitSeb: ah ok, fine

PR #3080, opened 2025-10-21
[LA64] Fixed LBT signal handling, also refined la64noext support

ksco:


src/libtools/signals.c
```diff
@@ -354,7 +354,30 @@ void copyUCTXreg2Emu(x64emu_t* emu, ucontext_t* p, uintptr_t ip) {
...
     #undef GO
     emu->ip.q[0] = ip;
-    emu->eflags.x64 = CONTEXT_REG(p, xFlags);
+#ifdef LA64
+    // TODO: move to ADJUST_ARCH
+    if (cpuext.lbt) {
+        struct lbt_context* lbt_ctx = NULL;
+        struct sctx_info* info = (struct sctx_info*)p->uc_mcontext.__extcontext;
+        while (info->magic && !lbt_ctx) {
+            if (info->magic == LBT_CTX_MAGIC)
```
xiangzhai, line 363: Hi @ksco 

There are no `struct lbt_context`, `struct sctx_info`, `unsigned long long __extcontext[0] __attribute__((__aligned__(16)));` for LA64_ABI_1.

How about [get_feature_flags_by_cpucfg](https://github.com/loongson/jdk/blob/loongarch-port/src/hotspot/cpu/loongarch/vm_version_loongarch.cpp#L124) and then [_features & CPU_LBT_X86](https://github.com/loongson/jdk/blob/loongarch-port/src/hotspot/cpu/loongarch/vm_version_loongarch.hpp#L294)? jdk works both for LA64 ABI 1.0 and 2.0.

Thanks,
Leslie Zhai

PR #3082, opened 2025-10-21
[RV64_DYNAREC] Added DB /2 FIST opcode

ksco:


src/dynarec/rv64/dynarec_rv64_helper.h
```diff
@@ -1401,7 +1401,7 @@
 #define emit_ror32c         STEPNAME(emit_ror32c)
 #define emit_shrd32c        STEPNAME(emit_shrd32c)
 #define emit_shld32c        STEPNAME(emit_shld32c)
-#define emit_shrd32         STEPNAME(emit_shld32)
```
ptitSeb, line 1404: oooof, nice catch!

PR #3086, opened 2025-10-22
[ENV] Introduced BOX64_PCLMULQDQ

ksco:


src/os/my_cpuid_common.c
```diff
@@ -105,7 +105,7 @@ void my_cpuid(x64emu_t* emu, uint32_t tmp32u)
                     | (BOX64ENV(cputype)?0:1)<<28     // HT / Multi-core
                     ;
             R_ECX =   1<<0      // SSE3
-                    | 1<<1      // PCLMULQDQ
```
ptitSeb, line 108: can you add VPCLMULQDQ also, behind both AVX2 and this one?

ksco: sure

PR #3092, opened 2025-10-24
[LA64_DYNAREC] Add more 00 opcodes.

phorcys:
For steam linux.
60 PUSHAD
61 POPAD
62 BOUND
F1 INT1
FF /5 JMP FAR Ed


ksco:
I will do the review Monday. @phorcys

ptitSeb:
Note that, instead of DEFAULT, on amr64 I generate a SIGILL with UDF for opcode that are invalid in 64bits (like PUSHAD/POPAD)

PR #3098, opened 2025-10-29
[LA64_DYNAREC] Replaced the ADDSL macro from RV64 to ALSL_D

ksco:


ptitSeb:
Something is wrong, LA64 tests are red

ksco:
Sorry, fixed.

ptitSeb:
Yep, interesting that the shift needs to be non 0

ksco:
Just noticed I made another typo in the fix commit.. I'll submit a new PR..

PR #3101, opened 2025-10-29
[STEAM] Added docs and some tweaks for Linux Steam, also added a new env var for it

ksco:


README_CN.md
```diff
@@ -81,9 +81,9 @@ Box64 的配置文件位于 `/etc/box64.box64rc` 和 `~/.box64rc`，均为 `.ini
 
 Box64 包装了 GTK 库，支持 gtk2 和 gtk3。
 
-### Linux Steam
+### Steam
 
-由于 Linux Steam 客户端应用是 32 位的，需要使用 Box86 或 Box32，但其本地服务器是 64 位的二进制文件。对于内存少于 6GB 的系统，可能需要使用交换文件以获得最佳性能。
```
ptitSeb, line 86: So the chinese version of the Steam paragraph is gone? The info about 6Gb of RAM can be important.

LukeShortCloud: On a related note, I was thinking about updating this exact part, too.

We should probably mention that 8 GB of system memory (RAM and swap combined) is required. Then an example can be provided that a system with 6 GB of RAM will need a 2 GB swap file or zram.

I still haven't done any benchmarks with zram on Arm but the results on x86 are impressive if you know how to configure it right. My online notes are down so I can't currently link to it right now. :-(

LukeShortCloud:
This may have got lost in the resolved review, just wanted to mention:

---

On a related note, I was thinking about updating this exact part, too.

We should probably mention that 8 GB of system memory (RAM and swap combined) is required. Then an example can be provided that a system with 6 GB of RAM will need a 2 GB swap file or zram.

I still haven't done any benchmarks with zram on Arm but the results on x86 are impressive if you know how to configure it right. My online notes are down so I can't currently link to it right now. :-(

ksco:
Sounds good to me, feel free to update the new STEAM.md after this is merged! Also, I think it’s less and less necessary to maintain the README translations manually, I think it’s possible to provide a in-place translation via Google Translate or others, I’ll check later.

PR #3104, opened 2025-10-31
[LA64_DYNAREC] Add more opcodes for steam.

phorcys:
06 PUSH ES
07 POP ES
36 SS
C8 ENTER
CB FAR RET
D7 XLAT
E4-E7  IN/OUT
FA CLI
FB STI
0F 0D /0 /2   PREFETCH
64 00 ADD Seg:Eb, Gb


ptitSeb:
@phorcys sorry, I just pushed some refactor on the segments handling, and `segs_serial` is gone. Can you update and fix the PR; just removing the write of `0` to `segs_serial` when writting a segment should be enough.

phorcys:
> @phorcys sorry, I just pushed some refactor on the segments handling, and `segs_serial` is gone. Can you update and fix the PR; just removing the write of `0` to `segs_serial` when writting a segment should be enough.

Got it, I'll update this patch. Thx.

PR #3107, opened 2025-10-31
[LA64_DYNAREC] Add macros help unaligned lock opcodes.

phorcys:
This patch add Macro for 32/64 bits lock opcodes.
Added lock add ed, gd/lock add ed, ib/id.

Please review this macro.

### DIFF
--- a/src/dynarec/la64/dynarec_la64_f0.c
+++ b/src/dynarec/la64/dynarec_la64_f0.c
@@ -62,11 +62,35 @@ uintptr_t dynarec64_F0(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 emit_add32(dyn, ninst, rex, ed, gd, x3, x4, x5);
             } else {
                 addr = geted(dyn, addr, ninst, nextop, &wback, x2, x1, &fixedaddress, rex, LOCK_LOCK, 0, 0);
-                MARKLOCK;
-                LLxw(x1, wback, 0);
-                ADDxw(x4, x1, gd);
-                SCxw(x4, wback, 0);
-                BEQZ_MARKLOCK(x4);
+                if (rex.w) {
+                    if (!ALIGNED_ATOMICxw) {
+                        ANDI(x3, wback, 0b111);
+                        BNEZ_MARK2(x3);
+                    }
+                    AMADD_DB_D(x1, gd, wback);
+                    B_MARK3_nocond;
+                } else {
+                    if (!ALIGNED_ATOMICxw) {
+                        ANDI(x3, wback, 0b11);
+                        BNEZ_MARK(x3);
+                    }
+                    // aligned 4byte
+                    AMADD_DB_W(x1, gd, wback);
+                    if (!ALIGNED_ATOMICxw) {
+                        B_MARK3_nocond;
+                        MARK;
+                        ANDI(x3, wback, 0b111);
+                        SLTI(x4, x3, 4);
+                        BEQZ_MARK2(x4); // addr %8 >4 , cross 8bytes or cross cacheline
+                        LOCK_32_IN_8BYTE(ADD_W(x4, x1, gd), x1, wback, x3, x4, x5, x6, x7);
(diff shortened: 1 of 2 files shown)


src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -1806,4 +1806,75 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+    op = atomic function
+    x1 = original ed
+    x2 = wback ed addr
+    x4 = result
+    x5 = 8 byte lower block for sc
+    x6 = aligned addr
+*/
+#define LOCK_3264_CROSS_8BYTE(op, x1, wback, x4, x5, x6) \
+    MV(x6, wback);                                       \
+    BSTRINS_D(x6, xZR, 2, 0);                            \
```
ptitSeb, line 1826: why not using `ANDI(x6, wback, ~0b111);` instead of `MV` + `BSTRINS_D` ?

phorcys: ANDI imm is zero extended.
```
The ANDI instruction performs the bitwise AND operation between the data in the general register rj and
the 12-bit immediate zero extension data; then the result is written into the general register rd.
```

ptitSeb: that's an odd choice to have the imm 0-extended while most other stuffs are sign extended!

phorcys: Indeed,  like 3230265646a48abef02a5c7873490e10ffc0afb5

src/dynarec/la64/dynarec_la64_f0.c
```diff
@@ -62,11 +62,35 @@ uintptr_t dynarec64_F0(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 emit_add32(dyn, ninst, rex, ed, gd, x3, x4, x5);
             } else {
                 addr = geted(dyn, addr, ninst, nextop, &wback, x2, x1, &fixedaddress, rex, LOCK_LOCK, 0, 0);
-                MARKLOCK;
-                LLxw(x1, wback, 0);
-                ADDxw(x4, x1, gd);
-                SCxw(x4, wback, 0);
-                BEQZ_MARKLOCK(x4);
+                if (rex.w) {
+                    if (!ALIGNED_ATOMICxw) {
```
ksco, line 66: We don't support this in rv64 and la64 for now.

ksco (changes requested):

src/dynarec/la64/dynarec_la64_f0.c
ptitSeb: But we can. It's just an optimisation that doesn't generate unaligned code (and save some space)

ksco: I know, but I think it's better to do nothing or implement this for all the lock opcodes.

ptitSeb: all opcodes ... at once? Tha seems like a lot of work? why not doing it gradualy?

ksco: Partially supporting an option sounds... weird?

phorcys: I'm ok with either.
I'm planning imp all unaligned lock ops with macros( except  one operand ops like bt/bts/btr/btc, or cmpxchg series)

ptitSeb: I think it's fine to have partial implementation of stuffs that are optimisation, but I leave that to you @ksco . If you preer all-once instead of gradual, no pb.

ksco: Sure, I don’t have strong opinions here.

ksco (approved):

PR #3127, opened 2025-11-05
[LA64_DYNAREC] Add 8bits lock op macros.

phorcys:
*  Add LOCK_8_ALIGNED_4BYTE, LOCK_8_IN_4BYTE macro.
  *  F0 00 LOCK ADD Eb, Gb

### DIFF
--- a/src/dynarec/la64/dynarec_la64_f0.c
+++ b/src/dynarec/la64/dynarec_la64_f0.c
@@ -52,6 +52,33 @@ uintptr_t dynarec64_F0(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
     GETREX();
 
     switch (opcode) {
+        case 0x00:
+            INST_NAME("LOCK ADD Eb, Gb");
+            SETFLAGS(X_ALL, SF_SET_PENDING, NAT_FLAGS_FUSION);
+            nextop = F8;
+            GETGD;
+            if (MODREG) {
+                ed = TO_NAT((nextop & 7) + (rex.b << 3));
+                emit_add8(dyn, ninst, ed, gd, x3, x4);
+            } else {
+                addr = geted(dyn, addr, ninst, nextop, &wback, x2, x1, &fixedaddress, rex, LOCK_LOCK, 0, 0);
+                if (cpuext.lam_bh) {
+                    AMADD_DB_B(x1, gd, wback);
+                } else {
+                    ANDI(x3, wback, 0b11);
+                    BNEZ_MARK(x3);
+                    LOCK_8_ALIGNED_4BYTE(ADD_W(x4, x1, gd), x1, wback, x3, x4, x5, x6, x7);
+                    B_MARK3_nocond;
+                    MARK;
+                    LOCK_8_IN_4BYTE(ADD_W(x4, x1, gd), x1, wback, x3, x4, x5, x6, x7);
+                    MARK3;
+                }
+                IFXORNAT (X_ALL | X_PEND) {
+                    emit_add8(dyn, ninst, x1, gd, x3, x4);
+                }
+            }
+            SMDMB();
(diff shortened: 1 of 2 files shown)


src/dynarec/la64/dynarec_la64_f0.c
```diff
@@ -52,6 +52,37 @@ uintptr_t dynarec64_F0(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+            GETGD;
+            if (MODREG) {
+                ed = TO_NAT((nextop & 7) + (rex.b << 3));
+                emit_add8(dyn, ninst, ed, gd, x3, x4);
+            } else {
+                addr = geted(dyn, addr, ninst, nextop, &wback, x2, x1, &fixedaddress, rex, LOCK_LOCK, 0, 0);
+                if (cpuext.lam_bh) {
+                    AMADD_DB_B(x1, gd, wback);
+                } else {
+                    if (!ALIGNED_ATOMICxw) {
```
ksco: This doesn't make sense. Byte atomic can never be unaligned.

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -1885,4 +1885,78 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+    use ll.w/sc.w to atomic operation . use amcas.db.w when possible
+    x1 = original ed
+    x2 = wback ed addr / aligned ed addr
+    x3 = 4 byte block for amcas write
+    x4 = result
+    x5 = 4 byte block
+    x6 = temp use
+    x7 = imm if inst use imm op
+*/
+#define LOCK_8_ALIGNED_4BYTE(op, x1, wback, x3, x4, x5, x6, x7) \
```
ptitSeb: using "x", like x1, x3, x4... is confusing as it's here macro parameters and not the globaly defined x3, x4 registers value...
I prefer "s" (as in scratch) to avoid confusion.

ksco: yeah it's not only confusing, but also dangerous.

ksco:
By the way, how were these opcodes tested?

src/dynarec/la64/dynarec_la64_f0.c
phorcys: `#define ALIGNED_ATOMICxw ((fixedaddress && !(fixedaddress & (((1 << (2 + rex.w)) - 1)))) || BOX64ENV(dynarec_aligned_atomics))`

Byte lock opcodes has no "nature alignment"
But LoongArch has only  ll.w/sc.w  ( amcas_db.b is exceptional)

So I create two macro:
LOCK_8_ALIGNED_4BYTE for 4-byte aligned  8-bits
LOCK_8_IN_4BYTE for 4-byte offset  1/2/3

ALIGNED_ATOMICxw here is to test  addr is  4byte aligned or BOX64_DYNAREC_ALIGNED_ATOMICS is turned on.
It's will not emit  LOCK_8_IN_4BYTE macro or addr ANDI test.

Else if addr is not 4-byte aligned and BOX64_DYNAREC_ALIGNED_ATOMICS is not turned on ,emit LOCK_8_IN_4BYTE related.

Indeed it looks quite strange, but it should work properly.

phorcys: My bad, BOX64_DYNAREC_ALIGNED_ATOMICS should not be tested.

src/dynarec/la64/dynarec_la64_helper.h
ptitSeb: Older macro will need to be changed too, but better be  in a separate PR.

ksco (approved):

PR #3137, opened 2025-11-07
[CORE] Print logs to stderr by default

ksco:


ptitSeb:
Are you sure it's a good idea? I directed to stdout by default to avoid triggering some error handler when some redirection is done?

ksco:
> Are you sure it's a good idea? I directed to stdout by default to avoid triggering some error handler when some redirection is done?

Oh, do you have any examples where logging to stderr would fail? I know that logging to stdout breaks some use of box64 with containers.

ptitSeb:
Nothing specific, but make sure to do some testing with it. Steam, bash, scripts, stuffs like that.

ksco:
> Nothing specific, but make sure to do some testing with it. Steam, bash, scripts, stuffs like that.

Sure, I launched Steam, which is okay. I'll do more tomorrow.

ksco:
I tried `box64-bash`, Steam, gog Linux installer. These are working fine.

PR #3139, opened 2025-11-08
[LA64_DYNAREC] Update lock 32/64 macro.

phorcys:
use s1-s7.


ksco (approved):

ksco:
All the macros are fixed now, right?

phorcys:
> All the macros are fixed now, right?

Yes,  except 16bits lock macro, which is not pushed yet.

PR #3168, opened 2025-11-12
[BOX32] Implemented getdents syscall

ksco:


src/emu/x86syscall_32.c
```diff
@@ -423,6 +437,31 @@ void EXPORT x86Syscall(x64emu_t *emu)
             if(R_EAX==0xffffffff && errno>0)
                 R_EAX = (uint32_t)-errno;
             break;*/
+        case 141: // getdents
+            native_linux_dirent_t* dirent_buffer = box_malloc(sizeof(native_linux_dirent_t) * (unsigned int)R_EDX);
```
ptitSeb, line 441: Old C (like the one on GCC 8) forbid local creating inside a case. You need to use global variable or create a local `{ }`....

ksco: And I also forgot to free this buffer...

PR #3172, opened 2025-11-13
[WRAPPER] Wrap libicu76

a9ix:
Hi, thank you for this project!

This should allow for running .NET (and other ICU version agnostic) applications on Debian Trixie without manually downloading older versions of the library.


ptitSeb:
Thanks, that can be usefull indeed!

PR #3184, opened 2025-11-17
[DYNAREC_PURGE] Decrement in_used before longjmp

devarajabc:
Signal handlers that use siglongjmp() to escape from dynarec blocks were abandoning blocks without decrementing their in_used reference count, causing blocks to remain permanently marked as in-use and preventing them from being freed even when they're no longer needed.

### DIFF
--- a/src/dynarec/dynablock.c
+++ b/src/dynarec/dynablock.c
@@ -221,6 +221,13 @@ void cancelFillBlock()
     LongJmp(GET_JUMPBUFF(dynarec_jmpbuf), 1);
 }
 
+void dynablock_leave_runtime(dynablock_t* db)
+{
+    if(!db) return;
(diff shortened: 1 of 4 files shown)


src/dynarec/dynablock.c
```diff
@@ -221,6 +221,12 @@ void cancelFillBlock()
     LongJmp(GET_JUMPBUFF(dynarec_jmpbuf), 1);
 }
 
+void dynablock_leave_runtime(dynablock_t* db)
+{
+    if(!db) return;
+    __atomic_fetch_sub(&db->in_used, 1, __ATOMIC_ACQ_REL);
```
ptitSeb, line 228: this should only be done is db->hot  is not 0. else, that means the block doesn't handle in_used!

PR #3188, opened 2025-11-18
[LA64_DYNAREC] Fix 8-bit lock macros, Add F0 08 LOCK OR Eb, Gb.

phorcys:
*  Fix 8-bit lock macros, using correct operand for non-lamcas path
  *  Fix 0F 00 LOCK ADD Eb, Gb
  *  add 0F 08 LOCK OR Eb, Gb

### DIFF
--- a/src/dynarec/la64/dynarec_la64_f0.c
+++ b/src/dynarec/la64/dynarec_la64_f0.c
@@ -56,7 +56,7 @@ uintptr_t dynarec64_F0(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
             INST_NAME("LOCK ADD Eb, Gb");
             SETFLAGS(X_ALL, SF_SET_PENDING, NAT_FLAGS_FUSION);
             nextop = F8;
-            GETGD;
+            GETGB(x7);
             if (MODREG) {
                 ed = TO_NAT((nextop & 7) + (rex.b << 3));
                 emit_add8(dyn, ninst, ed, gd, x3, x4);
@@ -65,13 +65,7 @@ uintptr_t dynarec64_F0(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 if (cpuext.lam_bh) {
                     AMADD_DB_B(x1, gd, wback);
                 } else {
-                    ANDI(x3, wback, 0b11);
-                    BNEZ_MARK(x3);
-                    LOCK_8_ALIGNED_4BYTE(ADD_W(x4, x1, gd), x1, wback, x3, x4, x5, x6);
-                    B_MARK3_nocond;
-                    MARK;
-                    LOCK_8_IN_4BYTE(ADD_W(x4, x1, gd), x1, wback, x3, x4, x5, x6, x7);
-                    MARK3;
+                    LOCK_8_OP(ADD_D(x4, x1, gd), x1, wback, x3, x4, x5, x6);
                 }
                 IFXORNAT (X_ALL | X_PEND) {
                     emit_add8(dyn, ninst, x1, gd, x3, x4);
@@ -122,6 +116,30 @@ uintptr_t dynarec64_F0(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                     emit_add32(dyn, ninst, rex, x1, gd, x3, x4, x5);
                 }
             }
+            break;
+        case 0x08:
+            INST_NAME("LOCK OR Eb, Gb");
+            SETFLAGS(X_ALL, SF_SET_PENDING, NAT_FLAGS_FUSION);
+            nextop = F8;
+            GETGB(x7);
+            if (MODREG) {
+                ed = TO_NAT((nextop & 7) + (rex.b << 3));
+                emit_or8(dyn, ninst, ed, gd, x3, x4);
+            } else {
+                addr = geted(dyn, addr, ninst, nextop, &wback, x2, x1, &fixedaddress, rex, LOCK_LOCK, 0, 0);
+                ANDI(x3, wback, 0b11);
+                SLLI_W(x3, x3, 3);
+                MV(x6, wback);
+                BSTRINS_D(x6, xZR, 1, 0); // aligned to 4-byte
+                SLL_W(x4, gd, x3);        // shift GETGB's BSTRPICK result to pos
+                AMOR_DB_W(x1, x4, x6);
+                SRL_W(x1, x1, x3);
+                BSTRPICK_D(x1, x1, 7, 0);
+                IFXORNAT (X_ALL | X_PEND) {
+                    emit_or8(dyn, ninst, x1, gd, x3, x4);
+                }
+            }
+            SMDMB();
             break;
         case 0x09:
             INST_NAME("LOCK OR Ed, Gd");
--- a/src/dynarec/la64/dynarec_la64_helper.h
+++ b/src/dynarec/la64/dynarec_la64_helper.h
@@ -1898,72 +1898,47 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
     }
 
 /*
-    LOCK_8_ALIGNED_4BYTE aligned 4 bytes 8bits op
-    use ll.w/sc.w to atomic operation . use amcas.db.w when possible
-    s1 = original ed
-    wback = ed addr / aligned ed addr
-    s3 = 4 byte block for amcas write
-    s4 = result
-    s5 = 4 byte block
-    s6 = temp use
-*/
-#define LOCK_8_ALIGNED_4BYTE(op, s1, wback, s3, s4, s5, s6)   \
-    if (cpuext.lamcas) {                                      \
-        LD_B(s1, wback, 0); /* amcas.b rd is sign extended */ \
-        MARKLOCK;                                             \
-        op;                                                   \
-        MV(s6, s1);                                           \
-        AMCAS_DB_B(s1, s4, wback);                            \
-        BNE_MARKLOCK(s1, s6);                                 \
-    } else {                                                  \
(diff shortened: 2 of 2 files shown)


src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -1906,13 +1906,14 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
     s4 = result
     s5 = 4 byte block
     s6 = temp use
+    x7 is used for imm or gb
```
ksco: x7?

src/dynarec/la64/dynarec_la64_f0.c
```diff
@@ -123,6 +123,30 @@ uintptr_t dynarec64_F0(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
             }
             break;
+        case 0x08:
+            INST_NAME("LOCK OR Eb, Gb");
+            SETFLAGS(X_ALL, SF_SET_PENDING, NAT_FLAGS_FUSION);
+            nextop = F8;
+            GETGB(x7);
+            if (MODREG) {
+                ed = TO_NAT((nextop & 7) + (rex.b << 3));
+                emit_add8(dyn, ninst, ed, gd, x3, x4);
```
ksco: Should be `emit_or8`.

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -1935,35 +1936,38 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+        MARKLOCK2;                                     \
+        MV(s6, s1);                                    \
+        op;                                            \
+        AMCAS_DB_W(s1, s4, wback);                     \
+        BNE_MARKLOCK2(s1, s6);                         \
+    } else {                                           \
+        if (wback != x2) {                             \
+            ed = x2;                                   \
+            MV(ed, wback);                             \
+        }                                              \
```
ksco: But what if `wback == x2`? Also, referring to x2 here is generally not a good idea; it makes the macro harder to use.

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -1935,35 +1936,38 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
     s4 = result
     s5 = 4 byte block
     s6 = mask
-    s7 = wback tmp
+    x7 is used for imm or gb
```
ksco: x7?

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -1935,35 +1936,38 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
-        OR(s4, s6, s4);                                    \
-        SC_W(s4, s7, 0);                                   \
-        BEQZ_MARKLOCK2(s4);                                \
+#define LOCK_8_IN_4BYTE(op, s1, wback, s3, s4, s5, s6) \
+    if (cpuext.lamcas) {                               \
+        LD_B(s1, wback, 0);                            \
+        MARKLOCK2;                                     \
+        MV(s6, s1);                                    \
+        op;                                            \
+        AMCAS_DB_W(s1, s4, wback);                     \
```
ksco: This is wrong, should be `AMCAS_DB_B`.

Btw, I think the current `LOCK_8_ALIGNED_4BYTE` and `LOCK_8_IN_4BYTE` do not make much sense -- if we have `AMCAS_DB_B` available, we should use it regardless of alignment, else a simple LL/SC can be used as a good enough fallback.

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -1935,35 +1936,38 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
-        ADDI_W(s6, xZR, -1);                               \
-        BSTRINS_W(s6, xZR, 31, 8);                         \
-        SLL_W(s6, s6, s3);                                 \
-        ANDN(s6, s5, s6);                                  \
-        OR(s4, s6, s4);                                    \
-        SC_W(s4, s7, 0);                                   \
-        BEQZ_MARKLOCK2(s4);                                \
+#define LOCK_8_IN_4BYTE(op, s1, wback, s3, s4, s5, s6) \
+    if (cpuext.lamcas) {                               \
+        LD_B(s1, wback, 0);                            \
```
ksco: `s1` is sign-extended, but the second operand `gd` is not; this can't be right. Change `LD_B` to `LD_BU`, and zero-extend s1 after `AMCAS_DB_B`.

src/dynarec/la64/dynarec_la64_f0.c
```diff
@@ -123,6 +123,30 @@ uintptr_t dynarec64_F0(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+                ANDI(x3, wback, 0b11);
+                SLLI_W(x3, x3, 3);
+                MV(x6, wback);
+                BSTRINS_D(x6, xZR, 1, 0); // aligned to 4-byte
+                SLL_W(x4, gd, x3);        // shift GETGB's BSTRPICK result to pos
+                AMOR_DB_W(x1, x4, x6);
+                SRL_W(x1, x1, x3);
+                BSTRPICK_D(x1, x1, 7, 0);
+                IFXORNAT (X_ALL | X_PEND) {
+                    emit_add8(dyn, ninst, x1, gd, x3, x4);
```
ksco: Should be `emit_or8`.

ksco (changes requested):

src/dynarec/la64/dynarec_la64_helper.h
phorcys: You are right, this patch should be reworked totally.
I'll change it to draft.

phorcys:
@ksco  please review ag.

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -1898,72 +1900,47 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+    if (cpuext.lamcas) {                            \
+        LD_BU(s1, wback, 0);                        \
+        MARKLOCK2;                                  \
+        MV(s6, s1);                                 \
+        op;                                         \
+        AMCAS_DB_B(s1, s4, wback);                  \
+        BSTRPICK_D(s1, s1, 7, 0);                   \
+        BNE_MARKLOCK2(s1, s6);                      \
+    } else {                                        \
+        ANDI(x3, wback, 0b11);                      \
```
ksco: `s3`?

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -1868,6 +1868,8 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
     if (wback != x2) {                                  \
         ed = x2;                                        \
         MV(ed, wback);                                  \
+    } else {                                            \
+        ed = wback; /* wback == x2, change it*/         \
```
ksco: `ed` is already initialized to `wback` before the `if`.

src/dynarec/la64/dynarec_la64_helper.h
phorcys: fixed.

src/dynarec/la64/dynarec_la64_helper.h
phorcys: fixed.

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -1898,72 +1898,47 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
-        BSTRINS_W(s6, xZR, 31, 8);                         \
-        SLL_W(s6, s6, s3);                                 \
-        ANDN(s6, s5, s6);                                  \
-        OR(s4, s6, s4);                                    \
-        SC_W(s4, s7, 0);                                   \
-        BEQZ_MARKLOCK2(s4);                                \
+#define LOCK_8_OP(op, s1, wback, s3, s4, s5, s6)    \
+    if (cpuext.lamcas) {                            \
+        LD_BU(s1, wback, 0);                        \
+        MARKLOCK2;                                  \
```
ptitSeb, line 1913: Isn't MARKLOCK2 needs to be placed before the LD_BU? If there is a loop, that means the reference changed, and so it needs to be fetch again?

phorcys: LoongArch amcas.{db,}.{b,h,w,d} rd,rk,rj  fetch orignal value in rd , sign extended.
So LD_BU only need to execute once.
`
AMCAS[_DB].{B/H/W/D} instruction performs a byte/half-word/word/double-word sized Compare-and-Swap operation on a specified address in memory: The byte/half-word/word/double-word value retrieved from memory (old memory value) is compared with the value stored in the [7:0]/[15:0]/[31:0]/[63:0] location of the general-purpose register rd (expected value), and the value stored in the [7:0]/[15:0]/[31:0]/[63:0] location of the general-purpose register rk (new value) is written to the same location in the memory only when the comparison results are equal. Regardless of whether the comparison results are equal or not, the old memory value is written to the general-purpose register rd after sign expansion.
`

ptitSeb: oh, you use the new loaded value as the new reference, without needing to reload again. I see.

PR #3189, opened 2025-11-18
[BOX32] Slient some warnings on padding fields

ksco:
These are just some paddings, don't take them too seriously, fixes the latest Linux Steam.


ksco:
~~Wait, it's auto-generated :(~~ Nope, it's not, but it might be a good idea to distinguish `_slient` or not, in case it is auto-generated in the future.

ptitSeb:
If it's just some padding, that a L or l should be used instead of p. Because failling to convert a pointer should be fatal. It's not just a warning, it's a abort for `to_ptrv`

ksco:
Sure but the signature is indeed a pointer.

PR #3206, opened 2025-11-23
Reverted some commits

ksco:
The sanity check appears to be incorrect, `__mask_was_saved` is already correctly handled in `my32_longjmp()`.


ptitSeb:
the sanity check should probably handle mask==0 too. but ok, lets revert this anyway.

PR #3234, opened 2025-11-27
[BOX32] Wrapped more symbols for libc

ksco:


src/wrapped32/wrappedlibc_private.h
```diff
@@ -442,10 +442,10 @@ GOW(ftruncate, iEil)
 GOW(ftruncate64, iEiI)
 //GOW(ftrylockfile, iEp)
 //GOM(fts_children, pEEpi) //%%
-//GOM(fts_close, iEEp)     //%%
-//GOM(fts_open, pEEpip)    //%%
-//GOM(fts_read, pEEp)      //%%
-// fts_set
+GO(fts_close, iFp)
```
ptitSeb: Should it be `iEp` to have wrapped `errno` on all those fts_* function?

ksco: indeed!

PR #3239, opened 2025-11-29
[BOX32] Fixed some inplace shrink/enlarge functions for libxrandr

ksco:


src/libtools/my_x11_conv.c
```diff
@@ -1495,20 +1495,20 @@ void* inplace_XRRMonitorInfo_enlarge(void* a, int n)
...
-            dst->automatic = src->automatic;
-            dst->noutput = src->noutput;
-            dst->x = src->x;
-            dst->y = src->y;
-            dst->width = src->width;
-            dst->height = src->height;
-            dst->mwidth = src->mwidth;
-            dst->mheight = src->mheight;
-            dst->outputs = from_ptrv(src->outputs);
+        for (int i = 0; i < n; ++i) {
```
ptitSeb, line 1498: Mmmm, actually, that for loop should be inverted too, else, if there is more then 1 element, the enlarge will overwrite itself.

ptitSeb: I'll fix it in a another commit

ksco: Indeed! Thanks. In the cs1.6 case, n is 1.

ptitSeb: Done with https://github.com/ptitSeb/box64/commit/c4e09fddb9e15de5d7b096af17dabc5fe0d2663c

PR #3244, opened 2025-12-05
Add lookup cache for customMalloc

devarajabc:
Allocation functions (map64_customMalloc, map128_customMalloc,
internal_customMalloc) always scanned blocks from i=0 to find free space.
Since consecutive allocations often succeed in the same block(`blocklist_t`, the memory pool), we can
skip the linear scan by trying the last used block first.

Add type-specific caches so each allocator remembers its last successful
block and tries it before falling back to linear scan. The shared last_block_index 
is still updated for findBlock compatibility.

  Test case: box64 + wine + steam.exe on rpi5

[Results](https://github.com/user-attachments/files/23969375/reslt333.txt):
  - map64_customMalloc:    99.9% hit rate (25,920 hits / 26 misses)
  - map128_customMalloc:   99.9% hit rate (3,484 hits / 3 misses)
  - internal_customMalloc: 73.9% hit rate (2,574 hits / 907 misses)
  - findBlock:             72.8% hit rate (14,410 hits / 5,374 misses)

  At cache hit time, the allocator managed 76-812 blocks (p25-p99),
  confirming that skipping linear scan provides measurable benefit.

### DIFF
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -109,8 +109,12 @@ static int                 c_blocks = 0;       // capacity of blocks for custom 
 static blocklist_t*        p_blocks = NULL;    // actual blocks for custom malloc
 static int                 setting_prot = 0;
 
-// Block lookup cache 
+// Block lookup cache (shared for findBlock/free)
 static int last_block_index = -1;
+// Per-type allocation caches for better hit rates
+static int last_block_index_map64 = -1;
+static int last_block_index_map128 = -1;
+static int last_block_index_list = -1;
 
 typedef union mark_s {
     struct {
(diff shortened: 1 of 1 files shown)


ptitSeb:
I will not be available this weekend so I'll probably only check on monday.

src/custommem.c
```diff
@@ -584,6 +588,25 @@ void* map128_customMalloc(size_t size, int is32bits)
 {
     size = 128;
     mutex_lock(&mutex_blocks);
+    // Try cached hint first
+    if(last_block_index_map128 >= 0 && last_block_index_map128 < n_blocks) {
+        blocklist_t* hint = &p_blocks[last_block_index_map128];
+        if(hint->block && (hint->type == BTYPE_MAP) && hint->maxfree && (hint->is32bits==is32bits)) {
```
xiangzhai: if (hint && ...)

src/custommem.c
```diff
@@ -694,21 +720,47 @@ void* map64_customMalloc(size_t size, int is32bits)
 {
     size = 64;
     mutex_lock(&mutex_blocks);
+    // Try cached hint first
+    if(last_block_index_map64 >= 0 && last_block_index_map64 < n_blocks) {
+        blocklist_t* hint = &p_blocks[last_block_index_map64];
+        if (hint->block
```
xiangzhai: Ditto

src/custommem.c
```diff
@@ -807,6 +860,26 @@ void* internal_customMalloc(size_t size, int is32bits)
     blockmark_t* sub = NULL;
     size_t fullsize = size+2*sizeof(blockmark_t);
     mutex_lock(&mutex_blocks);
+    // Try cached hint first
+    if(last_block_index_list >= 0 && last_block_index_list < n_blocks) {
+        blocklist_t* hint = &p_blocks[last_block_index_list];
+        if(hint->block && (hint->type == BTYPE_LIST) && hint->maxfree>=init_size && (hint->is32bits==is32bits)) {
```
xiangzhai: Ditto

ptitSeb:
Ok, LGTM too, merging.

xiangzhai:
Merry Christmas :christmas_tree:

PR #3280, opened 2025-12-15
[RCFILE] Added a new option BOX64_PROFILE

ksco:


src/tools/env.c
```diff
@@ -217,6 +218,34 @@ static void applyCustomRules()
...
+            SET_BOX64ENV_IF_EMPTY(dynarec_dirty, 1);
+            SET_BOX64ENV_IF_EMPTY(dynarec_forward, 1024);
+        } else if (!strcasecmp(box64env.profile, "fastest")) {
+            SET_BOX64ENV_IF_EMPTY(dynarec_callret, 1);
+            SET_BOX64ENV_IF_EMPTY(dynarec_bigblock, 3);
+            SET_BOX64ENV_IF_EMPTY(dynarec_safeflags, 0);
+            SET_BOX64ENV_IF_EMPTY(dynarec_strongmem, 0);
+            SET_BOX64ENV_IF_EMPTY(dynarec_dirty, 1);
+            SET_BOX64ENV_IF_EMPTY(dynarec_forward, 1024);
+        }
```
ptitSeb, line 253: Maybe a warning if profile is different then "default" and so is unknown?

ksco: Done

PR #3286, opened 2025-12-17
[DYNACHCHE] correct dynacache typo

luoqiaoling:


src/tools/env.c
```diff
@@ -1152,11 +1152,6 @@ int ReadDynaCache(const char* folder, const char* name, mapping_t* mapping, int
         fclose(f);
         return DCERR_DYNARCHVER;
     }
-    if(header.arch_version!=ARCH_VERSION) {
```
ptitSeb, line 1155: Why did you remove this?

ptitSeb: Oh, because it's there 2 times

PR #3297, opened 2025-12-22
More wrapped symbols for SDL2

Neltharion01:
[This game](https://github.com/chaosforgeorg/drl/releases/tag/0_10_0e) crashes on unwrapped SDL2/image/mixer symbols

Some of added symbols were not tested, because game doesn't use them (only checks dlsym() != NULL on load)

### DIFF
--- a/src/wrapped/wrappedsdl2.c
+++ b/src/wrapped/wrappedsdl2.c
@@ -884,13 +884,19 @@ EXPORT int my2_SDL_GetCPUCount(x64emu_t* emu)
     return ret;
 }
 
-static uintptr_t my_FilterEvents_callback_f = 0;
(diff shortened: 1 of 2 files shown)


src/wrapped/wrappedsdl2.c
```diff
@@ -842,6 +884,15 @@ EXPORT int my2_SDL_GetCPUCount(x64emu_t* emu)
     return ret;
 }
 
+static uintptr_t my_FilterEvents_callback_f = 0;
```
ptitSeb: This method will not work on multi-threaded program, if tht function is used in 2 different threads. Why not use the full callback method for this one?

Neltharion01: rewrote that part to avoid global variable

PR #3302, opened 2025-12-23
[SIGNAL] Moved certain warnings to debug level

ksco:


src/libtools/signals.c
```diff
@@ -1275,7 +1275,7 @@ void my_sigactionhandler_oldcode_64(x64emu_t* emu, int32_t sig, int simple, sigi
             siglongjmp(emu->jmpbuf, skip);
             #endif
         }
-        printf_log(LOG_INFO, "Warning, context has been changed in Sigactionhanlder%s\n", (sigcontext->uc_mcontext.gregs[X64_RIP]!=sigcontext_copy.uc_mcontext.gregs[X64_RIP])?" (EIP changed)":"");
+        printf_log(LOG_DEBUG, "Warning, context has been changed in Sigactionhanlder%s\n", (sigcontext->uc_mcontext.gregs[X64_RIP] != sigcontext_copy.uc_mcontext.gregs[X64_RIP]) ? " (EIP changed)" : "");
```
ptitSeb, line 1278: this one should stay at LOG_INFO, but the message is incomplete: it should mention that there is no jmpbuffer ready (and so the new RIP will not be satisfied)

ksco: This is the one repeated forever.

ptitSeb: Then there is a real issue. This specific message is an error, not a warning.

ptitSeb: Those message appear when the signal should have be resumed using a siglongjmp, but the jumpbuffer is not available for some reason. This is not ok to continue.

PR #3305, opened 2025-12-25
Make sure managed memory blocks are always aligned to 16 bytes

ksco:


ptitSeb:
so that was the source of the X64 issue with the new bridges...

ptitSeb:
Nice find.

Issue #222, opened 2022-01-31
Segmentation fault - BepInEx on valheim dedicated server


ibogatac:
Tested this on Oracle OCI running on a Ubuntu 20.04, valheim server runs fine, but BepInEx used for mods runs into a segmentation fault. This is my first time using box64 and box86.

`/home/ubuntu/valheim_server/valheim_server.x86_64
ERROR: ld.so: object 'libdoorstop_x64.so' from LD_PRELOAD cannot be preloaded (c                                                                                                                                                             annot open shared object file): ignored.
Dynarec for ARM64, with extension: ASIMD AES CRC32 PMULL ATOMICS PageSize:4096
Box64 with Dynarec v0.1.7 ca3247d built on Jan 25 2022 09:08:45
Using default BOX64_LD_LIBRARY_PATH: ./:lib/:lib64/:x86_64/:bin64/:libs64/
Using default BOX64_PATH: ./:bin/
Counted 33 Env var
BOX64 try to Preload libdoorstop_x64.so
Looking for /home/ubuntu/valheim_server/valheim_server.x86_64
Using emulated /home/ubuntu/valheim_server/doorstop_libs/libdoorstop_x64.so
Using native(wrapped) libc.so.6
Using native(wrapped) ld-linux-x86-64.so.2
Using native(wrapped) libpthread.so.0
Using native(wrapped) librt.so.1
Using emulated UnityPlayer.so
Using native(wrapped) libm.so.6
Using emulated /lib/x86_64-linux-gnu/libgcc_s.so.1
Using native(wrapped) libdl.so.2
Found UnityPlayer, hooking into it instead
[...]

PR #3331, opened 2026-01-05
[ELF] Fix SIGSEGV when calling dlinfo RTLD_DI_LINKMAP for emulated code

matthewabbott:
## Issue

When emulated code calls `dlinfo(handle, RTLD_DI_LINKMAP, &lmap)`, the returned `link_map->l_ld` would point at box64’s internal parsed copy of `.dynamic`, which contains raw virtual addresses from the ELF file. Those addresses weren’t relocated to the guest’s load base, so guest code that dereferenced tags like `DT_SYMTAB` would read nonsensical memory locations and crash:

```
SIGSEGV ... libdoorstop.so/plthook_enum
for accessing 0x172930 (code=1/prot=0)
```

In the case of the above error, the symbol table actually lived at `0x7fff00172930`, but `l_ld` exposed the file value `0x172930`.

## Fix

Add methods:
- `GetLoadedDynamicSection()` to return the `.dynamic` address loaded in ELF memory (not the internal copy).
- and `PatchLoadedDynamicSection()` which adds the load delta to pointer-type `d_ptr` entries so they point at the runtime addresses in the loaded ELF mapping. Pages might already be read only, so temporarily `mprotect` as needed.

Update `l_ld` assignments now use the loaded, patched section.

---

<details>
<summary>Reproduction (pre-fix)</summary>

Using [BepInEx 5.x (Linux)](https://github.com/BepInEx/BepInEx) with a Unity game on ARM64 via box64. BepInEx’s Doorstop/plthook calls `dlinfo(..., RTLD_DI_LINKMAP, ...)`, walks `l_ld` to find `DT_SYMTAB`, `DT_STRTAB`, etc., and segfaults when it dereferences the unpatched file address (e.g. `0x172930`). Possibly a known issue with bepinex/box64 compatibility: https://github.com/ptitSeb/box64/issues/222, https://github.com/BepInEx/BepInEx/issues/336.

Steps:
1) Install BepInEx 5.x (Linux).
2) Run a Unity game under box64 with BepInEx configured.
3) Doorstop’s `plthook_open_real()` calls `dlinfo(handle, RTLD_DI_LINKMAP, &lmap)`.
4) plthook iterates `lmap->l_ld` to find dynamic entries.
5) SIGSEGV on unpatched address (e.g. `0x172930`).
</details>

---

<details>
<summary>Design choices</summary>

- Pointer-tag allowlist switch statement ([`isDynamicTagPointer`](https://github.com/matthewabbott/box64/blob/8a6d35210/src/elfs/elfloader.c#L1515-L1554)):

Dynamic entries use a union, but only pointer-typed tags should be delta-adjusted. The idea here is that an explicit allowlist avoids patching size/flag tags like `DT_STRSZ` or `DT_FLAGS`. And then I also added numeric fallbacks (e.g. `0x6ffffef5` for `DT_GNU_HASH`) to try and keep compatibility on older headers. But this feels a bit ugly to me. Open to suggestions if you have other ideas about how to do this.

- Iterating to find `PT_DYNAMIC` ([`GetLoadedDynamicInfo`](https://github.com/matthewabbott/box64/blob/8a6d35210/src/elfs/elfloader.c#L1484-L1505)):

The loaded `.dynamic` address/size come from `PT_DYNAMIC` (`delta + p_vaddr`, `p_memsz`). The internal parsed copy (`h->Dynamic._64/_32`) isn’t usable for `l_ld`. Similar pattern to how [`dl_iterate_phdr_findsymbol`](https://github.com/matthewabbott/box64/blob/8a6d35210/src/elfs/elfloader.c#L1940-L1951) locates the dynamic section.

- Regarding `mprotect`ing in ([`PatchLoadedDynamicSection`](https://github.com/matthewabbott/box64/blob/8a6d35210/src/elfs/elfloader.c#L1556-L1607)):

PatchLoadedDynamicSection will patch `.dynamic`'s `d_ptr` values after load to keep the change localized and avoid restructuring the loader, but that means that it can run into pages with different protections (eg RELRO). So the idea is to check each page via `getProtection()`, temporarily add `PROT_WRITE` only where needed, roll back on failure, then restore protections. Untracked pages (protection=0) are left untouched. Perhaps it'd be better to just restructure the loader though if you prefer or if you have a suggestion.

- Idempotency:

`dynamic_patched` prevents double-adjusting the same ELF.
</details>

### DIFF
--- a/src/elfs/elfloader.c
+++ b/src/elfs/elfloader.c
@@ -1504,7 +1504,6 @@ static int GetLoadedDynamicInfo(elfheader_t* h, dynamic_info_t* info)
     return 0;
 }
 
-
 void* GetLoadedDynamicSection(elfheader_t* h)
 {
     dynamic_info_t info;
@@ -1569,11 +1568,20 @@ void PatchLoadedDynamicSection(elfheader_t* h)
     uintptr_t page_addr = dyn_addr & ~(box64_pagesize - 1);
     uintptr_t page_end = (dyn_addr + dyn_size + box64_pagesize - 1) & ~(box64_pagesize - 1);
 
-    uint32_t old_prot = getProtection(page_addr);
-    int need_mprotect = !old_prot || !(old_prot & PROT_WRITE);
-    if(need_mprotect) {
-        uint32_t new_prot = old_prot ? (old_prot | PROT_WRITE) : (PROT_READ | PROT_WRITE);
-        mprotect((void*)page_addr, page_end - page_addr, new_prot);
+    int need_restore = 0;
+    for(uintptr_t page = page_addr; page < page_end; page += box64_pagesize) {
+        uint32_t old_prot = getProtection(page);
+        if(old_prot && !(old_prot & PROT_WRITE)) {
+            if(mprotect((void*)page, box64_pagesize, (old_prot | PROT_WRITE) & ~PROT_CUSTOM)) {
+                for(uintptr_t restore = page_addr; restore < page; restore += box64_pagesize) {
+                    uint32_t restore_prot = getProtection(restore);
+                    if(restore_prot && !(restore_prot & PROT_WRITE))
+                        mprotect((void*)restore, box64_pagesize, restore_prot & ~PROT_CUSTOM);
+                }
+                return;
+            }
+            need_restore = 1;
+        }
     }
 
     if(box64_is32bits) {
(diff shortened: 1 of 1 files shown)


ptitSeb:
That looks good to me. I have no problem with your design choices.

The only way the patching can fail is if mprotect fail right? That should not happen I guess, so it should be safe. We'll see in practice if the "unprotect/restore" process is too time consuming or not.

matthewabbott:
Yeah RE: patch failure, the only hard failure is mprotect returning non‑zero, in which case we roll back any pages already made writable and return without setting `dynamic_patched`, the idea being to have behavior fall back to 'no patch'.

Other early exits are 'no `PT_DYNAMIC`' or `delta == 0`. As for overhead, it should be 1-2 pages per ELF (which I believe should be pretty small), and we leave untracked pages alone to avoid changing protection state.

ptitSeb:
Yeah, I just try to launch a software that loads tens of library (intel FPGA Quartus Prime Programmer) and there was no noticable longer delay... So, that looks good to me.

Thank your Pull Request.

matthewabbott:
Happy to help!

PR #3332, opened 2026-01-05
Bundle fixes

LukeShortCloud:


LukeShortCloud:
Somewhat related, here are the bundles you can upload for the 0.4.0 release on GitHub. It has a few new libraries since 0.3.8.

https://winesapos.lukeshort.cloud/box64-bundle-x86-libs-0.4.0.tar.gz
https://winesapos.lukeshort.cloud/box64-bundle-x86-pkgs-0.4.0.tar.gz

PR #3333, opened 2026-01-07
[LA64_DYNAREC] Refined more RCL/RCR opcodes

ksco:


ptitSeb:
Wow, there is even a modulo opcode!

PR #3345, opened 2026-01-09
[LA64_DYNAREC] Enabled more patterns for nativeflags

ksco:


ptitSeb:
the GETGWEW / GETGBEB is nice, but be sure that no emiter have issue when gd==ed (I know it's not the case anymore on amr64 for example, like with `emit_add8(...)`)

ksco:
> the GETGWEW / GETGBEB is nice, but be sure that no emiter have issue when gd==ed (I know it's not the case anymore on amr64 for example, like with `emit_add8(...)`)

Oh I missed this comment, I’ll double check later.

PR #3346, opened 2026-01-09
[LA64_DYNAREC] Fix IDIV Ew dividend construction

yzewei:
**Problem**

The IDIV Ew instruction in the LA64 dynarec implementation produces incorrect results due to improper dividend construction.

Running the following minimal test case (using inline asm to force IDIV Ew) returns 0 instead of the expected -10922.

Expected behavior: tf_h(-32766) should return -10922 (0xd556). Actual behavior: tf_h(-32766) returns 0 (0x0000).

**Repro Code**
```c
#include <stdio.h>

// inline asm to force IDIV Ew
__attribute__((noinline)) short tf_h(short c) {
    short q;
    __asm__ volatile (
        "movw  %[c], %%ax\n\t"
        "cwd\n\t"                // DX:AX = sign_extend(AX)
        "movw  $3, %%cx\n\t"
        "idivw %%cx\n\t"         // AX = DX:AX / CX
        "movw  %%ax, %[q]\n\t"
        : [q] "=r"(q)
        : [c] "r"(c)
        : "ax", "dx", "cx"
    );
    return q;
}

int main(void) {
    short v = (short)0x8002; // -32766
    short r = tf_h(v);
    printf("tf_h(%d) = %d (0x%04x)\n", v, r, (unsigned short)r);
    return 0;
}
```

**Repro Log (Before Fix)**
```
[BOX64] Transfert to main(1, ...)
2869102|0x40118f: Calling printf(0x402004, 0xFFFF8002, 0x0, ...) 
=> tf_h(-32766) = 0 (0x0000)
```

**Now**
```
 return 0x0[function: 0x30010020]
2870319|0x40118f: Calling printf(0x402004, 0xFFFF8002, 0xFFFFD556, ...) =>tf_h(-32766) = -10922 (0xd556)
 return 0x1F
 return 0x0
[BOX64] Emulation finished, EAX=0
[BOX64] Calling atexit registered functions (exiting box64)
[BOX64] Calling atexit registered functions
[BOX64] Calling fini for all loaded elfs and unload native libs
[BOX64] Calling atexit registered functions for elf: 0x382b2940//home/yzw/python-trans/box64-up/test/t1
[BOX64] Calling Fini[0] for /home/yzw/python-trans/box64-up/test/t1 @0x4010f0
[BOX64] Calling Fini for /home/yzw/python-trans/box64-up/test/t1 @0x401198
[BOX64] Unloaded main elf: Will Dec RefCount of 1 libs
[...]

PR #3354, opened 2026-01-12
[LA64_DYNAREC] Fix IDIV Ew with sign-extended DX:AX

yzewei:
**Problem**
After the previous [fix,](https://github.com/ptitSeb/box64/pull/3346) IDIV Ew still returns 0 in the minimal inline-asm test for negative inputs.

Root Cause: The DX:AX construction logic lost sign-extension. The combined dividend was interpreted as 0x00000000ffff8002 (positive large number) instead of 0xffffffffffff8002 (negative short), producing an incorrect quotient.

**Repro Case**
```c
#include <stdio.h>

// Inline asm to force IDIV Ew
__attribute__((noinline)) short tf_h(short c) {
    short q;
    __asm__ volatile (
        "movw  %[c], %%ax\n\t"
        "cwd\n\t"                // DX:AX = sign_extend(AX)
        "movw  $3, %%cx\n\t"
        "idivw %%cx\n\t"         // AX = DX:AX / CX
        "movw  %%ax, %[q]\n\t"
        : [q] "=r"(q)
        : [c] "r"(c)
        : "ax", "dx", "cx"
    );
    return q;
}

int main(void) {
    short v = (short)0x8002;
    short r = tf_h(v);
    printf("tf_h(%d) = %d (0x%04x)\n", v, r, (unsigned short)r);
    return 0;
}
```

**Fix Implementation**
Used BSTRPICK_D + BSTRINS_D to manually construct the DX:AX value.

Applied ADDI_W (add immediate word) to properly sign-extend the resulting 32-bit dividend into the 64-bit register before executing DIV_W.(need 0x00000000ffff8002 not 0xffffffffffff8002)

Validation
GDB check at div.w instruction confirms correct inputs:
```c
r15 (dividend) = 0xffffffffffff8002  <-- Correctly sign-extended
r14 (divisor)  = 0x3
r16 (quotient) = 0xffffffffffffd556  (-10922)
```

### DIFF
--- a/src/dynarec/la64/dynarec_la64_66.c
+++ b/src/dynarec/la64/dynarec_la64_66.c
@@ -1392,8 +1392,7 @@ uintptr_t dynarec64_66(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                         MARK3;
                     }
                     BSTRPICK_D(x2, xRAX, 15, 0);
-                    BSTRINS_D(x2, xRDX, 31, 16);
-                    ADDI_W(x2, x2, 0);
+                    BSTRINS_W(x2, xRDX, 31, 16);
                     DIV_W(x3, x2, ed);
                     MOD_W(x4, x2, ed);
                     BSTRINSz(xRAX, x3, 15, 0);


yzewei:
My bad on the last PR. I failed to validate the previous fix adequately. Fixed now.
PTAL! @ksco

src/dynarec/la64/dynarec_la64_66.c
```diff
@@ -1393,6 +1393,7 @@ uintptr_t dynarec64_66(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                     }
                     BSTRPICK_D(x2, xRAX, 15, 0);
                     BSTRINS_D(x2, xRDX, 31, 16);
+                    ADDI_W(x2, x2, 0);
```
ksco:
```suggestion
                    BSTRINS_W(x2, xRDX, 31, 16);
```

ksco:
Thank you for the fix, I didn’t know this! So I submitted another PR #3355 for DIV Ew, which is similar to this.

yzewei:
> Thank you for the fix, I didn’t know this! So I submitted another PR #3355 for DIV Ew, which is similar to this.

Glad to help! It was indeed a tricky edge case to spot.

Issue #3367, opened 2026-01-14
arc4random symbols not found even though libbsd provides them (wrappedlibbsd uses dlopen(NULL))


yzewei:
### Question
When loading x86_64 libstdc++ (e.g. via numpy), Box64 fails to resolve `arc4random*` symbols even though host `libbsd` exports them. This causes PLT relocation failure and prevents numpy from loading.

### Environment
- Host arch: loongarch64
- Box64 version / commit: master
- libbsd: /usr/lib64/libbsd.so.0.9.1 (exports arc4random*)
- Repro command: 
  BOX64_LOG=3 box64 python3 -c "import numpy"

### Logs
[BOX64] Warning, function arc4random not found in lib libbsd.so.0
[BOX64] Error: Symbol arc4random not found, cannot apply R_X86_64_JUMP_SLOT ... in /usr/lib/box64-x86_64-linux-gnu/libstdc++.so.6
[BOX64] Error: relocating Plt symbols in elf libstdc++.so.6
[BOX64] Error initializing needed lib libc.so.6
...
ImportError: numpy C-extensions failed to load

$ nm -D /usr/lib64/libbsd.so.0.9.1 | grep arc4random
0000000000004d98 T arc4random
0000000000004938 T arc4random_addrandom
0000000000005218 T arc4random_buf
0000000000004908 T arc4random_stir
0000000000005698 T arc4random_uniform

### Current libbsd wrapper
`wrappedlibbsd.c` uses PRE_INIT:
```c
#define PRE_INIT \
    if(1) lib->w.lib = dlopen(NULL, RTLD_LAZY | RTLD_GLOBAL); \
    else
```

This makes the wrapper use the main program global handle instead of dlopen("libbsd.so.0"). If libbsd was not loaded as RTLD_GLOBAL beforehand, dlsym(NULL, "arc4random") fails even though the file exports it.

### Two ways:
- Keeping wrappedlibbsd soft, but adding fallback symbols in wrappedlibc (GOM → my_arc4random*)?
- Or making libbsd a hard dependency (explicit dlopen("libbsd.so.0") / RTLD_GLOBAL)?

If we switch to a hard dependency, what impact do you anticipate across distros/environments where libbsd is not present or not global‑visible?

I can provide a patch if desired.

ptitSeb:
I think some version of libc don't have a separate libbsd anymore and all symbols are inside libc.
[...]

PR #3370, opened 2026-01-14
[WRAPPER] Add arc4random to libc

yzewei:
Add arc4random to libc to allow x86_64 NumPy wheels to load.

### DIFF
--- a/src/wrapped/wrappedlibc_private.h
+++ b/src/wrapped/wrappedlibc_private.h
@@ -2690,10 +2690,14 @@ GO(android_set_abort_message, vFp)
 #ifdef STATICBUILD
 GO(dummy_pFLp, pFLp)
 GO(dummy_pFpLLp, pFpLLp)
-GO(arc4random, uFv)
 #else
 // not needed in no-static build
 //GO(dummy_pFLp, pFLp)
 //GO(dummy_pFpLLp, pFpLLp)
-//GO(arc4random, uFv)
+#endif
+
+#ifdef LA64_ABI_1
+GOM(arc4random, uFv)
+#else
+GO(arc4random, uFv)
 #endif
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -818,7 +818,7 @@ if(NOT CI)
         OUTPUT "${BOX64_ROOT}/src/wrapped/generated/functions_list.txt"
         COMMAND "${PYTHON_EXECUTABLE}" "${BOX64_ROOT}/rebuild_wrappers.py"
         "${BOX64_ROOT}"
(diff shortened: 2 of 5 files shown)


src/wrapped/wrappedlibc_private.h
```diff
@@ -2695,5 +2695,5 @@ GO(arc4random, uFv)
 // not needed in no-static build
 //GO(dummy_pFLp, pFLp)
 //GO(dummy_pFpLLp, pFpLLp)
-//GO(arc4random, uFv)
+GOM(arc4random, uFv)
```
ksco, line 2700: why use GOM here, can't you just remove the comment `//`?

yzewei: GO tries to resolve arc4random from libc, but glibc doesn’t export it. I tested uncommenting GO(arc4random, uFv) and it fails:

```
[BOX64] Warning, function arc4random not found in lib libc.so.6
[BOX64] Error: Symbol arc4random not found, cannot apply R_X86_64_JUMP_SLOT ... in libstdc++.so.6
[BOX64] Error: relocating Plt symbols in elf libstdc++.so.6
```
So GOM is needed to provide a fallback my_arc4random and avoid relocation failure.

ksco:
```
$ readelf -Ws libc.so.6 | grep arc4random@
  2526: 0000000000038a50    92 FUNC    WEAK   DEFAULT   12 arc4random@@GLIBC_2.36
```

I have it in my glibc. You are from old world right? Maybe guard this under `#ifdef LA64_ABI_1`.

yzewei: I will do that, as I am indeed working in an ABI 1.0 environment. Thanks for the suggestion!

yzewei:
CI failed because rebuild_wrappers.py doesn't know LA64_ABI_1 without the CMake change,  need [https://github.com/ptitSeb/box64/pull/3371](url)

ksco:
It's simple enough and related, please put it in this PR for the ease of merging.

yzewei:
> It's simple enough and related, please put it in this PR for the ease of merging.

OK

ptitSeb:
But what about #3367 ?

yzewei:
> But what about #3367 ?

This PR is a minimal workaround for the missing symbol I hit in practice: only arc4random was unresolved on  LA64 ABI1/glibc, and other arc4random_* weren’t referenced. So I only wrapped arc4random to keep the change small.

#3367 is still valid and not fully addressed here. fixed in https://github.com/ptitSeb/box64/pull/3373

ptitSeb:
Please do not introduce a my_  specific to a system. Instead, create a weak version of the arc4random in libc.c, like I have done previously for `strlcpy` and `__strlcpy_chk` so the fix is more generic.

PR #3376, opened 2026-01-14
[WRAPPER] Provide weak arc4random in libc

yzewei:
Add a weak arc4random fallback in wrappedlibc so symbol resolution succeeds on hosts without arc4random in libc/libbsd. This avoids PLT relocation failures when loading libstdc++/numpy. #3370

### DIFF
--- a/src/wrapped/wrappedlibc.c
+++ b/src/wrapped/wrappedlibc.c
@@ -4000,12 +4000,13 @@ EXPORT char* secure_getenv(const char* name)
     return getenv(name);
 }
 
-#ifdef STATICBUILD
 uint32_t get_random32();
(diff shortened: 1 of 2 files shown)


src/wrapped/wrappedlibc.c
```diff
@@ -4000,12 +4000,13 @@ EXPORT char* secure_getenv(const char* name)
     return getenv(name);
 }
 
-#ifdef STATICBUILD
 uint32_t get_random32();
-__attribute__((weak)) uint32_t arc4random()
+EXPORT __attribute__((weak)) uint32_t arc4random(void)
```
ptitSeb: don't EXPORT and do weak at the same time, it's one or the other, not both.

PR #3386, opened 2026-01-17
[RV64_DYNAREC] Fix delta_ip range checks ([LA64_DYNAREC] too)

Aqua-Peak:
（1）“-0xfffffff” is seen as “unsigned int”。Here what is need shoule be INT_MIN。“-0x80000000L”  is equal to INT_MIN and "L" is important that seen as long。  the exmaple code（I have tested in RV64(Vision Star2）and arm64(Pi5).
```c
#include <stdio.h>
int main(){
  ssize_t a = 0;
  if(a < -0xffffffff)
     printf("-0xfffffff is seen as positive value \n");
  return 0;
}
```
(2)"0xffffffff". Here what is need shoule be INT_MAX that should be "0x7fffffff"。 (3) the same situation is existed in “la64 GETIP_”.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_helper.h
+++ b/src/dynarec/rv64/dynarec_rv64_helper.h
@@ -1138,7 +1138,7 @@
         } else if (_delta_ip == 0) {                              \
         } else if (_delta_ip >= -2048 && _delta_ip < 2048) {      \
             ADDI(xRIP, xRIP, _delta_ip);                          \
-        } else if (_delta_ip < 0 && _delta_ip >= -0x80000000L) {   \
+        } else if (_delta_ip < 0 && _delta_ip >= -0xffffffffL) {   \
             MOV32w(scratch, -_delta_ip);                          \
             SUB(xRIP, xRIP, scratch);                             \
         } else if (_delta_ip > 0 && _delta_ip <= 0xffffffffL) {    \


ksco:
Oh thank you for this, but I think changing `0xffffffff` to `0xffffffffL` should be enough here.

Alion-King:
```c
        } else if (_delta_ip >= -2048 && _delta_ip < 2048) {      \
            ADDI(xRIP, xRIP, _delta_ip);                          \
        } else if (_delta_ip < 0 && _delta_ip >= -0xffffffff) {   \
            MOV32w(scratch, -_delta_ip);                          \
            SUB(xRIP, xRIP, scratch);                             \
        } else if (_delta_ip > 0 && _delta_ip <= 0xffffffff) {    \
            MOV32w(scratch, _delta_ip);                           \
            ADD(xRIP, xRIP, scratch);                             \
        } else {      
```
using MOV32w represents that value  ranges from INT_MIN to INT_MAX. I thank INT_INT"-80000000L" and INT_MAX 0x7fffffff" may be better.

ksco:
No, it's an unsigned 32bit int.

Alion-King:
I see. changing 0xffffffff to 0xffffffffL is Ok.

ksco:
Can you change LoongArch counterpart too?

Alion-King:
OK , I have do  the work for rv64 and la64.

src/dynarec/la64/dynarec_la64_helper.h
```diff
@@ -1093,10 +1093,10 @@
         } else if (_delta_ip == 0) {                              \
         } else if (_delta_ip >= -2048 && _delta_ip < 2048) {      \
             ADDI_D(xRIP, xRIP, _delta_ip);                        \
-        } else if (_delta_ip < 0 && _delta_ip >= -0xffffffff) {   \
+        } else if (_delta_ip < 0 && _delta_ip >= -0x80000000L) {   \
```
ksco: You should change this to `-0xffffffffL` too.

src/dynarec/rv64/dynarec_rv64_helper.h
```diff
@@ -1138,10 +1138,10 @@
         } else if (_delta_ip == 0) {                              \
         } else if (_delta_ip >= -2048 && _delta_ip < 2048) {      \
             ADDI(xRIP, xRIP, _delta_ip);                          \
-        } else if (_delta_ip < 0 && _delta_ip >= -0xffffffff) {   \
+        } else if (_delta_ip < 0 && _delta_ip >= -0x80000000L) {   \
```
ksco: You should change this to `-0xffffffffL` too

src/dynarec/la64/dynarec_la64_helper.h
ptitSeb: or `-0xffffffffLL` to be safe

Alion-King: en， "-0xffffffffL" is better because of negative value translated to positive value.

PR #3400, opened 2026-01-22
[WRAPPER] wrapped ncurses curterm via set_curterm; add selinux symbol to (mkdir, find)

yzewei:
The following are fixes for running Docker containers on box64:

Runtime Environment:

  ***Hardware: loongarch64***
  
  **Kernel: 6.6.0-130.0.0.107.loongarch64**
  
  **box64: main**
  
 **x86 container: debian:sid-slim**
 
**Specific modifications:**

  **wrappedlibtinfo6**:
  
  GO added `_nc_basename`, `_nc_is_abs_path`, `_nc_rootname`, and `use_tioctl`.
  GOM added `_nc_tiparm`, `del_curterm`, `set_curterm`, and `setupterm`.
  DATAM added `cur_term`.
  The above modifications are due to errors encountered during the `clear/reset` translation runtime.
  
  `cur_term`, as a pointer to a TERMINAL type structure, needs to be encapsulated by DATAM. Because ncurses generally uses `reentrant` as a build parameter, the global symbol `cur_term` will not be updated. The actual pointer is stored in `_nc_prescreen._cur_term` or `SCREEN->_term`, and `cur_term` cannot be directly accessed. Therefore, a layout-matched TERMINAL shadow needs to be constructed.
  
  **wrappedselinux**:
  
  GO added `setfilecon` to fix missing symbols in the `find` function.
  GO added `context_free` `context_new` `context_str`  `context_type_get` `context_type_set` `getcon_raw` `getfscreatecon_raw`  `mode_to_security_class` `security_compute_create_raw` ...


yzewei:
PTAL！

ptitSeb:
For Global symbol like NCurses uses, they are some limitation with box64 wrapping... So I have added some hack to workaround those limitation. Take a look at `librarian/globalsymbols.c` to see how they are handled. You'll see that cur_term is already there. But indeed, it's not wrapped in tfinfo6... so strange.

You should not need to DATAM(cur_term), like for the other global variables... But what be missing probably is the appropriate calls to `my_updateGlobalTInfo();` and `my_checkGlobalTInfo()` at the right place, like with `my6_initscr()` for the ncurses6 version of it.

yzewei:
> For Global symbol like NCurses uses, they are some limitation with box64 wrapping... So I have added some hack to workaround those limitation. Take a look at `librarian/globalsymbols.c` to see how they are handled. You'll see that cur_term is already there. But indeed, it's not wrapped in tfinfo6... so strange.
> 
> You should not need to DATAM(cur_term), like for the other global variables... But what be missing probably is the appropriate calls to `my_updateGlobalTInfo();` and `my_checkGlobalTInfo()` at the right place, like with `my6_initscr()` for the ncurses6 version of it.

This can be resolved by pushing a new commit.

rebuild_wrappers.py
```diff
@@ -410,6 +410,14 @@ def add_symbol_name(funname: Union[str, None], funsname: Dict[ClausesStr, List[s
...
 							 or match("libegl", "libglxnvidia") \
 							 or match("glesv2", "libglxnvidia") \
+							 or match("libtinfo", "libncurses") \
+							 or match("libtinfo", "libncursesw") \
+							 or match("libtinfo", "libncurses6") \
+							 or match("libtinfo", "libncursesw6") \
+							 or match("libtinfo6", "libncurses") \
+							 or match("libtinfo6", "libncursesw") \
+							 or match("libtinfo6", "libncurses6") \
+							 or match("libtinfo6", "libncursesw6") \
```
ptitSeb: Are you sure you need that? libncurses and libtinfo are linked (libncuses load libtfinfo), so I doubt symbols appears in both library. Maybe the symbol in libncurses is not at the right place and should be in libtfinfo instead?

src/wrapped/wrappedlibtinfo6.c
```diff
@@ -64,6 +66,78 @@ EXPORT int my6_tputs(x64emu_t* emu, void* str, int affcnt, void* f)
...
+        case 1: return my->_nc_tiparm(expected, fmt, args[0]);
+        case 2: return my->_nc_tiparm(expected, fmt, args[0], args[1]);
+        case 3: return my->_nc_tiparm(expected, fmt, args[0], args[1], args[2]);
+        case 4: return my->_nc_tiparm(expected, fmt, args[0], args[1], args[2], args[3]);
+        case 5: return my->_nc_tiparm(expected, fmt, args[0], args[1], args[2], args[3], args[4]);
+        case 6: return my->_nc_tiparm(expected, fmt, args[0], args[1], args[2], args[3], args[4], args[5]);
+        case 7: return my->_nc_tiparm(expected, fmt, args[0], args[1], args[2], args[3], args[4], args[5], args[6]);
+        case 8: return my->_nc_tiparm(expected, fmt, args[0], args[1], args[2], args[3], args[4], args[5], args[6], args[7]);
+        case 9: return my->_nc_tiparm(expected, fmt, args[0], args[1], args[2], args[3], args[4], args[5], args[6], args[7], args[8]);
+        default: return my->_nc_tiparm(expected, fmt, args[0], args[1], args[2], args[3], args[4], args[5], args[6], args[7], args[8]);
```
ptitSeb: I don't think it's a good idea to just "give up" with default being the same as case 9. At least a a `printf_log(LOG_NONE, "Warning: my6__nc_tiparm with %d parameters not handled\n", n);` or something like that.

rebuild_wrappers.py
yzewei: You're right, I should remove the duplicate declarations.

```c
debian-optiplex7000:~/python-trans/Pynbl-QA/python_test$ readelf -Ws /usr/lib/x86_64-linux-gnu/libtinfo.so.6   | rg 'cur_term|use_tioctl|_nc_rootname'
   198: 0000000000015500    19 FUNC    GLOBAL DEFAULT   13 use_tioctl@@NCURSES6_TINFO_5.9.20150530
   264: 00000000000154d0    23 FUNC    GLOBAL DEFAULT   13 use_tioctl_sp@@NCURSES6_TINFO_5.9.20150530
   278: 0000000000035788     8 OBJECT  GLOBAL DEFAULT   24 cur_term@@NCURSES6_TINFO_5.0.19991023
   293: 000000000000fc30     9 FUNC    GLOBAL DEFAULT   13 _nc_rootname@@NCURSES6_TINFO_5.3.20021019
yzw@debian-optiplex7000:~/python-trans/Pynbl-QA/python_test$ readelf -Ws /usr/lib/x86_64-linux-gnu/libncurses.so.6 | rg 'cur_term|use_tioctl|_nc_rootname'
   102: 0000000000000000     0 OBJECT  GLOBAL DEFAULT  UND cur_term@NCURSES6_TINFO_5.0.19991023 (35)
```

src/wrapped/wrappedlibtinfo6.c
yzewei: The default value has been removed; the size limit is 9; a warning will be printed if the size exceeds the limit.

rebuild_wrappers.py
```diff
@@ -410,6 +410,10 @@ def add_symbol_name(funname: Union[str, None], funsname: Dict[ClausesStr, List[s
 							 or match("libegl", "glesv2") \
 							 or match("libegl", "libglxnvidia") \
 							 or match("glesv2", "libglxnvidia") \
+							 or match("libtinfo", "libncurses") \
+							 or match("libtinfo", "libncursesw") \
+							 or match("libtinfo", "libncurses6") \
+							 or match("libtinfo", "libncursesw6") \
```
ptitSeb: why is this still needed?

yzewei: sorry...

ptitSeb:
Thanks, merging this now.

PR #3410, opened 2026-01-26
[SYSINFO] Refactored sysinfo module

ksco:


src/os/sysinfo.c
```diff
@@ -0,0 +1,195 @@
...
+        } else if (strstr(line, "processor") != NULL) {
+            info->ncpu++;
+        }
+    }
+    if (info->ncpu) info->read_ncpu = 1;
+    fclose(f);
+
+lscpu:
+    if (!info->read_cpuname || !info->read_ncpu || !info->read_frequency) {
+        FILE* f = popen("lscpu", "r");
```
ptitSeb, line 83: there need to be a safeguard here, like in the previous code, with env. var. to prevent infinite recursion in case `lscpu` is an x86_64 version and box64 is used to run it.

src/os/sysinfo.c
```diff
@@ -0,0 +1,195 @@
...
+                    }
+                }
+            }
+        }
+        pclose(f);
+    }
+
+fallback:
+    if (!info->read_ncpu) {
+        FILE* f = popen("nproc", "r");
```
ptitSeb, line 123: same here

src/core.c
```diff
@@ -1030,7 +1031,9 @@ int initialize(int argc, const char **argv, char** env, x64emu_t** emulator, elf
         printf_log(LOG_INFO, "Bash detected, disabling banner\n");
         if (!BOX64ENV(nobanner)) {
             setenv("BOX86_NOBANNER", "1", 0);
-            setenv("BOX64_NOBANNER", "1", 0);
+            setenv("BOX64_NOBANNER", "0", 0);
```
ptitSeb: Why are you diabling NOBANNER in a section where NOBANNER should be enabled?
And why enabling log & sigsegv? this will potentialy mess-up all script shell redirection

ksco: sorry, I pushed some debugging code....

ksco:
Finally made the CI happy, please review again.

src/os/sysinfo.c
```diff
@@ -0,0 +1,202 @@
...
+                }
+            }
+        } else if (strstr(line, "processor") != NULL) {
+            info->ncpu++;
+        }
+    }
+    if (info->ncpu) info->read_ncpu = 1;
+    fclose(f);
+
+    if (getenv("BOX64_LSCPU_EXECUTED")) {
```
ptitSeb, line 77: that test should be after the lscpu label, not before.

ksco: oops, thanks!

PR #3413, opened 2026-01-26
[WRAPPER] Wrapped a few more symbols for libc

ksco:


src/wrapped/wrappedlibc_private.h
```diff
@@ -416,7 +416,9 @@ GOW(fgetsgent_r, iFSppLp)
 GO(fgetspent, pFS)
 GOW(fgetspent_r, iFSppLp)
 GOW(fgets_unlocked, pFpiS)
-//GO(__fgets_unlocked_chk, 
+#ifndef STATICBUILD
```
ptitSeb: ifdef cannot be asymetric in _private.h. Toy need a else with the same symbol commented instead.

ksco: lol okay I see, will do it later

PR #3414, opened 2026-01-26
[LA64_DYNAREC] Removed fastnan path on certain opcodes ([RV64_DYNAREC] too)

ksco:


ptitSeb:
There is the same issue with MINPS/MAXPS, but that can be for another PR.

ksco:
Ack

PR #3422, opened 2026-01-27
[WRAPPER] add libc/libselinux/libgmp missing symbols

yzewei:


src/wrapped/wrappedselinux.c
```diff
@@ -1,18 +1,210 @@
...
 #include "bridge.h"
 #include "librarian/library_private.h"
 #include "x64emu.h"
+#include "callback.h"
 
 const char* selinuxName = "libselinux.so.1";
 #define LIBNAME selinux
 
-#include "wrappedlib_init.h"
+#if defined(__has_include)
```
ptitSeb: Can you avoid including sekinux header,, by simply redifing the needed structures?

yzewei: ok

PR #3449, opened 2026-02-03
[WRAPPER] Wrapped more symbols for libgl

ksco:


src/wrapped/wrappedlibgl_private.h
```diff
@@ -2930,106 +2930,108 @@ GO(TexPageCommitmentARB, vFiiiiiiiii)
...
 GO(glXCreateGLXVideoSourceSGIX, pFpippip)
 GO(glXCreateNewContext, pFppipi)
-GO(glXCreatePbuffer, pFppp)
-GO(glXCreatePixmap, pFppp)
-GO(glXCreateWindow, pFpppp)
-GO(glXChooseVisual, pFpip)
+GO(glXCreatePbuffer, LFppp)
+GO(glXCreatePixmap, LFppLp)
+GO(glXCreateWindow, LFppLp)
+// GOM(glXChooseVisual, pFEpip)
```
ptitSeb, line 2955: you might need to use 
```
#define GL_GLEXT_PROTOTYPES 1
#include <GL/gl.h>
#include <GL/glext.h>
#include <GL/glx.h>

#pragma wrappers type_letters p Display*
#pragma wrappers type_letters p GC
#pragma wrappers type_letters p Visual*

```
in a custom header file to use wrapperhelper on libGL

There are many GOM here that should not, and where GO before...

ksco: Oh shoot, okay I will fix it

PR #3457, opened 2026-02-04
[WRAPPER] avoid running ldconfig (static x86)

yzewei:
In the x86 Fedora image: an RPM script (glibc-common) calls a statically linked x86 ldconfig, which Box64 cannot execute and displays some Lua error logs.
```
[16/16] Removing libgomp-0:15.2.1-5.fc43.x86_64                             100% |  11.0   B/s |   9.0   B |  00m01s
>>> Running %triggerin scriptlet: glibc-common-0:2.42-9.fc43.x86_64                                                 >>> Non-critical error in %triggerin scriptlet: glibc-common-0:2.42-9.fc43.x86_64                                   
>>> [RPM] lua script failed: [string "%transfiletriggerin(glibc-common-2.42-9.fc43.x86_64)"]:20: assertion failed!  
>>> Running %triggerpostun scriptlet: glibc-common-0:2.42-9.fc43.x86_64                                             
>>> Non-critical error in %triggerpostun scriptlet: glibc-common-0:2.42-9.fc43.x86_64                               
>>> [RPM] lua script failed: [string "%transfiletriggerpostun(glibc-common-2.42-9.fc43.x86_64)"]:20: assertion faile
Transaction failed: Rpm transaction failed.
```

### DIFF
--- a/src/wrapped/wrappedlibc.c
+++ b/src/wrapped/wrappedlibc.c
@@ -2711,6 +2711,9 @@ EXPORT int32_t my_execlp(x64emu_t* emu, const char* path)
 EXPORT int32_t my_posix_spawn(x64emu_t* emu, pid_t* pid, const char* fullpath,
     const posix_spawn_file_actions_t *actions, const posix_spawnattr_t* attrp,  char* const argv[], char* const envp[])
 {
+    // Some distros call a statically linked x86 ldconfig in scriptlets; skip it.
+    const char* base = fullpath ? strrchr(fullpath, '/') : NULL;
+    base = base ? base + 1 : fullpath;
     int self = isProcSelf(fullpath, "exe");
     int x64 = FileIsX64ELF(fullpath);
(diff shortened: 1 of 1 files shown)


ksco (changes requested):
This is not a good idea.

yzewei:
> This is not a good idea.

What would be a better approach, or a similar approach to Steam's?

ksco:
To Steam's what? A good idea would be to fix ldconfig.

yzewei:
> To Steam's what? A good idea would be to fix ldconfig.

Okay, I understand.

ksco:
For Docker usage, I think you can compile a dynamically linked x86_64 ldconfig and replace the original one with `-v`. Not sure about it, maybe worth a try?

yzewei:
> For Docker usage, I think you can compile a dynamically linked x86_64 ldconfig and replace the original one with `-v`. Not sure about it, maybe worth a try?

Yes, let me give it a try.

PR #3460, opened 2026-02-04
[WRAPPER] Fixed atfork handler

### LINKED ISSUE #3455: Possible issues in atfork handler registration and execution in wrappedlibc.c
While looking into the atfork handling logic, I noticed two behaviors that might indicate issues in how atfork handlers are registered and executed. Both of them are confusing me:

  1. `my___register_atfork` in `wrappedlibc.c` writes to `atforks[atfork_sz]` but never increments `atfork_sz`, so registered handlers appear to be silently lost.
  2. The parent/child atfork handler loops use `--i` instead of `++i`, so they would never execute even if handlers were registered correctly.

  Since atfork_sz is never incremented by `__register_atfork`, the broken loops are never reached with a nonzero count.

  Details:

  Missing increment in `__register_atfork`:

  src/wrapped/wrappedlibc.c (~line 2822) and src/wrapped32/wrappedlibc.c (~line 2612):
```c
  my_context->atforks[my_context->atfork_sz].prepare = (uintptr_t)prepare;
  my_context->atforks[my_context->atfork_sz].parent = (uintptr_t)parent;
  my_context->atforks[my_context->atfork_sz].child = (uintptr_t)child;
  my_context->atforks[my_context->atfork_sz].handle = handle;
  return 0;  // atfork_sz never incremented
```
  Compare with the working version in src/wrapped/wrappedlibpthread.c (~line 49):
```c
  int i = my_context->atfork_sz++;  // incremented here
  my_context->atforks[i].prepare = (uintptr_t)prepare;
```
  Wrong loop direction in parent/child handlers:
```c
  // these use --i but iterate forward (should be ++i)
  for (int i=0; i<my_context->atfork_sz; --i)
```
  Note: the prepare loop (reverse order, --i) is correct.
```c
  for (int i=my_context->atfork_sz-1; i>=0; --i)                                                                                                                                                            
      if(my_context->atforks[i].prepare)                                                                                                                                                                    
          EmuCall(emu, my_context->atforks[i].prepare);
```

  Affected locations:
  - src/emu/x64int3.c:62, 71
  - src/wrapped/wrappedlibc.c:601, 607
  - src/wrapped32/wrappedlibc.c:722, 728

  If my understanding is correct, a possible fix would be:

  - Increment `my_context->atfork_sz` after registering an atfork entry in both wrappedlibc.c files.
  - Change the parent/child loops from `--i` to `++i` in the locations listed above.

  If so, I'm happy to submit a PR with the fixes

devarajabc:
Increment atfork_sz after registering an entry in my___register_atfork and my32___register_atfork, matching the working pattern in wrappedlibpthread.c. Without the increment, each registration overwrites slot 0 and all prior handlers are lost.

The two bugs mask each other. Since atfork_sz is never incremented by __register_atfork, it stays 0, so the broken loops are never reached with a nonzero count. Programs that register atfork handlers via pthread_atfork (which calls wrappedlibpthread.c and does increment correctly) would hit Bug 2 though — their parent/child handlers silently never run.

Fixes #3455

### DIFF
--- a/src/emu/x64int3.c
+++ b/src/emu/x64int3.c
@@ -59,7 +59,7 @@ x64emu_t* x64emu_fork(x64emu_t* emu, int forktype)
         // error...
     } else if(v!=0) {  
         // execute atforks parent functions
-        for (int i=0; i<my_context->atfork_sz; --i)
+        for (int i=0; i<my_context->atfork_sz; ++i)
             if(my_context->atforks[i].parent)
                 EmuCall(emu, my_context->atforks[i].parent);
         if(forktype==3) {
@@ -68,7 +68,7 @@ x64emu_t* x64emu_fork(x64emu_t* emu, int forktype)
         }
     } else if(v==0) {
         // execute atforks child functions
-        for (int i=0; i<my_context->atfork_sz; --i)
+        for (int i=0; i<my_context->atfork_sz; ++i)
             if(my_context->atforks[i].child)
                 EmuCall(emu, my_context->atforks[i].child);
     }
--- a/src/wrapped/wrappedlibc.c
+++ b/src/wrapped/wrappedlibc.c
@@ -598,13 +598,13 @@ pid_t EXPORT my_fork(x64emu_t* emu)
         // error...
     } else if(v>0) {
         // execute atforks parent functions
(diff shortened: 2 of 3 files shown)


ptitSeb:
Thanks.

PR #3463, opened 2026-02-04
[SYSINFO] Fixed reentrancy handling

ksco:


src/os/sysinfo.c
```diff
@@ -79,11 +110,6 @@ static void readCpuinfo(sysinfo_t* info)
     fclose(f);
 
 lscpu:
-    if (getenv("BOX64_LSCPU_EXECUTED")) {
-        // in case it's the x86_64 lscpu, prevent infinite loop
-        goto fallback;
-    }
-    setenv("BOX64_LSCPU_EXECUTED", "1", 1);
     if (!info->read_cpuname || !info->read_ncpu || !info->read_frequency) {
         FILE* f = popen("lscpu", "r");
```
CRogers, line 114: Isn't this a bug? Previously, we used to set `BOX64_LSCPU_EXECUTED=1` before executing `lscpu` so the infinite loop would get prevented during the execution of the child `lscpu`. After this PR, we set `BOX64_SYSINFO_CACHED` right at the end of this function, _after `lscpu` is executed_, which fails to prevent the infinite loop!

PR #3465, opened 2026-02-05
[STEAM] Registered VERSION_CODENAME env. var. for better compatibility

ksco:


src/steam.c
```diff
@@ -269,6 +269,7 @@ void pressure_vessel(int argc, const char** argv, int nextarg, const char* prog)
             newargv[i-nextarg] = argv[i];
         }
     }
+    setenv("VERSION_CODENAME", "sniper", 1);
```
ptitSeb: but what if it's not sniper that is used?

PR #3471, opened 2026-02-05
Fix missing braces in x64run0f.c opcode 0x00 handler

devarajabc:
The else branch at line 89 was missing braces, causing the second switch statement to always execute regardless of rex.is32bits.


ptitSeb:
Right, thanks.

PR #3472, opened 2026-02-06
x87: Manually raise FE_OVERFLOW/FE_UNDERFLOW flags in LD2D conversion

yzewei:
**Problem Description (Motivation):** The Python NumPy software is throwing an error: A library relying on the `fenv` flag to detect precision loss failed in testing (`test_identity_long_short`).

The current `LD2D` (long double to double) conversion logic is implemented through integer bitwise operations. Because it doesn't involve host FPU hardware instructions, the CPU hardware doesn't automatically set the floating-point status register flags when a value overflows or underflows occurs.

**Result:** The NumPy exception detection issue has been fixed.

### DIFF
--- a/src/emu/x87emu_private.c
+++ b/src/emu/x87emu_private.c
@@ -3,6 +3,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <math.h>
+#include <fenv.h>
 
 #include "debug.h"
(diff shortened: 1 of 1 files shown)


ptitSeb:
It's a bit strange to update global status flags but not x87 status register but ok.

PR #3473, opened 2026-02-06
Fix undefined behavior in EXTRQ/INSERTQ

devarajabc:
Use 1LL instead of 1 in left shift for EXTRQ (0x78, 0x79) and INSERTQ (0x78, 0x79) mask generation. The shift amount (tmp8s+1) ranges 1-64 since tmp8s is masked with 0x3f (0-63), causing undefined behavior when shifting the 32-bit int literal 1 by more than 31.


ptitSeb:
I think `(1LL<<64)-1` is also UB, even if it should work on most architecture. A special case for tmp8s==63 might be needed.

ptitSeb:
LGTM

devarajabc:
I’m waiting for the CI because the environment preparation sometimes fails.

PR #3474, opened 2026-02-07
[WRAPPER] Fix atexit in the shared library failed to be called

xiangzhai:
Hi,

libAtExit.so (call `atexit` to register a function to be called at normal process termination) is not in the `DT_NEEDED` of atexit:

```
$ ldd atexit
	linux-vdso.so.1 =>  (0x00007ffc4aafd000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f5ff76e9000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f5ff731b000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f5ff78ed000)
```

Some good idea to replace the `closeAllDLOpened`?

Thanks,
Leslie Zhai

### DIFF
--- a/src/wrapped/wrappedlibdl.c
+++ b/src/wrapped/wrappedlibdl.c
@@ -497,7 +497,6 @@ int my_dlclose(x64emu_t* emu, void *handle)
     }
     --dl->dllibs[nlib].count;
     elfheader_t* h = GetElf(dl->dllibs[nlib].lib);
-    RunElfFini(h, emu);
     if((h && !h->gnuunique) || !h || actualy_closing)
         DecRefCount(&dl->dllibs[nlib].lib, emu);
     refreshTLSData(emu);


src/wrapped/wrappedlibdl.c
```diff
@@ -497,6 +497,7 @@ int my_dlclose(x64emu_t* emu, void *handle)
     }
     --dl->dllibs[nlib].count;
     elfheader_t* h = GetElf(dl->dllibs[nlib].lib);
+    RunElfFini(h, emu);
```
ptitSeb: The RunElfFini was not there because the DecRefCount will cann it when count is 0 and the library actualy unloaded.

src/core.c
```diff
@@ -681,8 +681,9 @@ void endBox64()
     printf_log(LOG_DEBUG, "Calling atexit registered functions (exiting box64)\n");
     CallAllCleanup(emu);
     printf_log(LOG_DEBUG, "Calling fini for all loaded elfs and unload native libs\n");
-    //void closeAllDLOpened();
-    //closeAllDLOpened();    // close residual dlopened libs. Disabled, seems like a bad idea, better to unload with proper dependancies
+    void closeAllDLOpened();
+    closeAllDLOpened();    // close residual dlopened libs. Disabled, seems like a bad idea, better to unload with proper dependancies
```
ptitSeb, line 685: closeAllDLOpened is supposed to be called when libdl in unloaded, so when it's count is 0 at unload of main elf

ptitSeb:
I cannot merge this PR. If a program DLOpen a lib 2 times for example, and unload it 1 time, the runelfini will be run, while it shouldn't as the lib is still in memory and should remain active.

xiangzhai:
Hi,

> If a program DLOpen a lib 2 times for example, and unload it 1 time

Yes, tests/misc/atexit just cover the case, it needs some good idea to replace the `closeAllDLOpened` indeed.

Thanks,
Leslie Zhai

PR #3480, opened 2026-02-09
Add missing ClearCache after ARCH_UDF

devarajabc:
MarkDynablock() and the always_test path in dynarec_native.c write ARCH_UDF to callret slots to trigger SIGILL for dirty-block detection, but don't flush the I-cache afterward. On ARM64/RV64/LA64 the I-cache and D-cache are not coherent, so the CPU keeps executing the stale instruction and the dirty check never fires.

InvalidDynablock() already calls ClearCache after the same ARCH_UDF write. Add the missing calls to match.


ptitSeb:
Indeed.

PR #3483, opened 2026-02-09
Fix 32-bit shift overflow in 64-bit BTR/BTC

devarajabc:
The interpreter path for LOCK BTR and BTC was using a 32-bit literal `1` for shifts. This caused UB when the shift amount was >= 32 in 64-bit mode (rex.w=1).

Replaced `1 << tmp8u` with `1LL << tmp8u` to correctly handle high bits.

### DIFF
--- a/src/emu/x64runf0.c
+++ b/src/emu/x64runf0.c
@@ -401,9 +401,9 @@ uintptr_t RunF0(x64emu_t *emu, rex_t rex, uintptr_t addr)
 #else
                     pthread_mutex_lock(&my_context->mutex_lock);
                     if(rex.w) {
-                        if(ED->q[0] & (1<<tmp8u)) {
+                        if(ED->q[0] & (1ULL<<tmp8u)) {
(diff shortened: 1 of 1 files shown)


devarajabc:
BTW, maybe we should use 1ULL (and 1U for the 32-bit paths) to avoid potential sign-bit UB ?

ptitSeb:
Technicaly, it should be `1ULL`.

devarajabc:
> Technicaly, it should be `1ULL`.

Ok, So i will sent other pr to replace all 1 to 1U and 1ULL for 64-bit path in the whole file ?
Or just in this one ?

ptitSeb:
Just this one. The `1U` is not usefull, I think the C compiler will infer the right type dirrectly.

devarajabc:
> Just this one. The `1U` is not usefull, I think the C compiler will infer the right type dirrectly.

ya, you are right. The `1U` is not usefull.

PR #3486, opened 2026-02-10
[LA64_DYNAREC] Close the fastnan path for MINSS/MAXSS/VMINSS/VMAXSS to correct NaN semantics.

yzewei:
**Problem:**
 LoongArch's `FMIN_S`/`FMAX_S` rules for determining NaN are inconsistent with x86's `MINSS`/`MAXSS`.

This causes NaN handling errors when `dynarec_fastnan=1` (error scenario observed: `numpy max`/`argmax` returns 1.0 instead of `nan`).

**Test Code:**
```
#gcc -O0 -mavx -o test test.c 
#include <immintrin.h>
#include <stdio.h>
#include <math.h>

int main() {
    float a = 1.0f;
    float b = NAN;

    __m128 va = _mm_set_ss(a);   // dest
    __m128 vb = _mm_set_ss(b);   // src
    __m128 vc = _mm_max_ss(va, vb); 

    float out;
    _mm_store_ss(&out, vc);
    printf("vmaxss out=%f\n", out);
    return 0;
}
```
If we need to check whether d0/d1 is NaN in order to preserve fastnan(FMAX_S), it might actually be slower.

### DIFF
--- a/src/dynarec/la64/dynarec_la64_avx_f3_0f.c
+++ b/src/dynarec/la64/dynarec_la64_avx_f3_0f.c
@@ -343,12 +343,8 @@ uintptr_t dynarec64_AVX_F3_0F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip,
             GETEYSS(v2, 0, 0);
             GETGYx_empty(v0);
             q0 = fpu_get_scratch(dyn);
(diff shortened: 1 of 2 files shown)


ptitSeb:
This is in par with ARM64 backend

PR #3489, opened 2026-02-11
Snapdragon 8 Elite Gen 5 optimization flags

airidosas252:
No idea if they are of any use in hindsight, but they might yield miniscule improvements.

Box64 compiled just fine with GCC 14.2.1 whilst using these flags.

Feel free to deny this PR if it's indeed useless.

### DIFF
--- a/docs/COMPILE.md
+++ b/docs/COMPILE.md
@@ -165,14 +165,22 @@ On a 64bit OS:
 
 On a 64bit OS:
 
+Targeting Snapdragon 845:
+
 ```
 -D SD845=1 -D CMAKE_BUILD_TYPE=RelWithDebInfo
 ```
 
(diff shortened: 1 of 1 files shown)


ptitSeb:
It's not useless, but indeed will not provide huge gain... I don't think the commit is enough tho, I'm pretty sure there are other places where the config name has to be added. You should check with some other existing arm64 profiles.

airidosas252:
Oops, should be fixed now.

ptitSeb:
Would you mind adding the entry in COMPILE.md also, so it's documented too.

airidosas252:
Entry now added. Please ignore the merge commit.

ptitSeb:
Thanks!

PR #3492, opened 2026-02-11
Fix incorrect bit shift in fpu_do_free

devarajabc:
The FPU tag word uses 2 bits per register. The previous implementation shifted the mask by 'i' instead of 'i * 2'.

### DIFF
--- a/src/emu/x87emu_private.c
+++ b/src/emu/x87emu_private.c
@@ -13,7 +13,7 @@
 
 void fpu_do_free(x64emu_t* emu, int i)
 {
-    emu->fpu_tags |= 0b11 << (i);   // empty
+    emu->fpu_tags |= 0b11 << (i*2);   // empty
     // check if all empty
     if(emu->fpu_tags != TAGS_EMPTY)
         return;


devarajabc:
It seems the workflow failed at the changed-files step due to a fetch depth issue. It might need fetch-depth: 0 in the checkout step to handle the history comparison correctly.

ptitSeb:
> It seems the workflow failed at the changed-files step due to a fetch depth issue. It might need fetch-depth: 0 in the checkout step to handle the history comparison correctly.

But why would that happens only on 2 seemingly random configurations?

devarajabc:
> > It seems the workflow failed at the changed-files step due to a fetch depth issue. It might need fetch-depth: 0 in the checkout step to handle the history comparison correctly.
> 
> But why would that happens only on 2 seemingly random configurations?

It passed on my own branch, but after opening the PR, it failed on 1 configuration, then 2 after a retry......
Maybe try other branch ?

devarajabc:
Thanks!

PR #3494, opened 2026-02-11
Fix incorrect dlinfo RTLD_DI_LINKMAP usage in initWrappedLib

kg233:
According to https://man7.org/linux/man-pages/man3/dlinfo.3.html and https://sourceware.org/git/?p=glibc.git;a=blob;f=dlfcn/dlinfo.c;hb=HEAD, dlinfo with RTLD_DI_LINKMAP expects a struct link_map **, not a struct link_map *.


ptitSeb:
Ah yes, indeed.

PR #3495, opened 2026-02-12
[ARM64_DYNAREC] Fix BICxw flag handling

devarajabc:
BICxw uses FEMIT which marks ARM64 NZCV flags as modified. However ARM64 BIC does NOT set NZCV — only BICS does.


ksco:
Okay I saw this was verified by ptitSeb, and it looks good to me too, let’s merge it in.

devarajabc:
Thanks

Issue #2389, opened 2025-02-19
[BUNDLE] Migrate libraries from git to the new bundle script


LukeShortCloud:
Library files currently in the top-level x64lib and x86lib directories should be migrated to now be part of the new x86 library bundle script.

Related: https://github.com/ptitSeb/box64/issues/2359

LukeShortCloud:
Here is a current inventory of what binary libraries are in the Box64 git repository compared to what is included in our [x86 library bundle](https://github.com/ptitSeb/box64/blob/main/docs/BUNDLE-X86-LIBS.md).

| [x86_64 Library](https://github.com/ptitSeb/box64/tree/v0.3.6/x64lib) | Bundled |
| --------------------- | ------------ |
| libcrypto.so.1.0.0 | No |
| libcrypto.so.1.1 | No |
| libgcc_s.so.1 | No |
| libmbedcrypto.so.3 (symlinked to 7) | No |
| libmbedcrypto.so.7 | No |
| libmbedtls.so.12 (symlinked to 14) | No |
| libmbedtls.so.14 | No |
| libmbedx509.so.0 (symlinked to 1) | No |
| libmbedx509.so.1 | No |
| libpng12.so.0 | No (only libpng16.so) |
| libssl.so.1.0.0 | No |
| libssl.so.1.1 | No |
| libstdc++.so.5 | No |
| libstdc++.so.6 | No |
| libunwind.so.8 | Yes |

| [x86 Library](https://github.com/ptitSeb/box64/tree/v0.3.6/x86lib) | Bundled |
| --------------------- | ------------ |
| libcrypto.so.1.0.0 | No |
| libcrypto.so.1.1 | No |
| libgcc_s.so.1 | No |
| libpng12.so.0 | No (only libpng16.so) |
| libssl.so.1.0.0 | No |
| libstdc++.so.5 | No |
| libstdc++.so.6 | No |
| libunwind.so.8 | Yes |
| libz.so.1 | No |

I also checked the Box86 git repository. It has the same but less libraries compared to Box64 (for Box32).

LukeShortCloud:
Up-to-date checklist:

- [x] libcrypto.so.1.0.0
- [x] libcrypto.so.1.1
- [x] libgcc_s.so.1
- [x] libmbedcrypto.so.3 (symlinked to 7) (x86_64 only)
- [x] libmbedcrypto.so.7 (x86_64 only)
[...]

PR #3503, opened 2026-02-12
[BUNDLE] Fix 'libstdc++.so.6' download

LukeShortCloud:
Related #2389


LukeShortCloud:
I'm not sure why this stopped working but these libraries were previously missing. They're back now.

```
$ ls -lah usr/lib/box64-x86_64-linux-gnu/ | grep -i libstdc
lrwxrwxrwx 1 user user   14 Feb 12 16:43 libstdc++.so -> libstdc++.so.6
lrwxrwxrwx 1 user user   19 Feb 12  2025 libstdc++.so.6 -> libstdc++.so.6.0.29
-rwxr-xr-x 1 user user 2.3M Feb 12  2025 libstdc++.so.6.0.29
$ ls -lah usr/lib/box64-i386-linux-gnu/
lrwxrwxrwx 1 user user   14 Feb 12 16:43 libstdc++.so -> libstdc++.so.6
lrwxrwxrwx 1 user user   19 Feb 12  2025 libstdc++.so.6 -> libstdc++.so.6.0.29
-rwxr-xr-x 1 user user 2.3M Feb 12  2025 libstdc++.so.6.0.29
```

LukeShortCloud:
Also, for completing the migration from git to the bundle, should I be including the real `libstdc++.so.5` or should that be a symlink? I see it's a real file here https://github.com/ptitSeb/box64/tree/v0.4.0/x64lib so I assume that's for a reason.

ptitSeb:
Yeah, I expect the lib to be different and the .6 not 100% retro-compatible with .5, hence the differents files.

Issue #2739, opened 2025-06-13
Turtle wow keeps disconnected after login


FrontMage:
Using winlator Winlator 10.0 (Hotfix) Mar 1 build with box64 0.3.2

Turtle wow can be found `https://turtle-wow.org/`

Box64 log as follow
```bash
Box64 with Dynarec v0.3.2  built on Feb 17 2025 15:44:18
Debug level is 1
Dynarec is on
Dynarec will continue block for 128 bytes on forward jump
Dynarec will not use weakbarrier on strong memory emulation
Dynarec will use only double for x87 emulation
Dynarec will try to generate x86 precise IEEE->int rounding
Dynarec will play it safe with x64 flags
Dynarec will not optimize CALL/RET
Dynarec will not use native flags if possible
Dynarec will print missing opcodes
BOX64 will not detect UnityPlayer.dll
Hack to force libX11 GLX extension present
Show Segfault signal even if a signal handler is present
Dynarec for ARM64, with extension: ASIMD AES CRC32 PMULL ATOMICS SHA1 SHA2 USCAT FLAGM FLAGM2 FRINT PageSize:4096 Running on Unknown CPU with 8 Cores
Will use Hardware counter measured at 19.2 MHz emulating 2.4 GHz
Box64 with Dynarec v0.3.2  built on Feb 17 2025 15:44:18
BOX64: Wine preloader detected, loading "/data/user/0/com.winlator/files/rootfs/opt/wine/bin/wine" directly
BOX64: Wine64 detected, WINEPRELOADRESERVE="000400000-000d26000"
BOX64: Didn't detect 48bits of address space, considering it's 39bits
WINE prereserve of 0x10000:0x8000 done
WINE prereserve of 0x110000:0x30000000 done
WINE prereserve of 0x7f000000:0x3000000 done
BOX64: Detected running wine with "WoW.exe"
Counted 57 Env var
BOX64 LIB PATH: BOX64 BIN PATH: ./:bin/:/data/user/0/com.winlator/files/rootfs/opt/wine/bin/:/data/user/0/com.winlator/files/rootfs/usr/local/bin/:/data/user/0/com.winlator/files/rootfs/usr/bin/
Looking for /data/user/0/com.winlator/files/rootfs/opt/wine/bin/wine
argv[1]="D:\twmoa_1172\WoW.exe"
Rename process to "wine"
Using native(wrapped) libdl.so.2
Using native(wrapped) libpthread.so.0
[...]

PR #3513, opened 2026-02-14
[BOX32] Fix ARPL (0x63) semantics and add regression test

FrontMage:
This PR fixes incorrect handling of opcode `0x63` in 32-bit mode (BOX32): it is `ARPL Ew, Gw` (not `MOVSXD`).

Changes:
- Implement correct ARPL semantics (conditional RPL update, ZF-only flag update) in interpreter and ARM64 dynarec.
- Fix opcode memory access decode for `0x63` in 32-bit mode.
- Add a `tests32/extensions/arpl` regression test.

This addresses: https://github.com/ptitSeb/box64/issues/2739

Fixes ptitSeb/box64#2739.


ptitSeb:
Oh, nice! Thanks for the analysis and fix!

PR #3515, opened 2026-02-15
Fix incrementing before mmap success check

devarajabc:
Move list->size increment after the mmap failure check. 
After mmap fails and returns 0, list->size has already been permanently incremented. 
On the next call to AllocDynarecMap, this loop iterates up to the new list->size, 
hitting the uninitialized list->chunks[i].

### DIFF
--- a/src/libtools/signals.c
+++ b/src/libtools/signals.c
@@ -64,7 +64,6 @@ x64_stack_t* sigstack_getstack() {
 }
 
 #ifndef DYNAREC
-typedef void dynablock_t;
 dynablock_t* FindDynablockFromNativeAddress(void* addr) {return NULL;}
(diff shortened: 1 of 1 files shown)


ptitSeb:
I understand the issue, but this was done that way to help with mutli-threaded access and avoid having competition when initializing the new list. Not sure this fix will handle heavy multi-threading well.

ptitSeb:
It would be safer to do the `int i = list->size++;` to after the potential point of failure, and just use `list->size` when trying to get the map size as it's readonly and not a critical information anyway.

devarajabc:
I see, you are right. The mmap failure path is nearly impossible to hit in practice. It's not worth it

devarajabc:
Maybe I can add a comment for it.

Issue #242, opened 2022-02-26
JIT for POWER?


madscientist159:
Just tried this with the UT2004 retail binary on a ppc64le box, and the main menu came up without any notable issues.  It's **extremely** slow and quite stuttery (far worse than QEMU), but I'm assuming that's mostly because of the missing JIT for the ppc64le architecture,

Would you be interested in porting the JIT to ppc64le?  It's the only open ISA with high performance CPUs on the market, and we'd be happy to provide a free development box in return for the effort.

Thank you!

ptitSeb:
So, RISCV64 Dynarec is in a good enough state now.

JeremyRand:
What would be needed at this point to move this forward? You mentioned hardware access -- what would be the obstacles to doing this with remote access?

TMinusBlastedRocket:
I have an IBM S824 (POWER8) I can give you root bare metal on

ptitSeb:
> I have an IBM S824 (POWER8) I can give you root bare metal on 24/7 (power company and internet outages excluded, im a random on the interwebs not a miracle worker)

I'm a bit busy for now, but lets talk about that beggining of next year, and see if we can make it work.

TMinusBlastedRocket:
> > I have an IBM S824 (POWER8) I can give you root bare metal on 24/7 (power company and internet outages excluded, im a random on the interwebs not a miracle worker)
> 
> I'm a bit busy for now, but lets talk about that beggining of next year, and see if we can make it work.

Cool

TMinusBlastedRocket:
> > I have an IBM S824 (POWER8) I can give you root bare metal on 24/7 (power company and internet outages excluded, im a random on the interwebs not a miracle worker)
> 
> I'm a bit busy for now, but lets talk about that beggining of next year, and see if we can make it work.

You still wanna do this?

ptitSeb:
> > > I have an IBM S824 (POWER8) I can give you root bare metal on 24/7 (power company and internet outages excluded, im a random on the interwebs not a miracle worker)
> > 
> > 
> > I'm a bit busy for now, but lets talk about that beggining of next year, and see if we can make it work.
> 
> You still wanna do this?

Not for the next 3 months I afraid, I'm still too busy to take on this in a reasonable way and timing.

madscientist159:
----- Original Message -----
> From: "ptitSeb" ***@***.***>
> To: "ptitSeb/box64" ***@***.***>
[...]

PR #3522, opened 2026-02-18
[CI] Add stale issues workflow

LukeShortCloud:


LukeShortCloud:
This workflow does the following:
- Runs once a week
- Checks for GitHub Issues that are 1 year or older without any replies
- If any are found, a "Stale" label is added to the issue. A message is also posted to warn that the issue is stale and will be closed soon
- If there are no updates for another 30 days, the issue is automatically closed
- We can add the label "long-term" to anything we care about keeping around
    - For example, issues like this we probably want to keep around: https://github.com/ptitSeb/box64/issues/242
    - We can also remove the "Stale" label which will keep it open for another year

You can read more about this official GitHub Action here: https://github.com/actions/stale

ptitSeb:
Ok, let's try this.

Issue #3529, opened 2026-02-23
Add libSDL3.so.0 wrapper (minimal, for OldUnreal Engine games)


runlevel5:
## Summary

OldUnreal's builds of Unreal Engine games (Unreal Gold, UT2004, and likely Unreal Tournament in the future) now link against **libSDL3.so.0** instead of SDL2. Box64 currently has SDL1 and SDL2 wrappers but no SDL3 wrapper, so these games fail to load:

```
[BOX64] Error loading needed lib libSDL3.so.0
[BOX64] Error loading one of needed lib
Error: Loading needed libs in elf ./ucc-bin-amd64
```

## Proposal

Add a **minimal libSDL3.so.0 wrapper** covering the ~72 functions actually used by these games. Analysis of all `.so` files across Unreal Gold and UT2004 shows they only use a small subset of SDL3's 1,270-function API surface:

### Functions needed (72 total)

**Window management (18):**
`SDL_CreateWindow`, `SDL_DestroyWindow`, `SDL_GetWindowFlags`, `SDL_GetWindowPosition`, `SDL_GetWindowSize`, `SDL_GetWindowSizeInPixels`, `SDL_GetWindowPixelFormat`, `SDL_GetWindowDisplayScale`, `SDL_GetWindowFullscreenMode`, `SDL_GetDisplayForWindow`, `SDL_RaiseWindow`, `SDL_RestoreWindow`, `SDL_SetWindowBordered`, `SDL_SetWindowFullscreen`, `SDL_SetWindowFullscreenMode`, `SDL_SetWindowIcon`, `SDL_SetWindowPosition`, `SDL_SetWindowSize`, `SDL_SetWindowTitle`, `SDL_SyncWindow`

**Display (7):**
`SDL_GetDisplays`, `SDL_GetDisplayName`, `SDL_GetDisplayBounds`, `SDL_GetCurrentDisplayMode`, `SDL_GetDesktopDisplayMode`, `SDL_GetFullscreenDisplayModes`, `SDL_GetPixelFormatName`

**OpenGL (12):**
[...]

PR #3530, opened 2026-02-23
[WRAPPER] Add SDL3 (libSDL3.so.0) wrapper for OldUnreal game support

runlevel5:
Add wrapper for SDL3 exposing 72 functions used by OldUnreal's Unreal Gold and UT2004 ports. 

`SDL_GL_GetProcAddress` uses `GOM` for GL proc address wrapping; all other functions are plain `GO` passthroughs. The `_private.h` lists all 1,271 SDL3 exported functions (1,199 commented out) per maintainer requirements. Uses ALTMY my3_ prefix to avoid collision with SDL1/SDL2 wrappers.

Ref: https://github.com/ptitSeb/box64/issues/3529


ptitSeb:
Can you try to run linux steam on your side? I'm afraid this might break it (an easy workaround would be to add libSDL3 in the list of "BOX64_PREFER_EMULATED" in `core.c`)

runlevel5:
I've added `libSDL3.so.0` to the `BOX64_PREFER_EMULATED` list in `src/core.c` as you suggested. This ensures apps shipping their own SDL3 (like Steam) will use the emulated version, while apps relying on the system SDL3 (like OldUnreal's ports) will still benefit from the native wrapper.

I also rebased onto latest `main` (picks up the EMMS opcode fixes) and regenerated the wrapper files correctly — the previous regeneration was missing `wrappedd3dadapter9_genvate.h` as input, which caused the `W` calling convention types to be dropped. This is now fixed; the generated files are purely additive with no removals.

I don't have access to a Linux Steam setup to test at the moment, but the `PREFER_EMULATED` addition should provide the safety net you're looking for.

ptitSeb:
LGTM, thanks!

PR #3536, opened 2026-02-23
[WRAPPER] Add libomp.so.5 wrapper mapping to native libgomp.so.1

runlevel5:
## Summary

- Adds a wrapper for `libomp.so.5` (LLVM OpenMP runtime) that maps to native `libgomp.so.1` (GCC OpenMP runtime)
- These two libraries are ABI-compatible for basic OpenMP runtime symbols
- Only 3 symbols are wrapped: `GOMP_parallel` (with callback trampoline), `omp_get_num_threads`, and `omp_get_thread_num`

## Motivation

OldUnreal's UT2004 build (v469e) links against `libomp.so.5` in its `Engine.so`. On systems where only `libgomp.so.1` is available natively (e.g. Fedora with GCC), this wrapper allows the game to run by transparently redirecting to the ABI-compatible GCC OpenMP runtime.

Tested on ppc64le (POWER9, Fedora 43) — UCC commandlet runner and UT2004 game binary both load and run correctly with this wrapper.

## Implementation Notes

- Uses `ALTMY my2_` prefix to avoid linker symbol collision with `wrappedgomp.c`'s existing `my_GOMP_parallel`
- Added `match("gomp", "libomp")` to `rebuild_wrappers.py` duplicate symbol allowlist (since both wrappers declare the same 3 symbols)
- `libompName` is set to `"libgomp.so.1"` so box64 dlopen's the native GCC runtime when an x86_64 binary requests `libomp.so.5`

Related: #3529


ptitSeb:
There are way more symbol in that library then just those 3 symbols. And what if the real libomp.so.5 is present in the system? why not using it first?

runlevel5:
> There are way more symbol in that library then just those 3 symbols. And what if the real libomp.so.5 is present in the system? why not using it first?

Yes, there are much more than 3 symbols. You get good point about using real libomp.so.5. I will make a new PR

PR #3543, opened 2026-02-24
[WRAPPER] Add libomp.so.5 wrapper

runlevel5:
Instead of piggy-backing on the libgomp like I did in https://github.com/ptitSeb/box64/pull/3536, this PR would wrap the whole lib


src/wrapped/wrappedlibomp_private.h
```diff
@@ -0,0 +1,833 @@
...
+//GO(kmpc_realloc,
+//GO(kmpc_set_affinity_mask_proc,
+//GO(kmpc_set_blocktime,
+//GO(kmpc_set_defaults,
+//GO(kmpc_set_disp_num_buffers,
+//GO(kmpc_set_library,
+//GO(kmpc_set_stacksize,
+//GO(kmpc_set_stacksize_s,
+//GO(kmpc_unset_affinity_mask_proc,
+//GO(omp_aligned_alloc, pFLLL)
```
ptitSeb: Why did you comment all those functions hat are fully defined?

src/wrapped/wrappedlibomp.c
```diff
@@ -0,0 +1,97 @@
...
+#include "x64emu.h"
+#include "emu/x64emu_private.h"
+#include "callback.h"
+#include "librarian.h"
+#include "box64context.h"
+#include "emu/x64emu_private.h"
+#include "myalign.h"
+
+// libomp.so.5 (LLVM OpenMP runtime) wraps against native libomp.so
+const char* libompName = "libomp.so";
```
ptitSeb, line 21: you can use
```C
#define ALTNAME "libgomp.so"
```
to have it fallback to libgomp if libomp is not available as you initialy intended.

runlevel5:
Thanks for the review feedback! I have two questions before applying the changes:

1. **Uncommenting fully-defined functions in `wrappedlibomp_private.h`**: When I uncomment all the `GOMP_*` and `omp_*` functions that have complete signatures, `rebuild_wrappers.py` fails with duplicate function errors because those same symbols are already declared in `wrappedgomp_private.h`. Should I:
   - Only uncomment functions unique to libomp (the `kmp_*`, `__kmp*`, `__kmpc*`, `kmpc_*`, `ompc_*` functions), leaving the `GOMP_*`/`omp_*` ones commented since the gomp wrapper already handles them?
   - Or is there another approach you had in mind?

2. **`#define ALTNAME "libgomp.so"`**: I'll add this — straightforward.

ptitSeb:
for 1. you need to add a new exception in `rebuild_wrappers.py`. There are a few others cases of duplicate symbol (like for sdl2 & sdl3).

runlevel5:
hmm the wrapper script now dropping W-types, I will revise what I did wrong here

ptitSeb:
Your PR seems to contains too many generated files for what sources have been changes.

runlevel5:
@ptitSeb You're right, sorry about that. I had a stale regeneration that bumped the version string (v2.2.0.18 → v2.5.0.24) in 9 unrelated generated files. I've reverted those and force-pushed — the PR now only contains the 12 libomp-specific files.

ptitSeb:
Thanks.

ptitSeb:
Side note: I think you enabled TERMUX and/or ANDROID in your local build, and I assume this is a mystake, espacialy on PowerPC architecture.

PR #3547, opened 2026-02-25
[LA64_DYNAREC] Fix call_n clobbering x64 arg regs in fast wrapper path

yzewei:
In the LA64 dynarec fast bridging path, the called `call_n()` function does not preserve the x64 caller's parameter register state.

For wrappers such as `pFp` (single GPR return), `A1` is always written back to `RDX`, and the call-overridden parameter registers (`RDI/RSI/RCX/R8/R9`) may leak into the emulated x64 state. This is semantically inconsistent with the slow path and can cause real workload crashes (e.g., `update-mime-database`, bus errors/SIGSEGV).

**Remediation:**

- In the fast path, save/restore the caller parameter registers around `call_n()`.

- Make the write-back operation of `RDX` dependent on `nret` (the count of the wrapper return register): only write `A1 -> RDX` if `nret > 1`; otherwise, restore the previous `RDX`.

### DIFF
--- a/box64-bundle-x86-libs.sh
+++ b/box64-bundle-x86-libs.sh
@@ -127,7 +127,7 @@ done
 # Manually create additional symlinks that are useful to have.
 cd "${dir_tmp}"/bundle-libs/usr/lib/box64-x86_64-linux-gnu/
 ln -s libmbedcrypto.so.7 libmbedcrypto.so.3
-ln -s libmbedcrypto.so.14 libmbedcrypto.so.12
+ln -s libmbedtls.so.14 libmbedtls.so.12
 ln -s libmbedx509.so.1 libmbedx509.so.0
 
 # generate the bundle libraries archive
--- a/src/dynarec/la64/dynarec_la64_00.c
+++ b/src/dynarec/la64/dynarec_la64_00.c
@@ -2647,7 +2647,27 @@ uintptr_t dynarec64_00(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                     if (tmp < 0 || (tmp & 15) > 1)
                         tmp = 0; // TODO: removed when FP is in place
                     if ((BOX64ENV(log) < 2 && !BOX64ENV(rolling_log)) && tmp) {
+                        int nret = tmp & 0xF;
+                        // Preserve caller-owned arg regs around the fast native wrapper call.
+                        ADDI_D(xSP, xSP, -64);
+                        ST_D(xRAX, xSP, 0);
+                        ST_D(xRDI, xSP, 8);
+                        ST_D(xRSI, xSP, 16);
+                        ST_D(xRDX, xSP, 24);
+                        ST_D(xRCX, xSP, 32);
+                        ST_D(xR8, xSP, 40);
+                        ST_D(xR9, xSP, 48);
                         call_n(dyn, ninst, (void*)(addr + 8), tmp);
+                        if (nret < 1)
+                            LD_D(xRAX, xSP, 0);
+                        LD_D(xRDI, xSP, 8);
+                        LD_D(xRSI, xSP, 16);
+                        if (nret < 2)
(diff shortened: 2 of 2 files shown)


yzewei:
Similar issues may exist in the RV64 and ARM64 code paths: the `call_n()` method in the fast wrapper path does not fully protect the x64 parameter registers after a native call, potentially leading to semantic inconsistencies with the slow path.

I do not have the corresponding hardware and therefore cannot reproduce the issue at runtime.

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -575,9 +575,16 @@ void call_n(dynarec_la64_t* dyn, int ninst, void* fnc, int w)
...
+    int nret = ((w > 0) ? w : -w) & 0xF;
     ST_D(xRSP, xEmu, offsetof(x64emu_t, regs[_SP]));
     ST_D(xRBP, xEmu, offsetof(x64emu_t, regs[_BP]));
     ST_D(xRBX, xEmu, offsetof(x64emu_t, regs[_BX]));
+    ST_D(xRDI, xEmu, offsetof(x64emu_t, regs[_DI]));
+    ST_D(xRSI, xEmu, offsetof(x64emu_t, regs[_SI]));
+    ST_D(xRCX, xEmu, offsetof(x64emu_t, regs[_CX]));
+    ST_D(xR8, xEmu, offsetof(x64emu_t, regs[_R8]));
+    ST_D(xR9, xEmu, offsetof(x64emu_t, regs[_R9]));
+    ST_D(xRDX, xEmu, offsetof(x64emu_t, regs[_DX]));
```
ksco: Why? It's caller-saved registers.

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -597,8 +604,17 @@ void call_n(dynarec_la64_t* dyn, int ninst, void* fnc, int w)
     // put return value in x64 regs
     if (w > 0) {
         MV(xRAX, A0);
+    }
+    if ((w > 0) && (nret > 1)) {
```
ksco: Why? RDX is a caller-saved registrer.

ksco (changes requested):

src/dynarec/la64/dynarec_la64_helper.c
yzewei: If the caller saves the data, then saving/restoring needs to be performed before and after calling call_n in dynarec_la64_00.c.

src/dynarec/la64/dynarec_la64_helper.c
yzewei: The same applies here; is register saving/restoring performed before and after calling call_n?

src/dynarec/la64/dynarec_la64_helper.c
ksco: CALLER-SAVED register.

ptitSeb:
Can you provide a simple way to reproduce the issue you are having?

ptitSeb:
The thing is: RDX is a Caller-saved register, as part of the System-V ABI. That means it's value will not be preserved upon the function call. So it is strange that you actually need to preserve it for your case to run. Maybe there is a speicfic function call that have specific ABI attributes that does needs RDX to be preserved, but in that case, which function?

yzewei:
> Can you provide a simple way to reproduce the issue you are having?


[tmp.zip](https://github.com/user-attachments/files/25541121/tmp.zip)

Unzip the zip package to obtain the x86 binary update-mime-database, and execute `box64 ./update-mime-database /usr/share/mime` to check for errors.

ksco:
image

works perfectly fine.

yzewei:
> image
> works perfectly fine.

I'm sorry, my other patches messed up the results.

ksco:
No problem, it happens sometimes ;)

PR #3549, opened 2026-02-25
[wrapped] Add some missing symbols

yzewei:
Solve the error of missing symbols in games like DNF


src/wrapped/wrappedlibpthread_private.h
```diff
@@ -229,8 +229,8 @@ GO(__pthread_rwlock_init, iFpp)
 GO(pthread_rwlock_init, iFpp)
 GO(__pthread_rwlock_rdlock, iFp)
 GO(pthread_rwlock_rdlock, iFp)
-//GO(pthread_rwlock_timedrdlock, 
-//GO(pthread_rwlock_timedwrlock, 
+GO2(pthread_rwlock_timedrdlock, iFpp, my_pthread_rwlock_timedrdlock)
```
ptitSeb: Mmmm, I don't understand what you are doing here? Why not using a simple GO for those 2?

yzewei: Indeed, GO2 is unnecessary.

src/wrapped/wrappedlibc_private.h
```diff
@@ -841,6 +841,7 @@ GO(inet_nsap_ntoa, pFipp)
 GO(inet_ntoa, pFu)
 GO(inet_ntop, pFippu)
 GOW(inet_pton, iFipp)
+GO2(__inet_pton_chk, iFippL, my___inet_pton_chk)
```
ptitSeb: Use a GOM, not a GO2

ptitSeb (approved):

PR #3551, opened 2026-02-25
[DYNAREC] Replace hardcoded jmpnext trampoline size with per-arch JMPNEXT_SIZE macro

### LINKED ISSUE #3544: Dynarec: hardcoded jmpnext trampoline size prevents architectures needing more than 2-3 instructions
### Context

I'm working on a PPC64LE (POWER9+) dynarec backend and ran into an issue with the jmpnext trampoline layout in the shared dynarec code.

### Problem

In `dynarec_native.c`, the jmpnext area size is hardcoded as `4*sizeof(void*)` (32 bytes) in about 10 places — in `CreateEmptyBlock`, `FillBlock64`, and the block layout comment. This 32-byte budget works for ARM64 (2 instructions), LA64 (3 instructions), and RV64 (3 instructions) because they all have either:

- A PC-relative load instruction (ARM64's `LDR literal`, LA64's `PCADDU12I`+`LD.D`, RV64's `AUIPC`+`LD`), and
- A direct branch-through-GPR instruction (`BR`)

PPC64LE lacks both of these. To load from a PC-relative address and branch indirectly, `CreateJmpNext` needs 5 instructions (20 bytes):

```asm
BCL   20,31,.+4     # get PC into LR (no AUIPC/LDR-literal equivalent)
MFLR  r12           # move LR to GPR
LD    r12,off(r12)  # load target address
MTCTR r12           # move to CTR (no direct branch-through-GPR)
BCTR                # branch via CTR
```

With 8 bytes for the self pointer + 20 bytes of code + 8 bytes for the target pointer = 36 bytes minimum, the 32-byte budget overflows.

### Possible solution

One approach I've taken in my local branch is to introduce a `JMPNEXT_SIZE` macro in `dynarec_arch.h` that each architecture defines:

```c
// ARM64, LA64, RV64 — unchanged behavior:
#define JMPNEXT_SIZE    (4*sizeof(void*))   // 32 bytes

// PPC64LE — needs extra room:
#define JMPNEXT_SIZE    (5*sizeof(void*))   // 40 bytes
```

Then the 10 hardcoded references in `dynarec_native.c` become:
- `4*sizeof(void*)` → `JMPNEXT_SIZE`
- `3*sizeof(void*)` → `JMPNEXT_SIZE - sizeof(void*)`
- `2*sizeof(void*)` → `JMPNEXT_SIZE - 2*sizeof(void*)`

For ARM64/LA64/RV64 this is a pure constant-for-constant substitution — identical compiled output, zero functional change.

### Question

Before I submit a PR for this, I wanted to ask:

1. Is the `JMPNEXT_SIZE` macro approach acceptable, or would you prefer a different mechanism (e.g. a function, a field in a struct, etc.)?
2. Should I submit just the refactoring (adding the macro + replacing hardcoded values for the 3 existing architectures) as a standalone PR, or would you prefer it bundled with the PPC64LE backend?
3. Are there any concerns about the jmpnext layout assumptions elsewhere that I might have missed? I noticed `arm64/updateflags_arm64.c` also references `3*sizeof(void*)` / `4*sizeof(void*)` directly, but since that's ARM64-only code it wouldn't need the macro.

Happy to adjust the approach based on your guidance. Thanks!

runlevel5:
## Summary

- Add `JMPNEXT_SIZE` macro to each architecture's section in `dynarec_arch.h`, all defined as `(4*sizeof(void*))` for ARM64, LA64, and RV64
- Replace all 10 hardcoded `4*sizeof(void*)`, `3*sizeof(void*)`, and `2*sizeof(void*)` jmpnext references in `dynarec_native.c` with expressions using `JMPNEXT_SIZE`
- Update the block layout comment to reflect the variable-size jmpnext area

## Problem

The jmpnext trampoline area was hardcoded as 32 bytes (`4*sizeof(void*)`) in ~10 places across `dynarec_native.c`. This works for ARM64/LA64/RV64 which fit their jmpnext code in 2-3 instructions, but prevents architectures that need more instructions (e.g. PPC64LE needs 5 instructions due to lacking PC-relative load and direct branch-through-GPR).

## Approach

Pure refactoring — no functional change. For all three existing architectures, `JMPNEXT_SIZE` evaluates to `4*sizeof(void*)`, producing identical compiled output. The `arm64/updateflags_arm64.c` references are left as-is since they're ARM64-specific code.

Fixes #3544

### DIFF
--- a/src/dynarec/dynarec_arch.h
+++ b/src/dynarec/dynarec_arch.h
@@ -39,6 +39,7 @@ extern void* create_updateflags();
 
 #define ARCH_NOP    0b11010101000000110010000000011111
 #define ARCH_UDF    0xcafe
+#define JMPNEXT_SIZE    (4*sizeof(void*))
 #elif defined(LA64)
 
 #define instruction_native_t        instruction_la64_t
@@ -69,6 +70,7 @@ extern void* create_updateflags();
 #define ARCH_UNALIGNED(A, B) 0
 extern uint32_t la64_crc(void* p, uint32_t len);
 #define ARCH_CRC(A, B)       return la64_crc(A, B)
+#define JMPNEXT_SIZE    (4*sizeof(void*))
 
 #elif defined(RV64)
 
@@ -101,6 +103,7 @@ extern uint32_t la64_crc(void* p, uint32_t len);
 #define ARCH_ADJUST(A, B, C, D) {}
 #define STOP_NATIVE_FLAGS(A, B) {}
 #define ARCH_UNALIGNED(A, B) arch_unaligned(A, B)
+#define JMPNEXT_SIZE    (4*sizeof(void*))
 #else
 #error Unsupported platform
 #endif
--- a/src/dynarec/dynarec_native.c
+++ b/src/dynarec/dynarec_native.c
@@ -381,7 +381,7 @@ uintptr_t native_pass2(dynarec_native_t* dyn, uintptr_t addr, int alternate, int
 uintptr_t native_pass3(dynarec_native_t* dyn, uintptr_t addr, int alternate, int is32bits, int inst_max);
 
 dynablock_t* CreateEmptyBlock(uintptr_t addr, int is32bits, int is_new) {
-    size_t sz = 4*sizeof(void*) + sizeof(dynablock_t);
+    size_t sz = JMPNEXT_SIZE + sizeof(dynablock_t);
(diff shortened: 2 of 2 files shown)


ptitSeb:
Seing how you are very active on box64 that last 2 weeks, can I ask in what context you are porting the Dynarec to PowerPC (not that I am complaining, I am just -very- curious)?

Issue #20, opened 2021-07-04
Cinebench R15


Heasterian:
Cinebench R15 crashes at start of CPU benchmark with missing opcode.

```
7235|0x5b29894: Unimplemented Opcode (F9) 0F C3 45 00 E8 93 98 FF FF 48 83 C6 04 48 89
```

theofficialgman:
unrelated to cinebenchr15 but also missing this opcode for geekbench5 to run
```14583|0x1c37966: Unimplemented Opcode (D2) 64 FF 50 10 89 C0 48 89 45 20 C7 45 0C 00 00```

ptitSeb:
@theofficialgman please don't hijack a ticket for something not related. Opcodes are quite different, there is no reason to regroup missing opcode inside a single ticket...

Heasterian:
As Cinebench R15 is working now, I'm closing issue.

theofficialgman:
> @theofficialgman please don't hijack a ticket for something not related. Opcodes are quite different, there is no reason to regroup missing opcode inside a single ticket...

my apologies, I will test the latest version today to make sure 2010f8d fully solved geekbench5

Would you rather have a discussion tab/main issue documenting all the unsupported opcodes or is this not necessary since you already know which ones aren't supported?

ptitSeb:
I prefer issues grouped per apps / games.

PR #3557, opened 2026-02-25
[DYNAREC] Fix protectDB/protectDBJumpTable stripping PROT_WRITE from data regions on large-page hosts (fix #3556)

### LINKED ISSUE #3556: Dynarec: handle mixed code+data host pages on systems with >4KB page sizes in protectDB/protectDBJumpTable
Following the ELF loader large-page fix in PR #3546, I've been hitting another class of issues on my PPC64LE system (64KB pages) related to how `protectDB()` and `protectDBJumpTable()` remove `PROT_WRITE` from pages containing dynarec blocks.

**The problem:**

On hosts with page sizes larger than 4KB (e.g. 64KB on POWER9, or 16KB/64KB on some AArch64 configs), a single host page can contain both executable code tracked by the dynarec *and* writable data buffers used by the application. When `protectDB()`/`protectDBJumpTable()` call `mprotect()` to remove `PROT_WRITE` from these mixed pages, it also strips writability from the data regions. This causes kernel syscalls like `read()` to return `EFAULT` when their target buffer happens to land on one of these pages — the kernel can't write to a non-writable page, and unlike userspace writes, this can't be caught and handled via SEGV.

On 4KB page systems this doesn't happen because the box64 internal page tracking granularity matches the host page size, so code and data don't share pages.

**Proposed fix:**

Add a helper function that checks whether a host page contains any writable non-dynarec regions outside the range being protected. If it does (mixed page), set `PROT_NEVERCLEAN` instead of calling `mprotect()`, which puts the block into `always_test` mode — the dynarec will re-validate the block's hash on each entry rather than relying on write-protection traps. This is slower but correct.

The check is gated behind `if(box64_pagesize > 4096)`, so on standard 4KB systems the `else` branch runs the original `mprotect()` path unchanged.

```c
if(box64_pagesize > 4096) {
    uintptr_t host_page = cur & ~(box64_pagesize-1);
    if(hostPageHasExternalWrite_locked(host_page, addr, addr+size)) {
        prot |= PROT_NEVERCLEAN;
    } else {
        mprotect((void*)cur, bend-cur, prot&~PROT_WRITE);
    }
} else {
    mprotect((void*)cur, bend-cur, prot&~PROT_WRITE);
}
```

**Risk:** Low — only triggers on hosts with >4KB pages, and the fallback (`PROT_NEVERCLEAN` / `always_test`) is an existing, well-tested code path already used for hotpages. Zero impact on 4KB systems.

This is a companion to the ELF loader fix in #3546 — together they address the two main places where large host pages cause issues.

Happy to open a PR for this if you'd like.

runlevel5:
## Summary

Fixes #3556.

On systems with host page sizes larger than 4KB (e.g. 64KB on PPC64LE POWER9, or 16KB/64KB on some AArch64 configs), `protectDB()` and `protectDBJumpTable()` call `mprotect()` to remove `PROT_WRITE` for dynarec write-protection. This can inadvertently strip writability from **data** regions sharing the same host page. Kernel syscalls like `read()` then fail with `EFAULT` because they cannot write to the now non-writable buffer page (unlike userspace writes, this cannot be caught via SEGV).

## Changes

- Add `hostPageHasExternalWrite_locked()` helper that scans all sub-ranges within a host page to detect "mixed" code+data pages where some writable sub-ranges are outside the dynarec-protected range
- In `protectDB()` and `protectDBJumpTable()`: when `box64_pagesize > 4096` and a mixed page is detected, use `PROT_NEVERCLEAN` (`always_test` mode with hash-based validation) instead of `mprotect()`
- Fix `dynablock.c` `always_test` handling: when `always_test==1` (NEVERCLEAN), skip the `protectDB()` call since on mixed pages it would either redundantly set `PROT_NEVERCLEAN` or counterproductively call `mprotect()`. Only re-protect when `always_test==2` (hotpage-based)

## Safety

- All changes gated by `if(box64_pagesize > 4096)` — **zero impact on 4KB page systems**
- No architecture-specific `#ifdef` — portable to any large-page host
- Build tested on PPC64LE POWER9 (Fedora 43, 64KB pages)
- Smoke tested with jq, sqlite3, busybox

### DIFF
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -2194,6 +2194,31 @@ uintptr_t getJumpAddress64(uintptr_t addr)
     #endif
 }
 
+// Helper: check if any sub-range in a host page has PROT_WRITE that is NOT part of the
+// dynarec-protected range [prot_start, prot_end). This detects mixed code+data host pages
+// on systems with large pages (e.g. 64KB) where mprotect to remove PROT_WRITE would also
+// strip writability from data regions, causing EFAULT on kernel writes (e.g. read() syscall).
+// Must be called with LOCK_PROT() held.
+static int hostPageHasExternalWrite_locked(uintptr_t host_page, uintptr_t prot_start, uintptr_t prot_end)
+{
+    uintptr_t scan = host_page;
+    uintptr_t host_end = host_page + box64_pagesize;
+    while(scan < host_end) {
+        uint32_t p = 0;
+        uintptr_t bend = 0;
+        rb_get_end(memprot, scan, &p, &bend);
+        if(bend > host_end)
+            bend = host_end;
+        if(p && (p & PROT_WRITE) && !(p & PROT_DYN)) {
+            if(scan < prot_start || bend > prot_end) {
+                return 1;
+            }
+        }
+        scan = bend;
+    }
+    return 0;
+}
+
 // Remove the Write flag from an adress range, so DB can be executed safely
 void protectDBJumpTable(uintptr_t addr, size_t size, void* jump, void* ref)
 {
@@ -2216,8 +2241,19 @@ void protectDBJumpTable(uintptr_t addr, size_t size, void* jump, void* ref)
         if(!(dyn&PROT_NEVERPROT)) {
             prot&=~PROT_CUSTOM;
             if(prot&PROT_WRITE) {
-                if(!dyn) 
-                    mprotect((void*)cur, bend-cur, prot&~PROT_WRITE);
+                if(!dyn) {
+                    if(box64_pagesize > 4096) {
+                        uintptr_t host_page = cur & ~(box64_pagesize-1);
+                        if(hostPageHasExternalWrite_locked(host_page, addr, addr+size)) {
+                            dynarec_log(LOG_INFO, "protectDBJumpTable: mixed code+data host page %p, using always_test instead of mprotect\n", (void*)host_page);
+                            prot |= PROT_NEVERCLEAN;
+                        } else {
+                            mprotect((void*)cur, bend-cur, prot&~PROT_WRITE);
+                        }
+                    } else {
+                        mprotect((void*)cur, bend-cur, prot&~PROT_WRITE);
+                    }
+                }
                 prot |= PROT_DYNAREC;
             } else 
                 prot |= PROT_DYNAREC_R;
@@ -2253,8 +2289,24 @@ void protectDB(uintptr_t addr, uintptr_t size)
         if(!(dyn&PROT_NEVERPROT)) {
             prot&=~PROT_CUSTOM;
             if(prot&PROT_WRITE) {
-                if(!dyn) 
-                    mprotect((void*)cur, bend-cur, prot&~PROT_WRITE);
+                if(!dyn) {
+                    // On large-page systems, check if removing PROT_WRITE from this host page
+                    // would also affect writable data regions sharing the same host page.
+                    // If so, use PROT_NEVERCLEAN (always_test mode) instead of mprotect,
+                    // because kernel syscalls (e.g. read()) cannot be caught via SEGV and
+                    // would return EFAULT if the buffer is on a non-writable page.
+                    if(box64_pagesize > 4096) {
+                        uintptr_t host_page = cur & ~(box64_pagesize-1);
+                        if(hostPageHasExternalWrite_locked(host_page, addr, addr+size)) {
+                            dynarec_log(LOG_INFO, "protectDB: mixed code+data host page %p, using always_test instead of mprotect\n", (void*)host_page);
+                            prot |= PROT_NEVERCLEAN;
+                        } else {
+                            mprotect((void*)cur, bend-cur, prot&~PROT_WRITE);
+                        }
+                    } else {
+                        mprotect((void*)cur, bend-cur, prot&~PROT_WRITE);
+                    }
+                }
                 prot |= PROT_DYNAREC;
             } else 
                 prot |= PROT_DYNAREC_R;
--- a/src/dynarec/dynablock.c
+++ b/src/dynarec/dynablock.c
@@ -404,9 +404,12 @@ dynablock_t* DBGetBlock(x64emu_t* emu, uintptr_t addr, int create, int is32bits)
             } else {
                 dynarec_log(LOG_DEBUG, "Validating block %p from %p:%p (hash:%X, always_test:%d) for %p\n", db, db->x64_addr, db->x64_addr+db->x64_size-1, db->hash, db->always_test, (void*)addr);
                 if(db->always_test) {
-                    if(db->always_test==2)
+                    if(db->always_test==2) {
                         db->always_test = 0;
-                    protectDB((uintptr_t)db->x64_addr, db->x64_size);
+                        protectDB((uintptr_t)db->x64_addr, db->x64_size);
+                    }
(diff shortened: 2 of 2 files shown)


xiangzhai:
Hi @runlevel5 

regression for LA64 16K pagesize:

```
      Start 20: longjumpInSignals_nocosim
20/57 Test #20: longjumpInSignals_nocosim ...........***Failed    0.06 sec
```

```
BOX64_LOG=0 ./box64 ../tests/test21 2>&1|tee after.log
diff -Naur ../tests/ref21.txt after.log
```

diff:

```
-si_addr: 0, si_code: 128, si_errno: 0 RIP offset: 0 TRAPERR=362 TRAPNO=13
+si_addr: 0, si_code: 128, si_errno: 0 RIP offset: 0 TRAPERR=0 TRAPNO=13
```

Thanks,
Leslie Zhai

ptitSeb:
> Hi @runlevel5
> 
> regression for LA64 16K pagesize:
> 
> ```
>       Start 20: longjumpInSignals_nocosim
> 20/57 Test #20: longjumpInSignals_nocosim ...........***Failed    0.06 sec
> ```
> 
> ```
> BOX64_LOG=0 ./box64 ../tests/test21 2>&1|tee after.log
> diff -Naur ../tests/ref21.txt after.log
> ```
> 
> diff:
> 
> ```
> -si_addr: 0, si_code: 128, si_errno: 0 RIP offset: 0 TRAPERR=362 TRAPNO=13
> +si_addr: 0, si_code: 128, si_errno: 0 RIP offset: 0 TRAPERR=0 TRAPNO=13
> ```
> 
> Thanks, Leslie Zhai

Which test of test21.c is failling here?

xiangzhai:
Hi @ptitSeb 

The third one (from line 70 to 74) of test21.c failing:

```
void test_cc()
{
        memcpy(exec_p, buff_cc, sizeof(buff_cc));
        if(!sigsetjmp(context_buf, 1)) {
                vFv_t f = (vFv_t)exec_p;
                f();
        }
        memcpy(exec_p, buff_cd03, sizeof(buff_cd03));
        if(!sigsetjmp(context_buf, 1)) {
                vFv_t f = (vFv_t)exec_p;
                f();
        }
 =>       memcpy(exec_p, buff_cd2d, sizeof(buff_cd2d));
        if(!sigsetjmp(context_buf, 1)) {
                vFv_t f = (vFv_t)exec_p;
                f();
 =>       }
        if(!sigsetjmp(context_buf, 1)) {
                vFv_t f = (vFv_t)NULL;
                f();
        }
}
```

Thanks,
Leslie Zhai

ptitSeb:
> Hi @ptitSeb
> 
> The third one (from line 70 to 74) of test21.c failing:
> 
> ```
> void test_cc()
> {
>         memcpy(exec_p, buff_cc, sizeof(buff_cc));
>         if(!sigsetjmp(context_buf, 1)) {
>                 vFv_t f = (vFv_t)exec_p;
>                 f();
>         }
>         memcpy(exec_p, buff_cd03, sizeof(buff_cd03));
>         if(!sigsetjmp(context_buf, 1)) {
>                 vFv_t f = (vFv_t)exec_p;
>                 f();
>         }
>  =>       memcpy(exec_p, buff_cd2d, sizeof(buff_cd2d));
>         if(!sigsetjmp(context_buf, 1)) {
>                 vFv_t f = (vFv_t)exec_p;
>                 f();
>  =>       }
>         if(!sigsetjmp(context_buf, 1)) {
>                 vFv_t f = (vFv_t)NULL;
>                 f();
>         }
> }
> ```
> 
> Thanks, Leslie Zhai

Ok. So my guess is: this "exec_p" buffer gets a "NEVER_CLEAN" attribute in the porcess, but the CRC doesn't change or something similar between the buff_cd03 & buff_cd2d code copy? mmm, no it would still trigger some error. This is strange.

ptitSeb:
I reproduce the issue on my side

runlevel5:
@xiangzhai would you be able to grant me remote access to your box so I could do a bit troubleshooting?

runlevel5:
I am going to try to re-produce on my side too. Just need to compile the kernel pagesize set to 16k on my ARM64 Linux

ptitSeb:
Ok, I have pushed a fix, it should be ok now.

Note that it wasn't a regression but latent an issue present in the dynarec.

runlevel5:
I've tested with ARM64 Linux 16K and can confirm the issue does not impact ARM64

xiangzhai:
Hi @runlevel5 

Multi-threading might unsafe?

```
diff --git a/src/custommem.c b/src/custommem.c
index a35e2822e..83a352f24 100644
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -2199,6 +2199,8 @@ static int hostPageHasExternalWrite_locked(uintptr_t host_page, uintptr_t prot_s
 {
     uintptr_t scan = host_page;
     uintptr_t host_end = host_page + box64_pagesize;
+    LOCK_PROT();
+    int result = 0;
     while(scan < host_end) {
         uint32_t p = 0;
         uintptr_t bend = 0;
@@ -2207,12 +2209,16 @@ static int hostPageHasExternalWrite_locked(uintptr_t host_page, uintptr_t prot_s
             bend = host_end;
         if(p && (p & PROT_WRITE) && !(p & PROT_DYN)) {
             if(scan < prot_start || bend > prot_end) {
-                return 1;
+                result = 1;
+                goto unlock;
             }
         }
         scan = bend;
     }
-    return 0;
+
+unlock:
+    UNLOCK_PROT();
+    return result;
 }
 
 // Remove the Write flag from an adress range, so DB can be executed safely
```

Thanks,
Leslie Zhai

runlevel5:
@xiangzhai let me review the code to confirm your concern, do you have any application in mind that breaks?

runlevel5:
@xiangzhai I have reviewed the changes. It is threadsafe. This change would actually introduce a deadlock.
The `_locked` suffix on `hostPageHasExternalWrite_locked` is a convention meaning "the caller must already hold the lock."
Please refer to line 2197.

Both call sites already acquire `LOCK_PROT()` before calling this function:

- `protectDB() — LOCK_PROT()` at line 2274, call at line 2296
- `protectDBJumpTable() — LOCK_PROT()` at line 2226, call at line 2243

`mutex_prot` is not a recursive mutex — it's initialized with `PTHREAD_MUTEX_ERRORCHECK` (line 2979):

```
pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_ERRORCHECK);
pthread_mutex_init(&mutex_prot, &attr);
```

And with `USE_CUSTOM_MUTEX` it's a `uint32_t` spinlock that also doesn't support re-entry.

Adding `LOCK_PROT()` inside `hostPageHasExternalWrite_locked` would attempt to lock `mutex_prot` a second time from the same thread, which:

- With `PTHREAD_MUTEX_ERRORCHECK`: returns `EDEADLK` (and mutex_lock doesn't check the return value, so it proceeds without the lock — undefined behavior)
- With the custom spinlock: hangs forever since `native_lock_storeifnull_d` spins when the value is non-zero

You can verify this empirically with a [minimal reproducer](https://gist.github.com/runlevel5/f5fd7f3085524f9e1a19fa08b6f3f3d7)

xiangzhai:
Hi @runlevel5 

> And with USE_CUSTOM_MUTEX it's a uint32_t spinlock that also doesn't support re-entry.

Thanks for your hint!

Leslie Zhai

PR #3561, opened 2026-02-25
[BUNDLE] Add manual symlinks first

LukeShortCloud:
before doing automatic symlinks of the base '.so' files. This fixes the ordering of symlinks.


LukeShortCloud:
I noticed some odd things like `libmbedtls.so.14` being symlinked down to `libmbedtls.so.12` (it's backwards). This fixes that.

Related to that specific library, I also noticed that `libmbedtls.so.12` eventually points to `libmbedtls.so.2.28.8` which looked odd. However, that's how it is packaged upstream. I guess the library is very backwards compatible.

PR #3566, opened 2026-02-26
[BOX32] Fix ELF loading on 64KB page systems

runlevel5:
On systems with 64KB pages (e.g. POWER9), box32 fails to load i386 ELF binaries because the typical load address `0x8048000` is not 64KB-aligned. The `mmap64` call with `MAP_FIXED` at that address returns `EINVAL` since the kernel requires page-aligned addresses.

The existing code tried to handle this case after the mmap, but the mmap itself already failed before reaching that logic. This fix checks alignment upfront — when `vaddr` is not page-aligned, it rounds down the mmap address to the page boundary and increases the mapping size to compensate, then sets `image` to the actual vaddr within the mapping.

Tested on a POWER9 system (Fedora 43, ppc64le, 64KB pages) with Grim Fandango Remastered (a 32-bit i386 GOG game). Before this fix, the game immediately failed with:

```
Error: cannot create memory map (@0x8048000 0x3239454) for elf "GrimFandango" error=22/Invalid argument
```

After the fix, the game loads successfully — all i386 libraries resolve, OpenGL context is created, and the game runs.

## Related PR

https://github.com/ptitSeb/box64/pull/3546

### DIFF
--- a/src/elfs/elfloader32.c
+++ b/src/elfs/elfloader32.c
@@ -158,13 +158,19 @@ int AllocLoadElfMemory32(box64context_t* context, elfheader_t* head, int mainbin
         raw = mmap64(from_ptrv(offs), sz, 0, MAP_ANONYMOUS|MAP_PRIVATE|MAP_NORESERVE|MAP_FIXED, -1, 0);
         image = (void*)(((uintptr_t)raw+max_align)&~max_align);
     } else {
-        image = raw = mmap64(from_ptrv(head->vaddr), sz, 0, MAP_ANONYMOUS|MAP_PRIVATE|MAP_NORESERVE|MAP_FIXED, -1, 0);
         if(from_ptr(head->vaddr)&(box64_pagesize-1)) {
-            // load address might be lower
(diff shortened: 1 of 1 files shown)


xiangzhai:
Hi @runlevel5 

> Tested on a POWER9 system (Fedora 43, ppc64le, 64KB pages) with Grim Fandango Remastered (a 32-bit i386 GOG game).

Cool~ And I will check Linux steam for LA64 (AOSC 16K pagesize) this weekend. I often use 4K pagesize for running Linux steam.

Thanks,
Leslie Zhai

xiangzhai:
Hi @runlevel5 

> And I will check Linux steam for LA64 (AOSC 16K pagesize) this weekend.

LA64 AOSC 16K pagesize failed to run Linux steam, changed to 4K pagesize worked.

Thanks,
Leslie Zhai

ptitSeb:
Yeah, I don't think stemawebhelper will load if it detect pagesize different than 4K, like with most CEF based software. (remember that box64 doesn't emulate a 4k pagesize for now).

PR #3571, opened 2026-02-26
Detect exit() from coroutine/fiber alternate stack and convert to emu quit

zqb-all:
When exit() is called with RSP outside the known main stack range (emu->init_stack/size_stack), convert to emu->quit instead of killing the process. This handles custom coroutine libraries that call exit() as a safety fallback from alternate stacks.

### DIFF
--- a/src/wrapped/wrappedlibc.c
+++ b/src/wrapped/wrappedlibc.c
@@ -4099,6 +4099,19 @@ EXPORT void my_exit(x64emu_t* emu, int code)
         emu->flags.quitonexit = 2;
         return;
     }
+    if(emu->size_stack && emu->init_stack) {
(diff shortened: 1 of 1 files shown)


ptitSeb:
I am wondering if that code
```C
emu->quit = 1;
R_EAX = code;
``` 
should be how my_exit should be handled all the time. I'll do some testing on my side before merging this one.

ptitSeb:
I'm closing this PR. This is not the right way to fix the root cause of the issue.

zqb-all:
Yep, totally agree this isn't a root-cause fix. It's essentially a WeMeet bug—hopefully future wemeet releases will catch and fix it.

PR #3574, opened 2026-02-26
Disable PCLMULQDQ for WeMeet QtWebEngineProcess to fix dynarec crash

zqb-all:
WeMeet login QR code display triggers a dynarec diff at PCLMULQDQ instruction, causing SIGILL and abort in libQt5WebEngineCore.so. Disable PCLMULQDQ emulation for QtWebEngineProcess as a workaround.


ksco:
If it's an opcode implementation issue, we'd better fix it instead of bypassing it.

zqb-all:
Yes, I think it's an implementation issue. I hope I'll have time to look into it further next week.

PR #3578, opened 2026-02-27
[WRAPPER] Fix SDL Vulkan bridge2 passing wrong fnc2 (regression from c7bc29329)

runlevel5:
## Summary

This fixes a regression introduced in commit c7bc29329 ("[WRAPPER] Add/Fix SDL_Vulkan_GetVkGetInstanceProcAddr wrapped function from SDL2/SDL3 to use bridge2") that breaks Vulkan in any SDL2/SDL3 game using `SDL_Vulkan_GetVkGetInstanceProcAddr`.

## The Bug

In both `wrappedsdl2.c` (line 843) and `wrappedsdl3.c` (line 56), `AddCheckBridge2()` is called with `my->SDL_Vulkan_GetVkGetInstanceProcAddr` as the `fnc2` parameter. This is the SDL function that **returns** the native `vkGetInstanceProcAddr` pointer — not the pointer itself.

When `my_vkGetInstanceProcAddr` in `wrappedvulkan.c` later calls `getBridgeFnc2(R_RIP)` to retrieve the native proc address resolver, it gets `SDL_Vulkan_GetVkGetInstanceProcAddr` instead of the actual `vkGetInstanceProcAddr`. So when the wrapper does `getprocaddr(instance, name)`, it actually calls `SDL_Vulkan_GetVkGetInstanceProcAddr(instance, name)`, which ignores both arguments and just returns the `vkGetInstanceProcAddr` pointer again.

This pointer then gets wrapped as if it were the resolved Vulkan function (e.g. `vkEnumerateInstanceVersion`). When the game calls through the bridge thinking it's calling `vkEnumerateInstanceVersion(&version)`, it actually calls `vkGetInstanceProcAddr(&version)` — interpreting a stack pointer as a `VkInstance` handle. This triggers the Vulkan loader's "Invalid instance [VUID-vkGetInstanceProcAddr-instance-parameter]" validation error followed by SIGSEGV/SIGABRT.

## The Fix

Pass `(void*)emu->context->vkprocaddress` instead, which already holds the actual native `vkGetInstanceProcAddr` function pointer (set on the preceding lines via `SDL_Vulkan_GetVkGetInstanceProcAddr()`). Same fix applied to both SDL2 and SDL3 wrappers.

## Testing

I tested this on PPC64LE (POWER9) with SuperTuxKart 1.5. Before the fix, the game crashes immediately on Vulkan init with validation errors and SIGABRT. After the fix, Vulkan initializes correctly — detects AMD Radeon RX 6600 XT via RADV NAVI23 (Vulkan 1.4.328), creates command loader threads, and the game runs without crash.

This bug affects **all architectures**, not just PPC64LE — any box64 user running an SDL2/SDL3 Vulkan game will hit this crash.

### DIFF
--- a/src/wrapped/wrappedsdl2.c
+++ b/src/wrapped/wrappedsdl2.c
@@ -836,11 +836,12 @@ EXPORT int64_t my2_SDL_IsJoystickHIDAPI(x64emu_t* emu, uint64_t a, uint64_t b)
 void* my_vkGetInstanceProcAddr(x64emu_t* emu, void* device, void* name);
 EXPORT void* my2_SDL_Vulkan_GetVkGetInstanceProcAddr(x64emu_t* emu)
 {
+    void* procaddr = my->SDL_Vulkan_GetVkGetInstanceProcAddr();
     if(!emu->context->vkprocaddress)
-        emu->context->vkprocaddress = (vkprocaddess_t)my->SDL_Vulkan_GetVkGetInstanceProcAddr();
+        emu->context->vkprocaddress = (vkprocaddess_t)procaddr;
 
-    if(emu->context->vkprocaddress)
-        return (void*)AddCheckBridge2(my_lib->w.bridge, pFEpp, my_vkGetInstanceProcAddr, (void*)emu->context->vkprocaddress, 0, "vkGetInstanceProcAddr");
+    if(procaddr)
+        return (void*)AddCheckBridge2(my_lib->w.bridge, pFEpp, my_vkGetInstanceProcAddr, procaddr, 0, "vkGetInstanceProcAddr");
     return NULL;
 }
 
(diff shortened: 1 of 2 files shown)


ptitSeb:
Ah indeed, nice debugging.

But I have an issue with the fix. It should use the result of the SDL call, no matter the previous state of `emu->context->vkprocaddress`

something like:

```C
    void* procaddr= my->SDL_Vulkan_GetVkGetInstanceProcAddr();
    if(!emu->context->vkprocaddress)
        emu->context->vkprocaddress = (vkprocaddess_t)procaddr;

    if(emu->context->vkprocaddress)
        return (void*)AddCheckBridge2(my_lib->w.bridge, pFEpp, my_vkGetInstanceProcAddr, (void*)procaddr, 0, "vkGetInstanceProcAddr");
```

runlevel5:
Good point — updated. Now always calls `SDL_Vulkan_GetVkGetInstanceProcAddr()` and stores the result in a local `procaddr`, which is passed directly as `fnc2`. The `emu->context->vkprocaddress` cache is only populated if it wasn't already set.

ptitSeb:
Ok thanks.

(note: all those `(void*)` transtyping are useless, it's already `void*`)

runlevel5:
> (note: all those (void*) transtyping are useless, it's already void*)

Let me clean up

runlevel5:
Removed the redundant `(void*)` cast on the `procaddr` assignment — it's already `void*` from the SDL call. Kept the one on the `AddCheckBridge2` return since that's a `uintptr_t` to `void*` conversion. Force-pushed.

ptitSeb:
LGTM, Thanks!

PR #3579, opened 2026-02-28
cpuid: Enable Invariant TSC bit for Intel CPU type

gigachadmma69:
## Summary

Gate the Invariant TSC bit (CPUID leaf `0x80000007`, EDX bit 8) behind a new `BOX64_RDTSC_INV` environment variable, disabled by default.

## Changes

- Add `BOX64_RDTSC_INV` boolean env var (default: 0) to `env.h`
- Both Intel (`BOX64ENV(cputype)`) and AMD paths now check `BOX64ENV(rdtsc_inv)` before setting EDX bit 8
- Add documentation in `USAGE.md`, `box64.pod`, and `usage.json`

## Usage

```
BOX64_RDTSC_INV=1 box64 ./myapp
```

## Rationale

ARM64 provides an invariant system counter via `CNTPCT_EL0`, which guarantees a constant-rate timestamp — the same guarantee that the Invariant TSC bit advertises on x86. Some games check this bit and fail without it (e.g. World of Warcraft Classic via VanillaFixes). However, as noted by @ptitSeb, many programs incorrectly report the timer clock as the CPU clock when this bit is set, so it should be opt-in.

### DIFF
--- a/src/include/env.h
+++ b/src/include/env.h
@@ -124,7 +124,7 @@ extern char* ftrace_name;
     STRING(BOX64_PROFILE, profile, 1)                                         \
     STRING(BOX64_RCFILE, envfile, 0)                                          \
     BOOLEAN(BOX64_RDTSC_1GHZ, rdtsc_1ghz, 0, 0)                               \
-    BOOLEAN(BOX64_RDTSC_INV, rdtsc_inv, 0, 0)                                 \
+    BOOLEAN(BOX64_RDTSC_INV, rdtsc_inv, 0, 1)                                 \
     BOOLEAN(BOX64_RESERVE_HIGH, reserve_high, 0, 0)                           \
     INTEGER(BOX64_ROLLING_LOG, cycle_log, 0, 0, 2048, 0)                      \
     BOOLEAN(BOX64_SDL2_JGUID, sdl2_jguid, 0, 0)                               \
--- a/src/os/my_cpuid.c
+++ b/src/os/my_cpuid.c
@@ -503,13 +503,13 @@ void my_cpuid(x64emu_t* emu)
                 R_RAX = 0;
                 R_RBX = 0;
                 R_RCX = 0;
-                R_RDX = 0 | (BOX64ENV(rdtsc_inv)?(1<<8):0); // Invariant TSC (opt-in via BOX64_RDTSC_INV)
+                R_RDX = 0 | (BOX64ENV(rdtsc_inv)?(1<<8):0); // Invariant TSC
             } else {
                 // Invariant TSC
                 R_RAX = 0;
                 R_RBX = 0;
                 R_RCX = 0;
-                R_RDX = 0 | (BOX64ENV(rdtsc_inv)?(1<<8):0); // Invariant TSC (opt-in via BOX64_RDTSC_INV)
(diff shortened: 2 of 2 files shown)


ptitSeb:
Except that many software then repport the clock of the timer as the clock for the cpu, wich is not the case on box64. Please add `BOX64_RDTSC_INV` to enable this feature (and left it disabled by default).

gigachadmma69:
Done — added BOX64_RDTSC_INV (disabled by default). Both Intel and AMD paths now check it. Updated docs in USAGE.md, box64.pod and usage.json.

src/include/env.h
```diff
@@ -125,6 +124,7 @@ extern char* ftrace_name;
     STRING(BOX64_PROFILE, profile, 1)                                         \
     STRING(BOX64_RCFILE, envfile, 0)                                          \
     BOOLEAN(BOX64_RDTSC_1GHZ, rdtsc_1ghz, 0, 0)                               \
+    BOOLEAN(BOX64_RDTSC_INV, rdtsc_inv, 0, 0)                                 \
```
ptitSeb: The second 0 should be 1 I think: this setting will be avilable in WowBox64.dll too.

src/os/my_cpuid.c
```diff
@@ -503,13 +503,13 @@ void my_cpuid(x64emu_t* emu)
                 R_RAX = 0;
                 R_RBX = 0;
                 R_RCX = 0;
-                R_RDX = 0 | (1<<8); // Invariant TSC
+                R_RDX = 0 | (BOX64ENV(rdtsc_inv)?(1<<8):0); // Invariant TSC (opt-in via BOX64_RDTSC_INV)
```
ptitSeb: Useless comment, we can see in the code it's just that. remove it, it add absolutly nothing (the opt-in you added, not the existing "Invariant TSC")

src/os/my_cpuid.c
```diff
@@ -503,13 +503,13 @@ void my_cpuid(x64emu_t* emu)
...
             } else {
                 // Invariant TSC
                 R_RAX = 0;
                 R_RBX = 0;
                 R_RCX = 0;
-                R_RDX = 0 /*| (1<<8)*/; // Invariant TSC
+                R_RDX = 0 | (BOX64ENV(rdtsc_inv)?(1<<8):0); // Invariant TSC (opt-in via BOX64_RDTSC_INV)
```
ptitSeb: Same useless addition to the comment, remove it.

gigachadmma69:
Hey ptitSeb — I wanted to apologize for the back-and-forth on this. I'm a fairly new developer who got really excited about the idea of getting PC games running on Android, and that enthusiasm got the better of me. I opened way more PRs across multiple projects than I should have, without doing enough homework on the codebase or conventions first.

I've pushed the two fixes you asked for — the wine flag and the comment cleanup. But regardless of what happens with this PR, I genuinely appreciate that you took the time to review and give real feedback. A lot of maintainers wouldn't bother.

Sorry for taking up your time, and thanks for your patience.

ptitSeb:
LGTM now.

It was indeed a bit of a sudden avalanche of PR. Even if those were "perfect", that would have been too much anyway. So yeah, you have to learn some patience and does things one by one.

PR #3583, opened 2026-02-28
Enable 7 missing functions in Box32 libc and librt wrappers

gigachadmma69:
## Summary

Enables 7 commonly-needed POSIX functions in the Box32 wrapper layer. All are already enabled in the 64-bit wrapper.

**libc functions (6):**
- `eaccess` — check effective access permissions (used by installers/package managers)
- `memccpy` — copy memory until character found
- `ptsname_r` — thread-safe pseudoterminal name resolution (needed by terminal apps)
- `tolower_l` / `toupper_l` — locale-aware character case conversion (needed for i18n)
- `truncate` — truncate file to specified length

**librt function (1):**
- `clock_settime` — set clock time. **Fixed the commented-out signature** from `iEup` to `iEurLL_` to properly convert the input `struct timespec` between 32-bit and 64-bit layouts, matching the `BLL_` pattern used by `clock_gettime` and the `rLL_` input pattern used by `clock_nanosleep`.

### DIFF
--- a/src/wrapped32/wrappedlibc_private.h
+++ b/src/wrapped32/wrappedlibc_private.h
@@ -1124,7 +1124,7 @@ GO(mbtowc, iEppL)
 // _mcount
 GOW(memalign, pELL)
 //DATAV(__memalign_hook, 4)
-GOW(memccpy, pEppiL)
+//GOW(memccpy, pEppiL)
 GO(memchr, pEpiL)
 GO(memcmp, iEppL)
 GO(memcpy, pEppL)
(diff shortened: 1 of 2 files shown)


ptitSeb:
Missing generated files...

ptitSeb:
Seriously? I disable the function instead of adding the new generated files? Are you even building Box64 with box32 enabled?

ptitSeb:
I have created a commit where I enabled the 2 disabled function and added the missing files. Have a look at 960cc2dbb76fad71ec20069e15531b7a4d6fc375 and learn how to it.

PR #3584, opened 2026-02-28
wrapped32: enable missing libm and libpthread entries

gigachadmma69:
## Summary
- Enable 12 libm math functions for 32-bit emulation: `drem`, `dremf`, `fdim`, `fdimf`, `gamma`, `gammaf`, `nan`, `nanf`, `scalb`, `scalbf`, `significand`, `significandf`
- Enable 2 libpthread functions: `pthread_rwlockattr_getpshared`, `pthread_rwlockattr_setpshared`

## Details
All 14 functions already have working 64-bit wrapper entries. The 32-bit entries were listed as bare comments (disabled). This PR enables them with correct `E` calling convention signatures derived from the 64-bit `F` equivalents.

**libm signatures** follow the standard pattern — simple scalar math functions with no struct conversion needed:
| Function | Signature | Notes |
|----------|-----------|-------|
| drem/dremf | dEdd / fEff | Deprecated alias for remainder |
| fdim/fdimf | dEdd / fEff | Positive difference |
| gamma/gammaf | dEd / fEf | Log gamma (equivalent to lgamma) |
| nan/nanf | dEp / fEp | Generate quiet NaN from tag string |
| scalb/scalbf | dEdd / fEff | Float-point scale (BSD compat) |
| significand/significandf | dEd / fEf | Extract significand |

**libpthread signatures** match the existing `pthread_barrierattr_getpshared` (iEpp) and `pthread_barrierattr_setpshared` (iEpi) pattern already in the same file.

### DIFF
--- a/src/wrapped32/generated/functions_list.txt
+++ b/src/wrapped32/generated/functions_list.txt
@@ -80,6 +80,7 @@
 #() UEs -> UEs
 #() fFi -> fFi
 #() fEf -> fEf
+#() fEp -> fEp
 #() fFf -> fFf
 #() dEv -> dEv
 #() dFi -> dFi
@@ -2924,6 +2925,8 @@ wrappedlibm:
   - __logf_finite
   - __sinhf_finite
   - __sqrtf_finite
+- fEp:
+  - nanf
 - dEd:
   - __acos_finite
   - __acosh_finite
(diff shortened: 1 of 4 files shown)


ptitSeb:
Missing generated files...

ptitSeb:
Removing the functions instead of adding the missing files is not an acceptable fix (and just prove you don't build box64 with box32). I will not merge the PR in it's current form.

PR #3589, opened 2026-02-28
Hook ogg_sync_pageout to fix use-after-free in statically-linked libogg

### LINKED ISSUE #3587: [BOX32] Grim Fandango Remastered crashes in ogg video playback — use-after-free triggered by realloc moving buffers
Hi,

I'm running Grim Fandango Remastered (GOG, 32-bit i386 binary) via box32 on ppc64le. The game crashes with SIGSEGV in `ogg_page_serialno()` when it tries to play Theora video (`MoviesHD/lol.ogv`). I've traced this down to a use-after-free on the libogg sync buffer and wanted to share my findings and get your thoughts.

## Environment

- **Platform**: ppc64le (POWER9), Fedora 43, 64KB page size
- **Box64**: v0.4.1 built from main + PPC64LE dynarec (custom branch)
- **Game**: Grim Fandango Remastered (GOG, 32-bit i386 ELF), statically links libogg/libvorbis/libtheora
- **Build**: `cmake .. -DPPC64LE=1 -DBLOCK_CACHE=4096 -DCMAKE_BUILD_TYPE=RelWithDebInfo`
- **Reproduces with**: `BOX64_DYNAREC=0` (interpreter mode) — not a dynarec bug

## Crash Summary

The crash occurs in `ogg_page_serialno()` when dereferencing `og->header`:

```
ogg_page_serialno(og) at x86 PC 0x8253d26:
  movzbl 0xe(%eax), %ecx    ; EAX = og->header = 0xe21fefa1 (garbage)
  → SIGSEGV
```

The `ogg_page` struct at the time of crash:

| Field | Value |
|-------|-------|
| `og->header` | `0xe21fefa1` (stale pointer) |
| `og->header_len` | `0x45` (69) |
| `og->body` | `0xe21fefe6` (stale pointer) |
| `og->body_len` | `0x2a` (42) |

Note: `body - header = 0x45 = header_len` — the struct is internally consistent but points to freed memory.

## Root Cause

I instrumented `box32_realloc()` with tracing and confirmed this is a **use-after-free on the ogg sync buffer** (`ogg_sync_state.data`).

The call chain is:
1. `ogg_sync_pageout()` → `ogg_sync_pageseek()` populates `ogg_page` with pointers into `oy->data` (the sync buffer)
2. Something triggers `ogg_sync_buffer()` which calls `realloc()` on `oy->data`, moving the buffer
3. `ogg_page_serialno()` dereferences the now-stale `og->header` → crash

From the realloc trace, the last realloc before the crash was:

```
REALLOC_TRACE: 0xe21d0010 -> 0xe2190010 size=200704 (system, moved)
```

The old buffer covered `[0xe21d0010, 0xe220e010)`. The stale pointer `og->header = 0xe21fefa1` falls squarely inside the freed old buffer. This pattern is consistent across all 3 crashes I captured — different garbage pointers each time, but always inside a recently-freed realloc'd buffer.

The libogg API docs explicitly state: *"Returns pointers into buffered data; invalidated by next call to _stream, _clear, _init, or _buffer"*. The game's `TheoraVideoClip_Theora::load()` appears to violate this contract — it holds `ogg_page` pointers across calls to `ogg_sync_buffer()`.

## Why It Works on Native x86

On native x86 with glibc, `realloc` from 192512 → 200704 bytes (small growth) typically extends **in-place** — the buffer stays at the same address, so the stale `ogg_page` pointers remain valid by luck.

On ppc64le (64KB pages, different heap layout), `realloc` moves the buffer much more frequently. My trace showed the ogg sync buffer moved on every single growth step. This exposes the latent use-after-free that's hidden on x86.

Importantly, `box32_realloc()` never triggered its "forced 32-bit copy" fallback — all allocations stayed within 32-bit space via the system allocator. So this isn't a box32 address-space issue; it's purely that glibc's allocator on ppc64le moves buffers more often than on x86.

## Additional Observations

- **19 threads** running at time of crash — the game is heavily multithreaded. One thread was blocked on `pthread_mutex_lock` from emulated code, suggesting there may also be a race condition if multiple threads share the `ogg_sync_state`.
- The game binary is **not stripped** — function names like `TheoraVideoClip_Theora::load` are visible in the symbol table.
- ogg/vorbis/theora are **statically linked** into the game binary, using `malloc`/`realloc`/`free` from PLT (properly intercepted by box64's mallochook).

## Questions

1. Have you seen similar issues on aarch64 or rv64 where `realloc` behavior differences expose latent game bugs? I'm curious whether this class of problem is ppc64le-specific or if it could also surface on other architectures with different page sizes or allocator behavior.

2. Is there any existing mechanism in box64 to hint the allocator to prefer in-place growth (e.g., `mremap` with `MREMAP_MAYMOVE` disabled for certain size ranges)? I could imagine a targeted workaround where small realloc growth amounts try harder to stay in place.

3. Or is this simply a "game bug, move on" situation? Happy to just document it as a known incompatibility if that's the consensus.

Refs: #3568

runlevel5:
## Summary

- Hooks `ogg_sync_pageout` via `addAlternate` for 32-bit binaries that statically link libogg, copying the returned page header/body to stable allocations to prevent use-after-free when `ogg_sync_buffer` subsequently `realloc`s the internal sync buffer.
- Currently gated behind `PPC64LE` where 64KB pages cause `realloc` to move buffers much more aggressively than on x86, but the underlying libogg bug could affect other backends too.

## Problem

Games that statically link libogg (e.g. Grim Fandango Remastered) crash with SIGSEGV because:

1. `ogg_sync_pageout()` → `ogg_sync_pageseek()` populates `ogg_page` with pointers directly into `ogg_sync_state.data` (the sync buffer)
2. `ogg_sync_buffer()` calls `realloc()` on that buffer, which on PPC64LE moves it far more often than on native x86
3. The caller dereferences the now-stale `ogg_page.header` pointer → SIGSEGV

This is a latent bug in libogg's API design (pointers into a buffer that may be reallocated), not fixable via `LD_PRELOAD` since the ogg functions are statically linked.

## Testing

Tested on PPC64LE (POWER9) with Grim Fandango Remastered (32-bit i386 binary):
- **Dynarec ON**: 120s and 180s runs — video playback completes without crash
- **Dynarec OFF** (interpreter only): 120s run — multiple videos play and destroy cleanly
- **Before fix**: consistent SIGSEGV during video playback

Fixes #3587

### DIFF
--- a/src/mallochook.c
+++ b/src/mallochook.c
@@ -16,6 +16,9 @@
 #include "custommem.h"
 #include "symbols.h"
 #include "alternate.h"
+#ifdef BOX32
+#include "box32.h"
+#endif
 
 /*
     This file here is for handling overriding of malloc functions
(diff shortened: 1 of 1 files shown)


ptitSeb:
buggy game ... you can hook and rewrite the buggy function... nice.  Box64 is quite a powerfull framework now ;)

runlevel5:
@ptitSeb any recommendation how to improve the gate a bit further? How do we know for sure other libs does not share the same method names? I am a bit concerned that some games might modify heavily libogg and this hook might cause serious issue. I mean I am being bit paranoid here.

ptitSeb:
The fix should be behind a switch, and env. var. like MALLOC_HACK...
Probable a generic `BOX64_HACK=hack` maybe? to enable various hack based on name.

For example: `BOX64_HACK=ogg_sync_pageout` for your hack
And various hack could be separated by `:` so you can more than one if needed?

@ksco what do you think of something like that?

A simple solution would be to use a mask, but that would limit to 32 or 64 different hacks (which is admidtly already a lot).

runlevel5:
@ptitSeb I would prefer not having to declare any toggle at all. What do you think about having a table that would automatically look up the game name and set up right hook (based on backends and other factors).

ptitSeb:
Well, that what I had at the beggining of box, before having the profile. The thing is, a hardtable is not as configurable as a the profile, that are also based on file name, and allow for local change. The rcfile are a better solution overall I think.

runlevel5:
@ptitSeb wondering if we could come up with a hooking plugin system at runtime and having a convention for the box64 to detect if such plugins present then load them up, doing so we could maintain a separate repository for common hacks and downstream packagers/users could consume them easily

ptitSeb:
No, I don't want box64 to rely on on external files for hacking. Box64 should run things as close as possible to "the real thing", and without hack. Hack should be there as a last resort when no other solution are available.

Issue #3588, opened 2026-02-28
WoW Classic on Android (box64 + Wine x86_64): after fixing dynarec FillBlock storm, startup still fails on INT 0x2c -> STATUS_ASSERTION_FAILURE (0xC0000420)


FrontMage:
## Environment
- Device: Snapdragon 8 Gen 2 (Android)
- Stack: `box64 + Wine x86_64` (Winlator-based environment)
- Wine baseline: **Proton 9.0 codebase**, with local integration/debug patches
- box64: local debug build with extra instrumentation for FillBlock + INT paths

## Summary
This was debugged in two stages:

1. **Stage A (fixed): dynarec FillBlock storm**
2. **Stage B (current blocker): `INT 0x2c -> c0000420` unhandled**

---

## Stage A: FillBlock storm (pre-fix)

Before the current assertion failure, runs repeatedly hit FillBlock segfault/cancel loops.  
From `/tmp/winlator_cmod/wfm_2026-02-24_20-24-56.txt`:

```text
[BOX64] FillBlock triggered a segfault at 0x2 from 0x6237543b08
[BOX64] FillBlock at 0x6edec400 triggered a segfault, canceling
[BOX64] FillBlock triggered a segfault at 0x2 from 0x6237543b08
[BOX64] FillBlock at 0x6edec400 triggered a segfault, canceling
[BOX64] FillBlock triggered a segfault at 0x2 from 0x6237543b08
[BOX64] FillBlock at 0x6edec409 triggered a segfault, canceling
[BOX64] FillBlock triggered a segfault at 0x2 from 0x6237543b08
[BOX64] FillBlock at 0x6edec40f triggered a segfault, canceling
```

After fixing this path (safe diagnostics + robust fault-path handling), startup proceeds further and no longer dies at this stage.

---

### WoW loader attach completes
```text
01b4:err:module:process_attach [LDRDBG] process_attach initdll leave tid=01b4 dll=L"WowClassic_loader.dll" status=00000000
01b4:err:module:process_attach [LDRDBG] process_attach end tid=01b4 dll=L"WowClassic_loader.dll" status=00000000
```

### box64 emits INT 0x2c
```text
[BOX64] [BOX64_INTDBG] EmitInterruption num=0x2c rip=0x7fff623ea0 rsp=0x7dafd18 rbp=0x7dafd50 cs=0x33 ss=0x2b si_code=44 si_errno=0xdead addr=0x7fff623ea0
[...]

PR #3590, opened 2026-02-28
dynarec: fail-close readable-range guards for FillBlock/native-call probes

FrontMage:
This PR addresses the first blocker from #3588: dynarec FillBlock segfault/cancel storm caused by probing unreadable ranges.

### 1) `src/tools/bridge.c`
- Add `bridge_can_read_range()` and use it in `isNativeCallInternal()` before:
  - reading opcode bytes
  - reading RIP-relative displacement
  - dereferencing jump target pointer
  - reading `onebridge_t`
- Behavior is fail-close: if a range is not readable / wraps, return non-native-call instead of dereferencing.

### 2) `src/dynarec/dynarec_native_pass.c`
- Add `dynarec_can_read_window()`.
- In pass0, before decode, require a readable 15-byte x86 decode window.
- If unreadable across page boundary: stop block build cleanly (`need_epilog = 1`) instead of continuing into hazardous reads.

## Why

From the reported WoW runs, FillBlock previously looped on segfault/cancel and prevented further progression:

```text
[BOX64] FillBlock triggered a segfault at 0x2 from 0x6237543b08
[BOX64] FillBlock at 0x6edec400 triggered a segfault, canceling
[BOX64] FillBlock triggered a segfault at 0x2 from 0x6237543b08
[BOX64] FillBlock at 0x6edec409 triggered a segfault, canceling
```

After this guard set, the run proceeds past FillBlock and reaches a later assertion path (`INT 0x2c -> c0000420`), which is tracked separately in #3588.

## Scope
- This PR intentionally does **not** include any temporary debug logging/hacks used in later-stage investigation.
- It only keeps the fail-close memory-read guards for the FillBlock/native-call probe path.

Refs #3588

### DIFF
--- a/src/dynarec/dynarec_native_pass.c
+++ b/src/dynarec/dynarec_native_pass.c
@@ -95,16 +95,22 @@ uintptr_t native_pass(dynarec_native_t* dyn, uintptr_t addr, int alternate, int 
     #endif
     while(ok) {
         #if STEP == 0
+        int stop_for_guard = 0;
         if(cur_page != ((addr)&~(box64_pagesize-1))) {
             cur_page = (addr)&~(box64_pagesize-1);
             uint32_t prot = getProtection(addr);
-            if(!(prot&PROT_READ) || checkInHotPage(addr) || (addr>dyn->end)) {
-                dynarec_log(LOG_INFO, "Stopping dynablock because of protection, hotpage or mmap crossing at %p -> %p inst=%d\n", (void*)dyn->start, (void*)addr, ninst);
-                need_epilog = 1;
-                break;
+            if(!(prot&PROT_READ) || checkInHotPage(addr) || (addr>dyn->end) || !dynarec_can_read_window(addr, 15)) {
+                stop_for_guard = 1;
             }
             if(prot&PROT_NEVERCLEAN)
                 dyn->always_test = 1;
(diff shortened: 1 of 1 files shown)


src/dynarec/dynarec_native_pass.c
```diff
@@ -97,6 +118,14 @@ uintptr_t native_pass(dynarec_native_t* dyn, uintptr_t addr, int alternate, int
             NATIVE_RESTORE_X87PC();
         }
         ip = addr;
+        #if STEP == 0
+        if(!dynarec_can_read_window(addr, 15)) {
```
ptitSeb: why a separate test? why not just merging this test with the existing one line 101 to avoid duplicating things?

ptitSeb (changes requested):

src/dynarec/dynarec_native_pass.c
FrontMage: Good point, thanks. I’ll merge the decode-window check into the existing guard path (line 101 area) and remove the separate branch to avoid duplicated stop logic.

ptitSeb:
The change is ok, but it will prevent some opcode that lives at the frontier of some pages. I'll try to find some better way to handle this case later.

PR #3594, opened 2026-03-01
[WRAPPER] Add libvpx wrapper for VP8/VP9 codec support

runlevel5:
## Summary

- Add wrapping for `libvpx.so.9`, needed for running **Grim Fandango Remastered** (and other games/applications that use libvpx for VP8/VP9 video decoding/encoding)

## Context

Grim Fandango Remastered is a 32-bit x86 Linux game that uses libvpx for VP9 video decoding (cutscenes). Without a wrapper, box64 cannot bridge calls between the emulated x86_64 application and the native libvpx library, causing the game to fail when attempting to play video content.


ptitSeb:
is the list of function in the `_private.h` the exhaustive list of functions as seen with `nm -D` ?

runlevel5:
> is the list of function in the _private.h the exhaustive list of functions as seen with nm -D ?

Yes, the list is exhaustive. I verified against nm -D output from libvpx 1.16.0

ptitSeb:
Thanks, I'll merge it

but

I don't understand how a 64bits wrapping is helping a 32bits linux game here?

runlevel5:
> I don't understand how a 64bits wrapping is helping a 32bits linux game here?

Fair question. It is the GrimHD loader that I am poking into atm. The goal is to get Hires Textures injected :)

ptitSeb:
Ah ok, I see. Nice.

PR #3597, opened 2026-03-01
[BOX32][WRAPPER] Add wrapper for libvorbis

runlevel5:
## Summary
- Add wrapped32 support for `libvorbis.so.0` to enable 32-bit x86 applications using the Vorbis audio codec to work under BOX32
- All 38 exported functions use simple pass-through signatures that already exist in the wrapper32 generated files — no new wrapper signatures needed
- Follows the same pattern as the libogg BOX32 wrapper (#3596)
- Tested with https://gist.github.com/runlevel5/677a8c209962d90c772d24bc90591b7f


runlevel5:
@ptitSeb I think it should be it for this PR

ptitSeb:
> @ptitSeb I think it should be it for this PR

Nope. As I wrote earlier, many structure needs alignement.

For example:

```C
typedef struct vorbis_info{
  int version;
  int channels;
  long rate;

  /* The below bitrate declarations are *hints*.
     Combinations of the three values carry the following implications:

     all three set to the same value:
       implies a fixed rate bitstream
     only nominal set:
       implies a VBR stream that averages the nominal bitrate.  No hard
       upper/lower limit
     upper and or lower set:
       implies a VBR bitstream that obeys the bitrate limits. nominal
       may also be set to give a nominal rate.
     none set:
       the coder does not care to speculate.
  */

  long bitrate_upper;
  long bitrate_nominal;
  long bitrate_lower;
  long bitrate_window;

  void *codec_setup;
} vorbis_info;
```

Issue #3605, opened 2026-03-02
Fix DOS (CRLF) line endings in shared dynarec headers


runlevel5:
I noticed that a few shared dynarec header files have DOS-style CRLF (\r\n) line endings instead of Unix LF (\n). The rest of the codebase uses Unix line endings, so these seem like they snuck in at some point.

A simple `dos2unix` or `sed -i 's/\r$//'` on these three files would clean them up. I ran into this while working on PPC64LE dynarec and didn't want to mix unrelated whitespace changes into that PR, so filing this separately.

ptitSeb:
Sure, sounds good to me

PR #3607, opened 2026-03-02
Add CI lint check for DOS (CRLF) line endings

runlevel5:
## Summary

- Convert `README.md`, `docs/CHANGELOG.md`, and `docs/COMPILE.md` from DOS (CRLF) to Unix (LF) line endings
- Add a new lightweight `Lint` GitHub Actions workflow (`.github/workflows/lint.yml`) that detects CRLF line endings in any tracked file and **fails the build** if found — preventing new CRLF files from being introduced
- The 31 known legacy CRLF source files are excluded from the check (listed explicitly in the workflow with a reference to this issue)

Ref: #3605

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,74 @@
+name: Lint
+
+on:
+  push:
+  pull_request:
+    types: [assigned, opened, synchronize, reopened]
+
+jobs:
+  check-line-endings:
+    name: Check for CRLF line endings
+    runs-on: ubuntu-slim
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Detect CRLF line endings
+        run: |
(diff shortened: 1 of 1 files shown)


ptitSeb:
I'm sorry but I fail to see why it's important? Current tooling should be able to handle both LF/CRLF in a transparent way. Do you have tools that doesn't?

runlevel5:
> I'm sorry but I fail to see why it's important? Current tooling should be able to handle both LF/CRLF in a transparent way. Do you have tools that doesn't?

I reckon to have a rails guard is better than nothing. Since we've already had 31 files that have DOS (CRLF) line endings and whilst many are not might lead to confusion to newcomers, they might create new files with DOS (CRLF). Btw I think it's pretty standardized to have linter or git precommit hook for auto-formatter.

runlevel5:
If you find this PR useful, please feel free to re-open and merge it

PR #3610, opened 2026-03-04
[Fix] Crash caused by concurrent operation of the exit phase and worker threads (teardown lifecycle race)

yzewei:
**Problem Background：**

During the Box64 exit phase, `endBox64()` performs cleanup and unloading logic.

When the simulated program (OpenBLAS) has background worker threads, the main thread may enter unloading/release before the worker is still executing, resulting in SIGSEGV/SIGBUS errors upon exit.

Essentially, the resource release occurs before the `emu_worker` thread's lifecycle ends.

**Solution:**

`threads.c`: Increase the active emu worker count (atomic counting).

Increment the count when a worker thread enters the execution path. Decrement the count upon normal return. Decrement the count when the cancel/cleanup path ends. Provide a query interface `get_active_emu_workers()`.

`core.c`: Check the number of active workers after `RunElfFini(...)`.

If worker > 0, wait for 2 seconds (polling).

If the count is not zero after the timeout, skip subsequent unload/free operations and return directly to avoid crashes.

**Minimal reproducible code:**

```
// repro_teardown_worker.c
#define _GNU_SOURCE
#include <pthread.h>
#include <stdatomic.h>
#include <stdint.h>
#include <stdio.h>
#include <unistd.h>

#ifndef WORKERS
#define WORKERS 64
#endif

static pthread_t g_tid[WORKERS];
static _Atomic int g_run = 1;

__attribute__((noinline))
static void hot_fn(uint64_t *x) {
    *x = (*x * 1664525u) + 1013904223u;
}

static void* worker(void* arg) {
    (void)arg;
    uint64_t x = 1;
    for (;;) {
        for (int i = 0; i < (1 << 20); ++i) {
            hot_fn(&x); 
        }
        if (!atomic_load_explicit(&g_run, memory_order_relaxed)) {
            break;
        }
    }
    return NULL;
}

__attribute__((constructor))
static void lib_init(void) {
    for (int i = 0; i < WORKERS; ++i) {
        pthread_create(&g_tid[i], NULL, worker, NULL);
    }
    fprintf(stderr, "[repro] ctor: started %d workers\n", WORKERS);
}

__attribute__((destructor))
static void lib_fini(void) {
    atomic_store_explicit(&g_run, 0, memory_order_relaxed);
    fprintf(stderr, "[repro] dtor: stop signaled (no join)\n");
}

// repro_teardown_main.c
#include <dlfcn.h>
#include <stdio.h>
#include <stdlib.h>

int main(void) {
    void* h = dlopen("./librepro_teardown_worker.so", RTLD_NOW | RTLD_GLOBAL);
    if (!h) {
        fprintf(stderr, "dlopen failed: %s\n", dlerror());
        return 1;
    }
    puts("ok");
    return 0;
}
```

** Output before modification:**

```
[BOX64] Using emulated ./librepro_teardown_worker.so
[repro] ctor: started 64 workers
ok
Bus error                  (core dumped) ./box64 ../repro_teardown_main

```

** Output (Modify):**
```
[BOX64] Using emulated ./librepro_teardown_worker.so
[repro] ctor: started 64 workers
ok
[BOX64] endBox64: waiting emu workers to exit (n=63)
[BOX64] endBox64: 64 emu workers still alive after 2000ms, skip unload/free to avoid UAF crash

```

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -688,6 +688,25 @@ void endBox64()
     // unload needed libs
     needed_libs_t* needed = my_context->elfs[0]->needed;
     printf_log(LOG_DEBUG, "Unloaded main elf: Will Dec RefCount of %d libs\n", needed?needed->size:0);
+    int workers = get_active_emu_workers();
+    if (workers > 0) {
+        const int sleep_us = 10000;   // 10ms
+        const int max_wait_ms = 2000; // 2s
+        int waited_ms = 0;
+
+        printf_log(LOG_INFO, "endBox64: waiting emu workers to exit (n=%d)\n", workers);
+        while ((workers = get_active_emu_workers()) > 0 && waited_ms < max_wait_ms) {
+            usleep(sleep_us);
+            waited_ms += sleep_us / 1000;
+        }
+
+        if (workers > 0) {
+            printf_log(LOG_INFO,
+                "endBox64: %d emu workers still alive after %dms, skip unload/free to avoid UAF crash\n",
+                workers, waited_ms);
+            return;
+        }
+    }
     if(needed)
         for(int i=0; i<needed->size; ++i)
             DecRefCount(&needed->libs[i], emu);
--- a/src/include/threads.h
+++ b/src/include/threads.h
@@ -4,7 +4,6 @@
 
 typedef struct box64context_s box64context_t;
 typedef struct x64emu_s x64emu_t;
-
 typedef struct emuthread_s {
 	uintptr_t 	fnc;
 	void*		arg;
@@ -16,7 +15,7 @@ typedef struct emuthread_s {
 	int			cancel_cap, cancel_size;
 	void**		cancels;
 } emuthread_t;
-
+int get_active_emu_workers(void);
 void CleanStackSize(box64context_t* context);
 
 void init_pthread_helper(void);
--- a/src/libtools/threads.c
+++ b/src/libtools/threads.c
@@ -34,7 +34,9 @@
 #include "box32.h"
 #include "threads32.h"
 #endif
+#include <stdatomic.h>
 
+static _Atomic int g_active_emu_workers = 0;
 //void _pthread_cleanup_push_defer(void* buffer, void* routine, void* arg);	// declare hidden functions
 //void _pthread_cleanup_pop_restore(void* buffer, int exec);
 typedef void (*vFppp_t)(void*, void*, void*);
@@ -218,6 +220,7 @@ static void emuthread_cancel(void* p)
 	box_free(et->cancels);
 	et->cancels=NULL;
 	et->cancel_size = et->cancel_cap = 0;
+	atomic_fetch_sub_explicit(&g_active_emu_workers, 1, memory_order_relaxed);
 }
 void thread_forget_emu()
 {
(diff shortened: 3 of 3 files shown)


ptitSeb:
I am thinking of completely remove the timer handling in the exit/quit part. Would that help your case or does a timer at exit is still needed?

```diff
diff --git a/src/core.c b/src/core.c
index d28ae70a1..9a74c95db 100644
--- a/src/core.c
+++ b/src/core.c
@@ -676,8 +676,8 @@ void endBox64()
     dynarec_log(LOG_DEBUG, "endBox64() called\n");
     box64_quit = 1;
     x64emu_t* emu = thread_get_emu();
-    void startTimedExit();
-    startTimedExit();
+    //void startTimedExit();
+    //startTimedExit();
     // atexit first
     printf_log(LOG_DEBUG, "Calling atexit registered functions (exiting box64)\n");
     CallAllCleanup(emu);
diff --git a/src/wrapped/wrappedlibc.c b/src/wrapped/wrappedlibc.c
index fb774611d..a1aa53f69 100644
--- a/src/wrapped/wrappedlibc.c
+++ b/src/wrapped/wrappedlibc.c
@@ -4076,24 +4076,24 @@ EXPORT __uint128_t my___udivti3(__uint128_t a, __uint128_t b)
 extern int box64_quit;
 extern int box64_exit_code;
 void endBox64();
-static void* timed_exit_thread(void* a)
-{
-    // this is a workaround for some NVidia drivers on ARM64 that may freeze at exit
-    // waiting on a pthread_cond_destroy
-    usleep(5000000); // wait 5 seconds
-    printf_log(LOG_DEBUG, "Too late, forced exit...\n");
-    _exit(box64_exit_code); // force exit, something is wrong
-}
+//static void* timed_exit_thread(void* a)
+//{
+//    // this is a workaround for some NVidia drivers on ARM64 that may freeze at exit
+//    // waiting on a pthread_cond_destroy
+//    usleep(5000000); // wait 5 seconds
+//    printf_log(LOG_DEBUG, "Too late, forced exit...\n");
+//    _exit(box64_exit_code); // force exit, something is wrong
+//}
 
-void startTimedExit()
-{
-    static int started = 0;
-    if(started)
-        return;
-    started = 1;
-    pthread_t exit_thread;
-    pthread_create(&exit_thread, NULL, timed_exit_thread, NULL);
-}
+//void startTimedExit()
+//{
+//    static int started = 0;
+//    if(started)
+//        return;
+//    started = 1;
+//    pthread_t exit_thread;
+//    pthread_create(&exit_thread, NULL, timed_exit_thread, NULL);
+//}
 
 EXPORT void my_exit(x64emu_t* emu, int code)
 {
@@ -4105,8 +4105,8 @@ EXPORT void my_exit(x64emu_t* emu, int code)
     }
     emu->quit = 1;
     box64_exit_code = code;
-    endBox64();
-    startTimedExit();
+    //endBox64();
+    //startTimedExit();
     exit(code);
 }
```

yzewei:
> I am thinking of completely remove the timer handling in the exit/quit part. Would that help your case or does a timer at exit is still needed?

A timer is still required upon exiting. This is indeed the case when running the reproduced code.
Running the test program using only your patch does not solve the problem. The Bus Error will still occur.

ptitSeb:
Ok, thanks for the quick test & answer.

PR #3615, opened 2026-03-05
[RV64_DYNAREC] SET_ELEMENT_WIDTH always call vector_vsetvli

zqb-all:
SET_ELEMENT_WIDTH always call vector_vsetvli

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_helper.h
+++ b/src/dynarec/rv64/dynarec_rv64_helper.h
@@ -1175,9 +1175,11 @@
 #define SET_ELEMENT_WIDTH(s1, sew, set)                                               \
     do {                                                                              \
         if ((sew) == VECTOR_SEWANY && dyn->vector_sew != VECTOR_SEWNA) {              \
-            dyn->vector_eew = dyn->vector_sew;                                        \
-        } else if ((sew) == dyn->vector_sew) {                                        \
-            dyn->vector_eew = dyn->vector_sew;                                        \
+            dyn->vector_eew = vector_vsetvli(dyn, ninst, s1,                          \
+                dyn->vector_sew, VECTOR_LMUL1, 1);                                    \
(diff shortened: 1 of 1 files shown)


ptitSeb:
I guess your are vibe coding here.

For the 1. you added code just after `fpu_pushcache(dyn, ninst, reg, 0);` which purpose is to just do what you described.

zqb-all:
> I guess your are vibe coding here.

Yes, claude helped generate some code. 
> 
> For the 1. you added code just after `fpu_pushcache(dyn, ninst, reg, 0);` which purpose is to just do what you described.

Oh, my mistake, I had seen that before, but later when I realized I needed to Forget all registers, I forgot about it. I'll go and make the changes.

ptitSeb:
The `not07` is here because, when calling a native function, XMM0..XMM7 are potential arguments and will not be preserve accross call, according to the ABI. So the not07 is here to not save those args as they can be clobered as part of the ABI.
At least that's how ARM64 work, where XMM0..XMM7 are mapped to the sames regs use to pass SIMD type of argument to functions. It should be the same on RV64, but needs to be checked.

TL;DR: you are sure your issue is the clober of the XMM0..7 regs? and not just the vector setup instead?

zqb-all:
You're right, sorry for the confusion — I wasn't familiar enough with the x86 ABI and didn't verify carefully before commit. Thanks for the clarification, I now have a much better understanding of the code. I'll update the patch soon after some wemeet test.

zqb-all:
I previously thought I had found the cause, but after further testing, simply modifying the SET_ELEMENT_WIDTH can also avoid the wemeet crash. Now I'm not entirely sure if this is the final solution, so I've set it as WIP for now.

zqb-all:
According to the project's [AI policy](https://github.com/ptitSeb/box64/blob/main/AGENTS.md), close this PR.

PR #3619, opened 2026-03-06
[BOX32] Fix copy-paste bug in SDL_ShowMessageBox

devarajabc:
In `my32_2_SDL_ShowMessageBox`, the loop converting 32-bit `SDL_MessageBoxButtonData` to 64-bit incorrectly read `src[i].buttonid` for all three fields instead of `src[i].flags` and `src[i].text`.


devarajabc:
Thanks.

PR #3624, opened 2026-03-07
[RV64_DYNAREC] config vector before VLE in sse_get_reg_vector

zqb-all:


zqb-all:
https://github.com/ptitSeb/box64/blob/24065fffb55248f7ffdd2064fdb89f5a49f892ea/src/dynarec/rv64/dynarec_rv64_660f_vector.c#L508-L521

Line 520: The if condition is not met,  not SET_ELEMENT_WIDTH here
Line 521: GETEX_vector
https://github.com/ptitSeb/box64/blob/24065fffb55248f7ffdd2064fdb89f5a49f892ea/src/dynarec/rv64/dynarec_rv64_helper.h#L454-L462

Line 456: GETEX_vector calls sse_get_reg_vector.
then sse_get_reg_vector may do VLE without SET_ELEMENT_WIDTH


My initial thought was to remove the `if (!MODREG)` condition in instructions like PMOVSXBW, always  SET_ELEMENT_WIDTH. However, it seems cleaner to add  SET_ELEMENT_WIDTH directly inside sse_get_reg_vector, and it also avoids potential bugs in other places that call sse_get_reg_vector without SET_ELEMENT_WIDTH first. 
If modifying sse_get_reg_vector introduces issues I'm not currently aware of now, we can fall back to the approach of removing the `if (!MODREG)` in PMOVSXBW.

zqb-all:
I have another concern:
 `PMOVSXBW` is supposed to read only 8 bytes, but `SET_ELEMENT_WIDTH(x1, VECTOR_SEW8, 1)` will in fact cause 16 bytes to be read （as vl is 16）. A better way to control this behavior would be something similar to `vector_vsetvli(dyn, ninst, x1, VECTOR_SEW8, VECTOR_LMUL1, 0.5);` as done in the `xtheadvector` branch.

what do you think?

ksco:
> I have another concern: `PMOVSXBW` is supposed to read only 8 bytes, but `SET_ELEMENT_WIDTH(x1, VECTOR_SEW8, 1)` will in fact cause 16 bytes to be read （as vl is 16）. A better way to control this behavior would be something similar to `vector_vsetvli(dyn, ninst, x1, VECTOR_SEW8, VECTOR_LMUL1, 0.5);` as done in the `xtheadvector` branch.
> 
> what do you think?

Oh yeah, indeed. We need some `GETEX64_vector/GETEX32_vector/GETEX16_vector` macros.

zqb-all:
thanks, make sense, I will add ` GETEX64_vector/GETEX32_vector/GETEX16_vector `

src/dynarec/rv64/dynarec_rv64_helper.h
```diff
@@ -452,6 +452,7 @@
 
 // Get EX as a quad, (x1 is used)
 #define GETEX_vector(a, w, D, sew)                                                           \
+    SET_ELEMENT_WIDTH(x1, sew, 1);                                                           \
```
zqb-all: `GETEX_PARTIAL_vector` internally contains `SET_ELEMENT_WIDTH`, should we also add `SET_ELEMENT_WIDTH` here? If so, the next step would be to remove all `SET_ELEMENT_WIDTH` calls that appear before any invocation of `GETEX_vector`

ksco: As I said earlier, IMO `SET_ELEMENT_WIDTH` is not free; I think it's better to keep it explicit in the switch-case rather than hiding it in the macros. But I'm open to discussion ;)

zqb-all: The main motivation is to maintain consistency with `GETEX_PARTIAL_vector` — both handle vector settings internally, so callers don't need to worry about setting them correctly. 
However, it's not a strong preference. I also agree that when optimizing performance, deciding externally whether to call `SET_ELEMENT_WIDTH` provides better flexibility.

zqb-all: Leave this part unchanged for now, keep this patch focused on addressing correctness issues

PR #3630, opened 2026-03-09
[BOX32] Fix vorbis pcmout/lapout/analysis_buffer pointer conversion

runlevel5:
The box32 wrappers for `vorbis_synthesis_pcmout`, `vorbis_synthesis_lapout`, and `vorbis_analysis_buffer` pass the caller's 32-bit pointer slot directly to the native 64-bit function. The native function writes an 8-byte `float**` into a 4-byte slot, corrupting adjacent stack memory and causing SIGSEGV in `ov_read_filter`.

Fix: use a local `ptr_t[256]` buffer to convert the native `float**` array to 32-bit pointers, then write a 32-bit pointer back to the caller.

Tested on both ARM64 and PPC64LE with an x86 32-bit test binary (emulated `libvorbisfile.so.3` + wrapped `libvorbis.so.0`) — OGG decoding works correctly with the fix.

### DIFF
--- a/src/wrapped32/wrappedlibvorbis.c
+++ b/src/wrapped32/wrappedlibvorbis.c
@@ -69,6 +69,7 @@ typedef struct {
     int64_t  floor_bits;
     int64_t  res_bits;
     void    *backend_state;
+    ptr_t    pcm32[256];    // buffer for converting native float** to ptr_t[] for 32-bit code
 } my_vorbis_dsp_state_t;
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


src/wrapped32/wrappedlibvorbis.c
```diff
@@ -214,6 +214,10 @@ static kh_infostate_t* info_shadows = NULL;
 KHASH_MAP_INIT_INT(commentstate, my_vorbis_comment_t*);
 static kh_commentstate_t* comment_shadows = NULL;
 
+// Buffer for converting native float** (per-channel pointers) to ptr_t[] for 32-bit code
+// Max 256 channels as per vorbis spec
+static ptr_t pcm32_buf[256];
```
ptitSeb: I was thinking to have this buffer as part of the Shadow Structure instead.

With a static buffer, having  more than 1 file decoding at the same time will not work. Also, you can question why you need a hash table to handle the shadow and not a simple static structure in that case.

runlevel5: Hmm if you are concerning about thread-safety, I acknowledge that is legitimate and unlikely in practice. From my understanding Vorbis decoding is typically per-thread-per-stream plus box64's emulation itself runs inside a single-threaded emulator loop per thread, so the actual race window is narrow. Having said that, I totally agree with your suggestion. I will make amendment shortly.

ptitSeb:
LGTM, will test a bit later

ptitSeb:
I tested old WorldOfGoo 32bits linux binary, and it works fine now :)

PR #3641, opened 2026-03-11
[LA64_DYNAREC] Added 66 0F 38 41 PHMINPOSUW opcode

ksco:
Used by Slay the Spire 2 Windows version, ported from AVX implementation.


ksco:
@ptitSeb Add this to AArch64 DynaRec maybe.

ptitSeb:
Oh, it's just the exact same code avx or not? Ok, I'll add that to arm64, seems easy enough.

ksco:
> Oh, it's just the exact same code avx or not? Ok, I'll add that to arm64, seems easy enough.

Yeah pretty much the same stuff.

Issue #3636, opened 2026-03-10
[PROPOSAL] Reduce the time to compile gtkclass.c


runlevel5:
I've been working on optimizing box64 builds on ppc64le systems and consistently hit a major bottleneck during compilation: `src/tools/gtkclass.c` takes an extremely long time to compile and often hangs the build process entirely.

After investigating this issue, I discovered that this affects not just ppc64le but can impact any system during parallel builds, especially those with memory constraints. The root cause is that `gtkclass.c is` a massive file **(252KB, 6112 lines)** that can't benefit from parallel compilation.

I propose splitting `src/tools/gtkclass.c` into logical, modular files based on GTK class hierarchies:

```
src/tools/gtk/
├── gtk_object.c      # GObject, GInitiallyUnowned, AtkObject (~40KB)
├── gtk_widget.c      # GtkWidget, GtkContainer, GtkGrid (~45KB)  
├── gtk_window.c      # GtkWindow, GtkDialog, GtkNotebook (35KB)
├── gtk_controls.c    # GtkButton, GtkEntry, GtkLabel, etc (40KB)
├── gtk_multimedia.c  # GStreamer wrappers (50KB)
├── gtk_dbus.c       # D-Bus related wrappers (30KB)
└── gtk_common.h     # Shared definitions and macros
```

Benefits:

- **Parallel compilation**: 6 files can compile simultaneously instead of 1 sequential bottleneck
- **Reduced memory pressure**: Each compilation unit ~40-50KB instead of 252KB
- **Faster incremental builds**: Changes only recompile affected modules
- **Architecture compatibility**: Eliminates hangs on ppc64le and other memory-constrained systems
- **Logical organization**: GTK class families grouped together
- **Easier debugging**: Compilation errors isolated to specific modules
- **Selective optimization**: Can apply different compiler flags per module if needed
- **Future extensibility**: New GTK classes can be added to appropriate modules

I am more than happy to give this a go if you like the idea @ptitSeb

ptitSeb:
Are you sure you can split without changing how things works? I'm aware of the gtkclass issue, so I'm fine with splitting it, but it does require a lot of work, there is a ton of macro and many static functions (that should stay static).

runlevel5:
Let me give a go.

runlevel5:
@ptitSeb I tried but failed to produce any PoC, this is way too big than I imagine. Let's revisit this work in the future

PR #3642, opened 2026-03-11
[REFACTOR] Split monolithic gtkclass.c into many files

runlevel5:
Status: DRAFT (please do not review it yet)

~2.5x faster build time

I've tested with GTK3 apps and Steam on my ARM64 and PPC64LE box. Having said that I would love to ask for more testing from other contributors (especially those who have access to LA64 or RV64 machines)

Fixes https://github.com/ptitSeb/box64/issues/3636


ptitSeb:
I don't understand. You moved the code inside separate headers, but the C file just do the same as before, and include all those headers at once... The resulting compilation object should be exactly the same as before after the pre-processor phase. I don't see how this would increase  compilation performances.

runlevel5:
> I don't understand. You moved the code inside separate headers, but the C file just do the same as before, and include all those headers at once... The resulting compilation object should be exactly the same as before after the pre-processor phase. I don't see how this would increase compilation performances.

let me check, I think there must be some cache on my side that gives me wrong impression it would help reduce the compilation time

runlevel5:
Yes you are right. It is a stale cache on my side that gives me false pictures of my changes. I am going to retry with different approach (or find out if it is not possible)

PR #3645, opened 2026-03-11
[TEST] Added first unittest

ksco:
This may rely on PR #3643 and #3644 being merged.  Covers what's fixed in #3644, make sure similar mistake won't happen again.


ksco:
CI is not happy. I need some time to debug. Closing for now.

ptitSeb:
maybe nasm is missing or something like that?

ksco:
Ohhh, indeed, let me try again with nasm.

ksco:
@ptitSeb it's still not happy on Arm, can you test this locally? I don't have a Arm machine at hand right now.

ptitSeb:
I tested the nasm file localy, and it's working fine.
Note that RISC-V configuration are also red, not only ARM64 ones.

ksco:
Okay, I'll find some time to debug CI... maybe tomorrow..

PR #3650, opened 2026-03-11
 [PPC64LE_DYNAREC] Fix SCRATCH_USAGE placement for MOV opcodes (0xC6, 0xC7)

runlevel5:
- Remove unnecessary `SCRATCH_USAGE(1)` from the `MODREG` (register) path of opcode 0xC6 (MOV Eb, Ib), since scratch register `x3` is only needed alongside `geted()` in the memory path


runlevel5:
Closed in favor of https://github.com/ptitSeb/box64/pull/3652

Issue #2608, opened 2025-05-07
[NON4KPAGE] AllocLoadElfMemory multiblocks wrong alignment


xiangzhai:
Hi,

[jtreg testcase](https://github.com/openjdk/jdk/blob/master/test/hotspot/jtreg/vmTestbase/vm/jit/LongTransitions/libLTTest.cpp#L3397) `objdump -d /your/path/jdk/build/linux-x86_64-server-release/images/test/hotspot/jtreg/native/libLTTest.so `
```
0000000000046b50 <Java_vm_jit_LongTransitions_LTTest_nativeFnc52>:
   46b50:       55                      push   %rbp
...
=>   47fff:       b8 01 00 00 00          mov    $0x1,%eax
                     ^--- expected
   48004:       e8 67 a0 fb ff          callq  2070 <fprintf@plt>
...
```

[AllocLoadElfMemory multiblocks wrong alignment](https://github.com/ptitSeb/box64/blob/main/src/elfs/elfloader.c#L302) failed to decode `0x3f00047fff` expected `B8 01 00 00 00`:
```
...
[BOX64] New Instruction x64:0x3f00047fff, native:0xffc8d68628
=> [BOX64] 0x3f00047fff: B8 00 00 00 00  MOV Reg, Id
                            ^--- wrong
[BOX64] 0xffc8d68628: 1 emitted opcodes, inst=631, barrier=0 state=0/1(0), set=0/0, use=0, need=80/80, fuse=0, sm=0(0/0), pred=630, last_ip=0x3f00047fed Q0:XMM0
        0380000c        ORI             xRAX_r12, xZR, 0x0
...
```

After changed the `balign` with `box64_pagesize`:
```
diff --git a/src/elfs/elfloader.c b/src/elfs/elfloader.c
index 2dadbda4..91b1df2e 100644
--- a/src/elfs/elfloader.c
+++ b/src/elfs/elfloader.c
@@ -299,8 +299,8 @@ int AllocLoadElfMemory(box64context_t* context, elfheader_t* head, int mainbin)
             uint8_t prot = ((e->p_flags & PF_R)?PROT_READ:0)|((e->p_flags & PF_W)?PROT_WRITE:0)|((e->p_flags & PF_X)?PROT_EXEC:0);
             // check if alignment is correct
             uintptr_t balign = head->multiblocks[n].align-1;
-            if(balign<4095) balign = 4095;
[...]

PR #3657, opened 2026-03-12
[ELFLOADER] handle PT_LOAD host-page overlaps on NON4KPAGE hosts

yzewei:
## Problem

On non-4K page systems, it's possible for adjacent `PT_LOAD` segments in x86_64 ELF to fall into the same host page.

For example, in an x86 Node.js application, there might be a situation where an `R|X` segment and an adjacent `R` segment share the same 16K page.

Under the current logic, inconsistencies between host page permissions and segment-level permissions occur during loading, ultimately leading to SIGSGEV.

## Solution Modification

For this type of shared host page situation: adopt a permission stacking strategy.

## Modifications

Main Changes:

- Added a local helper for host page scope and ELF flags to host `PROT_*`

- On NON4K page systems, added `ElfPhdrNeedsCopyLoad64()` checks:

- Force copy-load if the current `PT_LOAD` does not completely cover the host page.

- Force copy-load if the host page covered by the current `PT_LOAD` overlaps with other `PT_LOAD`s.

- Retain the original mmap fast path, but only for segments that "completely cover the host page and do not share the host page with other `PT_LOAD`".

- After all segments are loaded, recalculate the final permission union by host page and uniformly execute `mprotect`.

## Impact

For 4K page systems, existing paths remain unchanged.

For NON4K page systems, only insecure `PT_LOAD`s will revert to copy-load; the original fast path will still be preserved as much as possible. ## Verification

I verified the following issue locally:

- The startup issue of `node --version` on the LoongArch big page system has been fixed.
- download [nodejs](https://nodejs.org/dist/v24.14.0/node-v24.14.0-linux-x64.tar.xz) 
- PT_LAOD like this:

```
  LOAD           0x0000000002c00000 0x0000000003000000 0x0000000003000000
                 0x00000000000006a1 0x00000000000006a1  R E    0x200000
  LOAD           0x0000000002c01000 0x0000000003001000 0x0000000003001000
                 0x000000000379c6f2 0x000000000379c6f2  R      0x1000

```

### DIFF
--- a/src/elfs/elfloader.c
+++ b/src/elfs/elfloader.c
@@ -211,6 +211,57 @@ const char* ElfPath(elfheader_t* head)
     return head->path;
 }
 
+static uintptr_t ElfPageStart(uintptr_t addr)
+{
+    return addr & ~(box64_pagesize-1);
+}
+
+static uintptr_t ElfPageEnd(uintptr_t addr, size_t size)
+{
+    return ALIGN(addr + size);
+}
+
+static uint32_t ElfFlagsToProt(uint32_t flags)
+{
+    uint32_t prot = 0;
+    if(flags & PF_R) prot |= PROT_READ;
+    if((flags & PF_W) || box64_unittest_mode) prot |= PROT_WRITE;
+    if(flags & PF_X) prot |= PROT_EXEC;
+    return prot;
+}
+
+static int ElfPhdrNeedsCopyLoad64(elfheader_t* head, int phidx, uintptr_t offs)
+{
+    Elf64_Phdr* e = &head->PHEntries._64[phidx];
+    if(box64_pagesize <= 4096)
+        return 0;
+
+    uintptr_t seg_addr = e->p_paddr + offs;
+    uintptr_t seg_addr_end = seg_addr + e->p_memsz;
+    uintptr_t host_start = ElfPageStart(seg_addr);
+    uintptr_t host_end = ElfPageEnd(seg_addr, e->p_memsz);
+
+    // Partial host-page coverage must be copy-loaded so adjacent PT_LOADs can
+    // share the host page safely while we defer the final protection fixup.
+    if(seg_addr != host_start || seg_addr_end != host_end)
+        return 1;
+
+    for(size_t i = 0; i < head->numPHEntries; ++i) {
+        Elf64_Phdr* other = &head->PHEntries._64[i];
+        if(i == phidx || other->p_type != PT_LOAD || !other->p_flags || !other->p_memsz)
+            continue;
+
+        uintptr_t other_addr = other->p_paddr + offs;
+        uintptr_t other_host_start = ElfPageStart(other_addr);
+        uintptr_t other_host_end = ElfPageEnd(other_addr, other->p_memsz);
+
+        if(other_host_end > host_start && other_host_start < host_end)
+            return 1;
+    }
+
+    return 0;
+}
+
 int AllocLoadElfMemory32(box64context_t* context, elfheader_t* head, int mainbin)
 #ifndef BOX32
 { return 1; }
@@ -300,7 +351,7 @@ int AllocLoadElfMemory(box64context_t* context, elfheader_t* head, int mainbin)
             head->multiblocks[n].size = e->p_filesz;
             head->multiblocks[n].align = e->p_align;
             // HACK: Mark all the code pages writable in unittest mode because some tests mix code and (writable) data...
-            uint8_t prot = ((e->p_flags & PF_R)?PROT_READ:0)|(((e->p_flags & PF_W) || box64_unittest_mode)?PROT_WRITE:0)|((e->p_flags & PF_X)?PROT_EXEC:0);
+            uint8_t prot = ElfFlagsToProt(e->p_flags);
             // check if alignment is correct
             uintptr_t balign = head->multiblocks[n].align-1;
             if (balign < (box64_pagesize - 1)) balign = box64_pagesize - 1;
@@ -315,6 +366,8 @@ int AllocLoadElfMemory(box64context_t* context, elfheader_t* head, int mainbin)
             if(!e->p_filesz)
                 try_mmap = 0;
             if(e->p_align<box64_pagesize)
+                try_mmap = 0;
+            if(ElfPhdrNeedsCopyLoad64(head, i, offs))
                 try_mmap = 0;
             if(try_mmap) {
                 printf_dump(log_level, "Mmaping 0x%lx(0x%lx) bytes @%p with prot %x for Elf \"%s\"\n", head->multiblocks[n].size, head->multiblocks[n].asize, (void*)head->multiblocks[n].paddr, prot, head->name);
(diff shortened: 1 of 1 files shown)


yzewei:
Additional note: deferred mprotect cannot resolve this issue; it cannot recover the overwritten EXEC statements.

xiangzhai:
Hi,

I just only fixed one NON4KPAGE [issue](https://github.com/ptitSeb/box64/issues/2608) I often confused by other NON4KPAGE issues https://github.com/ptitSeb/box64/issues?q=NON4KPAGE  so I just reviewed the patch and asked  @yzewei  (might not related) question: what if `.text` section overlap `.got` section? https://github.com/ptitSeb/box64/issues/2579#issuecomment-2885557266  

RFC to other NON4KPAGE hunters.

Thanks,
Leslie Zhai

src/elfs/elfloader.c
```diff
@@ -426,17 +479,50 @@ int AllocLoadElfMemory(box64context_t* context, elfheader_t* head, int mainbin)
...
-    // PROT_WRITE before a later segment that shares the same page has been read into memory
-    for (int j = 0; j < n; j++) {
-        if(!(head->multiblocks[j].flags & PF_W)) {
-            uintptr_t start = head->multiblocks[j].paddr & ~(box64_pagesize-1);
-            uintptr_t end = ALIGN(head->multiblocks[j].paddr + head->multiblocks[j].asize);
-            for(uintptr_t page = start; page < end; page += box64_pagesize) {
-                uint32_t prot = getProtection(page);
-                if(prot && !(prot & PROT_WRITE))
-                    mprotect((void*)page, box64_pagesize, prot & ~PROT_CUSTOM);
+    if(box64_pagesize > 4096) {
```
xiangzhai: > Additional note: deferred mprotect cannot resolve this issue; it cannot recover the overwritten EXEC statements.

I also asked why `deferred mprotect` is not able to cover the `if(box64_pagesize > 4096)`.

xiangzhai: I hold the view that it is better to recover the overwritten EXEC for `deferred mprotect` and merge the `if(box64_pagesize > 4096)` to `deferred mprotect`.

xiangzhai (changes requested):

src/elfs/elfloader.c
yzewei: Instead of checking page size, the system will handle all cases requiring permission stacking directly, without using the get_protection function.

src/elfs/elfloader.c
```diff
@@ -426,17 +479,38 @@ int AllocLoadElfMemory(box64context_t* context, elfheader_t* head, int mainbin)
...
     }
-    // deferred mprotect: apply final protections after all segments are loaded
-    // this avoids the case where mprotect on a shared host page (e.g. 64KB) strips
-    // PROT_WRITE before a later segment that shares the same page has been read into memory
-    for (int j = 0; j < n; j++) {
-        if(!(head->multiblocks[j].flags & PF_W)) {
-            uintptr_t start = head->multiblocks[j].paddr & ~(box64_pagesize-1);
-            uintptr_t end = ALIGN(head->multiblocks[j].paddr + head->multiblocks[j].asize);
-            for(uintptr_t page = start; page < end; page += box64_pagesize) {
-                uint32_t prot = getProtection(page);
```
xiangzhai, line 437: Request for other reviewers due to moved `getProtection`.

yzewei:
@runlevel5 I've modified the logic here; could you please review it and see if there are any potential negative impacts?

src/elfs/elfloader.c
yzewei: @runlevel5

runlevel5: @yzewei let me have a look tomorrow

src/elfs/elfloader.c
```diff
@@ -211,6 +211,57 @@ const char* ElfPath(elfheader_t* head)
...
+{
+    return addr & ~(box64_pagesize-1);
+}
+
+static uintptr_t ElfPageEnd(uintptr_t addr, size_t size)
+{
+    return ALIGN(addr + size);
+}
+
+static uint32_t ElfFlagsToProt(uint32_t flags)
```
runlevel5, line 224: Good refactoring +1

src/elfs/elfloader.c
```diff
@@ -426,17 +479,38 @@ int AllocLoadElfMemory(box64context_t* context, elfheader_t* head, int mainbin)
...
+    
+    // Perform final page protection after all PT_LOAD segments have been loaded.
+    // Recalculate from the loaded blocks instead of calling getProtection(page),
+    // Maintain the cumulative combination of all segment flags for overlapping pages.
+    if(n) {
+        uintptr_t min_hostpage = UINTPTR_MAX;
+        uintptr_t max_hostpage = 0;
+        for(int j = 0; j < n; ++j) {
+            uintptr_t seg_page_start = ElfPageStart(head->multiblocks[j].paddr);
+            uintptr_t seg_page_end = seg_page_start + head->multiblocks[j].asize;
```
runlevel5, line 491: A bit of inconsistency. I can see `asize` was computed from the `unaligned paddr`:

```
asize = (e->p_memsz + (e->p_paddr & balign) + (box64_pagesize - 1)) & ~(box64_pagesize - 1);
```

`asize` already accounts for the sub-page offset of `paddr`. When you add it to `ElfPageStart(paddr)` (which rounds down), this could overshoot by up to one page.

Consider whether this should instead be:

```
uintptr_t seg_page_end = ElfPageEnd(head->multiblocks[j].paddr, head->multiblocks[j].asize);
```

Or more precisely, using the actual segment extent: `ElfPageEnd(paddr, original_memsz)`. The current calculation might set protections on one extra page past the segment's actual range. In practice this is likely harmless (it's within the overall ELF reservation) but is logically imprecise.

src/elfs/elfloader.c
```diff
@@ -426,17 +479,38 @@ int AllocLoadElfMemory(box64context_t* context, elfheader_t* head, int mainbin)
...
+                    continue;
+                uint32_t seg_prot = ElfFlagsToProt(head->multiblocks[j].flags);
+                final_prot |= seg_prot;
+                covered = 1;
+            }
+            if(!covered) continue;
+            setProtection_elf(page, box64_pagesize, final_prot);
+            if(mprotect((void*)page, box64_pagesize, final_prot & ~PROT_CUSTOM) != 0) {
+                printf_log(LOG_INFO, "Warning, mprotect failed on ELF host page %p (%s)\n",
+                    (void*)page, strerror(errno));
```
runlevel5: hmm I can see that `setProtection_elf()` is called before `mprotect()`. If `setProtection_elf` modifies `errno` internally (e.g., through memory allocation or locking), the `strerror(errno)` may not reflect the `mprotect` failure

runlevel5:
I am a bit concerned there might be possible regression for 4K pagesize system with unconditional `mprotect` on all pages. The old code only called `mprotect` for non-writable segments `(!(flags & PF_W))`. The new code calls `mprotect` on every page of every segment unconditionally. On 4K-page systems where segments don't overlap, this means extra `mprotect` syscalls for writable segments that previously didn't need them. The setProtection_elf + mprotect pair is called for every covered host page, even when the segment was already mmap'd with the correct protections. Though I think the performance impact if there are extra sys calls during ELF load. On second thought it is unlikely to be measurable.

runlevel5:
@yzewei @xiangzhai please test on 4K-pag systems to confirm no regressions (the `ElfPhdrNeedsCopyLoad64` early-return handles this, but the mprotect loop change affects all page sizes)

src/elfs/elfloader.c
yzewei: That's right. I reviewed the code again.

Here, `asize` is intentionally set to the host page-aligned coverage size of the loaded block, which I use as the expanded host page range in the final processing stage.

Using `ElfPageEnd(paddr, asize)` here might treat `asize` as the original unaligned segment size, which could actually expand the range to an additional host page in cases like overlapping PT_LOAD layouts that trigger this work.

A more explicit long-term cleanup approach would be to drive final processing directly from the original PT_LOAD range (`p_paddr + offs`, `p_memsz`, `p_flags`), rather than from `multiblocks[].asize`, but I can process this later. What do you think?

yzewei:
> I am a bit concerned there might be possible regression for 4K pagesize system with unconditional `mprotect` on all pages. The old code only called `mprotect` for non-writable segments `(!(flags & PF_W))`. The new code calls `mprotect` on every page of every segment unconditionally. On 4K-page systems where segments don't overlap, this means extra `mprotect` syscalls for writable segments that previously didn't need them. The setProtection_elf + mprotect pair is called for every covered host page, even when the segment was already mmap'd with the correct protections. Though I think the performance impact if there are extra sys calls during ELF load. On second thought it is unlikely to be measurable.

Yes, that's correct.

Because of the unified final protection process, 4K hosts will now also undergo the same final recalculation step. Therefore, this may introduce an additional `mprotect()` system call compared to the previous delayed path that only used `!(PF_W)`.

The purpose of this approach is to unify the final protection model for 4K and non-4K hosts and to stop relying on the old `getProtection(page)` replay logic. So this ensures that the goal is achieved correctly.

It only happens during ELF loading, not during stable execution, so the performance impact should be minimal.

yzewei:
> @yzewei @xiangzhai please test on 4K-pag systems to confirm no regressions (the `ElfPhdrNeedsCopyLoad64` early-return handles this, but the mprotect loop change affects all page sizes)

Agreed.

4K regression testing is indeed necessary.

I've only reproduced and verified the non-4K scenario locally so far. However, I don't have a suitable 4K page environment on hand. Could you please help me verify it? Thank you very much!

src/elfs/elfloader.c
yzewei: That's right. I modified the code to save `errno` immediately after `mprotect()` fails, and then format the log message.

runlevel5:
> @yzewei I've only reproduced and verified the non-4K scenario locally so far. However, I don't have a suitable 4K page environment on hand. Could you please help me verify it? Thank you very much!

I tested on ARM64 4k and 16k, the ctest all passed and `node --version` runs correctly.
I also tested on PPC64LE 64k, the ctest too passed and `node --version` crashed (that is totally different issue) but I could confirm ELF loads correctly (copy-load path)

yzewei:
> > @yzewei I've only reproduced and verified the non-4K scenario locally so far. However, I don't have a suitable 4K page environment on hand. Could you please help me verify it? Thank you very much!
> 
> I tested on ARM64 4k and 16k, the ctest all passed and `node --version` runs correctly. I also tested on PPC64LE 64k, the ctest too passed and `node --version` crashed (that is totally different issue) but I could confirm ELF loads correctly (copy-load path)

Thanks!

yzewei:
@ptitSeb What are your thoughts on this?

ptitSeb:
I need to think about it, ELF Loader is quite brittle for now.

PR #3661, opened 2026-03-12
[ARM64_DYNAREC] Fixed scratch register conflict for BZHI opcode

ksco:
GETED may use x1 as ed.


ptitSeb:
I suspect the previous commit about WoW64, that didn't trigger a new build, actualy broke the build...

ksco:
> I suspect the previous commit about WoW64, that didn't trigger a new build, actualy broke the build...

Oh shoot

PR #3668, opened 2026-03-13
[PPC64LE_DYNAREC] Add OR (0x08-0x0D), AND (0x20-0x25), XOR (0x30-0x35) and SUB (0x28-0x2D) opcodes and emitters

runlevel5:
Related to https://github.com/ptitSeb/box64/issues/242

### DIFF
--- a/src/dynarec/ppc64le/dynarec_ppc64le_00.c
+++ b/src/dynarec/ppc64le/dynarec_ppc64le_00.c
@@ -64,7 +64,6 @@ uintptr_t dynarec64_00(dynarec_ppc64le_t* dyn, uintptr_t addr, uintptr_t ip, int
             INST_NAME("ADD Eb, Gb");
             SETFLAGS(X_ALL, SF_SET_PENDING, NAT_FLAGS_FUSION);
             nextop = F8;
-            SCRATCH_USAGE(0);
             // Get Gb (source byte register)
             gd = ((nextop & 0x38) >> 3) + (rex.r << 3);
             if (rex.rex) {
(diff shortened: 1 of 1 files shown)


ptitSeb (approved):

src/dynarec/ppc64le/dynarec_ppc64le_00.c
```diff
@@ -200,6 +200,566 @@ uintptr_t dynarec64_00(dynarec_ppc64le_t* dyn, uintptr_t addr, uintptr_t ip, int
             i64 = F32S;
             emit_add32c(dyn, ninst, rex, xRAX, i64, x3, x4, x5, x6);
             break;
+        case 0x08:
+            INST_NAME("OR Eb, Gb");
+            SETFLAGS(X_ALL, SF_SET_PENDING, NAT_FLAGS_FUSION);
+            nextop = F8;
+            SCRATCH_USAGE(0);
```
ksco: Sorry, but I don't understand: why `SCRATCH_USAGE(0);` used here as well as all the uses below?

runlevel5: I followed the same pattern used by the existing ADD opcodes (0x00–0x05) — for example, `ADD Eb, Gb` at line 67 and `ADD Ed, Gd` at line 120 already use `SCRATCH_USAGE(0)` in the same way.

My understanding is that `SCRATCH_USAGE(0)` marks the instruction as not clobbering scratch registers, so `READFLAGS_FUSION` in pass0 can look back past it when fusing native flags.  I assumed it was the intended convention for these opcode handlers....

That said, I'm not 100% sure the usage is correct for byte-register ops since they do use x4/x5 for BF_EXTRACT. Could you clarify what the right approach should be here? Happy to adjust.

ksco: Mark scratch usage to 0 only makes sense for instructions that never affect eflags, like `MOV`, `LEA`, you can refer to the LoongArch DynaRec for this.

runlevel5:
@ksco could you please help review the PR again? Many thanks

runlevel5:
@ksco 

> For the switch case part, being nitpicking, it would be better to have some GETED, GETGD, GETEB, GETGB macros to eliminate repeated code, like other DynaRecs.

Yes, I am going to do that in the next PR

PR #3669, opened 2026-03-13
[PPC64LE_DYNAREC] Add AND (0x20-0x25) and XOR (0x30-0x35) opcodes and emitters

runlevel5:
I bases this branch on https://github.com/ptitSeb/box64/pull/3668 but it could be reviewed individually


ptitSeb:
The sub code appear in this PR and the previous one. Don't makes us check the same code twice...

PR #3679, opened 2026-03-19
[LA64_DYNAREC] Refactored AVX infrastructure

ksco:
This is a simplified version with minimal changes to the current implementation. By removing ymm zero tracking, it fixes the main issue, so we can debug AVX without worrying about infra bugs. A more sophisticated AVX infrastructure will need a full rewrite.


ptitSeb:
Wel, ymm zero tracking might not be relevant for an architecture featuring 256-bit with vector, and I assume opcodes that can 0 out the upper 128bit already.

ksco:
Well, no. 128-bit opcodes will put the high 128-bit in an undefined state, so the ymmzero tracking is necessary for performance.

ptitSeb:
Ok, so ymm 0 tracking will come back later.

PR #3683, opened 2026-03-19
[WRAPPER] Add wrapper for libmvec.so.1

runlevel5:
Wrap the SSE2 vectorized sin/cos functions from glibc libmvec using native scalar sin()/cos() from libm. The Broken Sword Reforged game happens to require these 2 functions. 

ref: https://github.com/ptitSeb/box86-compatibility-list/issues/549


ptitSeb:
It's acceptable, but barely. You unwrap the verctorized sin/cos instead of using the mvec functions themselves.

runlevel5:
@ptitSeb I've just amended the PR, that is to only to fallback to scalar path for non-aarch64/x86_64

src/wrapped/wrappedlibmvec.c
```diff
@@ -0,0 +1,86 @@
...
+        memcpy(&emu->xmm[0], &result, sizeof(v2df));
+    } else {
+        emu->xmm[0].d[0] = cos(emu->xmm[0].d[0]);
+        emu->xmm[0].d[1] = cos(emu->xmm[0].d[1]);
+    }
+}
+
+// Try to load the native libmvec.so.1 and resolve vectorized sin/cos.
+// On aarch64 and x86_64: native libmvec exists in glibc, use it.
+// On ppc64le and others: dlopen fails, fall back to scalar sin/cos.
```
runlevel5, line 73: Hopefully one day https://sourceware.org/pipermail/glibc-cvs/2020q1/068984.html it would be upstreamed for ppc64

PR #3689, opened 2026-03-20
[WRAPPER] ioctl number x86_64 -> ppc64le translation

runlevel5:
Return to Monkey Island game refuses to run, digging in a bit I realize ioctl number of ppc64le is totally different to that x86_64, aarch64.


ptitSeb:
Need to check if Loongarch & RISC-V don't need something like that @ksco

ksco:
Thank you for the reminder, but I think we are good here ;) https://elixir.bootlin.com/linux/v6.19.8/A/ident/TCGETS

ksco:
See it in action: https://godbolt.org/z/d3483j66a

PR #3698, opened 2026-03-22
[WRAPPER] Wrapped 1 more symbol for libc

ksco:
Used by CS:GO legacy


ptitSeb:
Static build kind-of red. I guess this needs some `#ifdef`...

PR #3707, opened 2026-03-25
Fix ogg_sync_pageout hook infinite recursion after 2df07b59

runlevel5:
- Fix infinite recursion in the `ogg_sync_pageout` hook caused by commit 2df07b59 (`[DYNAREC] Refactored how Alt are undled with Dynarec`)
- Add `suppressAlternate`/`unsuppressAlternate` API to temporarily bypass alternate resolution for a specific address

I have tested it on PPC64LE, but love to get help from other backend contributors to see my changes would cause any issue

### DIFF
--- a/src/include/alternate.h
+++ b/src/include/alternate.h
@@ -11,6 +11,8 @@ void* getAlternate(void* addr);
 void addAlternate(void* addr, void* alt);
 void addCheckAlternate(void* addr, void* alt);
 void cleanAlternate(void);
+void suppressAlternate(void* addr);
+void unsuppressAlternate(void* addr);
 #ifdef HAVE_ALTJUMP
 uintptr_t getAlternateJump(void* addr, int is32bits);
 #endif
--- a/src/mallochook.c
+++ b/src/mallochook.c
@@ -881,8 +881,12 @@ static void*     ogg_page_body_copy = NULL;
 
 static int my_ogg_sync_pageout_32(void* oy, void* og)
 {
-    // Call the original emulated ogg_sync_pageout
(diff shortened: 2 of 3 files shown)


ptitSeb:
Thanks for the issue, but this is absolutly not a fix.

The issue is infinite loops in memory allocators, and you create a workaround for a single ogg call? what about all the other call that also use  memory?

ptitSeb:
I'll push something more generic, as it can also happens with all hooked symbols from mallochook.c

ptitSeb:
I have pushed this https://github.com/ptitSeb/box64/commit/be199c581287b1c7d0ed665ce45b3db616722aac that should fix your issue with ogg and all the potnetial others with mallochook

ptitSeb:
Mmmm, I don't think this, or anything else, will work with the new alt handling, as the dynarec will always shadow the address. I might need to think of something more there. Maybe forcing the 1st step as interpreter?

ptitSeb:
I will probably work on a better solution latter, to avoid this passage via the Interpreter...

ptitSeb:
With https://github.com/ptitSeb/box64/commit/43dcbf3773a3770cb3d6fe0b802d7390b332ee9b I have added a way to get Dynarec used for Alt address. The issue should be complelty gone now, no need for this hack.

ptitSeb:
@runlevel5 I'm closing this PR, I think it's obsolete now. Do you confirm it's working fine with latest box64 now?

runlevel5:
@ptitSeb I will be back from my holiday tomorrow, please leave it to me to close this PR. I have not yet tested your fix yet

PR #3734, opened 2026-04-02
[INTERP] Fixed nan propagation edge-cases for various opcodes

ksco:


ptitSeb:
Why this change? isn't the NAN nature (quiet or signaling) part of the emulated data and not to be touched?

ksco:
> Why this change? isn't the NAN nature (quiet or signaling) part of the emulated data and not to be touched?

x86 will convert it to QNaN.

ptitSeb:
Really? Mmm... strange.

PR #3739, opened 2026-04-03
[LA64_DYNAREC] Reworked fastnan handling

ksco:


ksco:
Sometimes I forgot to run ctest, that's when it failed....

PR #3740, opened 2026-04-03
[INTERP][DYNAREC] Reworked PF computaion for ANDN opcode

ksco:


src/dynarec/la64/dynarec_la64_avx_0f38.c
```diff
@@ -78,6 +78,9 @@ uintptr_t dynarec64_AVX_0F38(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip,
                 BNEZ(gd, 8);
                 ORI(xFlags, xFlags, 1 << F_ZF);
             }
+            if (BOX64DRENV(dynarec_safeflags)) {
+                IFX (X_PF) emit_pf(dyn, ninst, gd, x2, x5);
```
ptitSeb, line 82: no cputype check here? also, no AF flags handling?

ksco: AF is cleared by `CLEAR_FLAGS()` above; cputype is still unsupported in LA64 DynaRec.

ptitSeb: Ok, all good then.

PR #3802, opened 2026-04-22
[WRAPPER] Fixed more wrapping issues

ksco:
Meh...


src/wrapped/wrappedlibblas_private.h
ksco, line 1: This one is strange, because I noticed it's kinda newly wrapped. So I suppose it was wrapped with the wrapperhelper?

ptitSeb: Well, openblas_config.h have this:

```C
#ifdef OPENBLAS_USE64BITINT
typedef BLASLONG blasint;
#else
typedef int blasint;
#endif
```
And `OPENBLAS_USE64BITINT` is defined for me in my openblas_config.h line 21 for the aarch64 architecture

ptitSeb: So, the long instead of int seems correct from my point of view

ksco: I rolled back changes to this file.

src/wrapped/wrappedzstd_private.h
```diff
@@ -248,7 +248,7 @@
 //GO(ZBUFFv07_recommendedDInSize, 
 //GO(ZBUFFv07_recommendedDOutSize, 
 //GO(ZDICT_addEntropyTablesFromBuffer, 
-GO(ZDICT_finalizeDictionary, LFpLpLppuH) // last param afet LFpLpLppu is ZDICT_params_t, a struct with iuu
+GO(ZDICT_finalizeDictionary, LFpLpLppuH)
 GO(ZDICT_getDictHeaderSize, LFpL)
```
ptitSeb, line 252: can we keep this comment?

ksco: Yes, addressing it now.

src/wrapped/wrappedlibsm_private.h
```diff
@@ -41,7 +41,7 @@ GO(SmsReturnProperties, vFpip)
 GO(SmsSaveComplete, vFp)
 GO(SmsSaveYourself, vFpiiii)
 GO(SmsSaveYourselfPhase2, vFp)
-//GOM(SmsSetErrorHandler, BFEB)
+//GOM(SmsSetErrorHandler, pFEp)
 GO(SmsShutdownCancelled, vFp)
 
-GO(dummy_vFpiiLiip, vFpiiLiip) // for error handler
```
ptitSeb, line 47: Can we keep this comment

PR #3803, opened 2026-04-22
[WRAPPERHELPER] Added a new --check-only option

ksco:


wrapperhelper/src/main.c
```diff
@@ -53,6 +54,7 @@ enum bits_state {
 };
 
 int is_gst = 0;
+int is_check_only = 0;
```
rajdakin: Why is this a global variable? (Note that `is_gst` is actually declared in `parse.c` as `extern int is_gst`, so it is necessarily a global, but you give `is_check_only` as an argument to the functions. The other variables are declared as local to `main`, like `ms`.)

rajdakin (changes requested):
LGTM except for that global variable.

rajdakin (approved):

PR #3822, opened 2026-05-06
[DYNAREC] Restored sep entries hidden dynablock reference

ksco:
This fixes Linux Steam SIGBUS issues with DynaCache.

### DIFF
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -1498,6 +1498,10 @@ int MmaplistAddBlock(mmaplist_t* list, int fd, off_t offset, void* orig, size_t 
             // adjust guest source addresses with delta_map
             bl->x64_addr += delta_map;
             bl->x64_readaddr += delta_map;
+            for (int j = 0; j < bl->sep_size; ++j) {
+                // SEP native entries also carry a hidden dynablock reference.
+                *(dynablock_t**)(bl->block + bl->sep[j].nat_offs - sizeof(void*)) = bl;
+            }
             *(uintptr_t*)(bl->jmpnext+2*sizeof(void*)) = RelocGetNext();
             if(bl->relocs && bl->relocsize)
                 ApplyRelocs(bl, delta, delta_map, mapping_start);
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -1514,8 +1514,7 @@ int MmaplistAddBlock(mmaplist_t* list, int fd, off_t offset, void* orig, size_t 
             } else {
                 for(int i=0; i<bl->sep_size; ++i) {
                     uint32_t x64_offs = bl->sep[i].x64_offs;
-                    uint32_t nat_offs = bl->sep[i].nat_offs;
-                    if(addJumpTableIfDefault64(bl->x64_addr+x64_offs, (bl->dirty || bl->always_test)?bl->jmpnext:(bl->block+nat_offs)))
+                    if (addJumpTableIfDefault64(bl->x64_addr + x64_offs, bl->jmpnext))
                         bl->sep[i].active = 1;
                     else
                         bl->sep[i].active = 0;


ptitSeb:
Looks good, but I don't see where the new `db->sep_nolink` is set/unset in the PR?

ksco:
> Looks good, but I don't see where the new `db->sep_nolink` is set/unset in the PR?

Oops, I think I messed up this patch when trying to do a rebase with conflicts...

ksco:
Okay I added that line back.

src/custommem.c
```diff
@@ -1498,6 +1498,7 @@ int MmaplistAddBlock(mmaplist_t* list, int fd, off_t offset, void* orig, size_t
             // adjust guest source addresses with delta_map
             bl->x64_addr += delta_map;
             bl->x64_readaddr += delta_map;
+            bl->sep_nolink = (bl->sep_size > 0);
```
ptitSeb: so, once set, this is never unset? so SEP will never be used again?

ksco: For blocks loaded from cache, yes. But the direct addressing of SEP can be available again via the `jmpnext -> LinkNext()` path.

src/libtools/signals.c
```diff
@@ -817,7 +817,8 @@ void my_box64signalhandler(int32_t sig, siginfo_t* info, void * ucntx)
                         for(int i=0; i<db->sep_size; ++i) {
                             uint32_t x64_offs = db->sep[i].x64_offs;
                             uint32_t nat_offs = db->sep[i].nat_offs;
-                            if(addJumpTableIfDefault64(db->x64_addr+x64_offs, (db->always_test)?db->jmpnext:(db->block+nat_offs)))
+                            void* sep_target = (db->always_test || db->sep_nolink) ? db->jmpnext : (db->block + nat_offs);
```
ptitSeb: I don't understand why sep_nolink should be used here? If the block as been checked as good, sep_nolink should be removed, the block is valid...

src/dynarec/dynablock.c
```diff
@@ -424,7 +424,8 @@ dynablock_t* DBSwitchPrevious(x64emu_t* emu, dynablock_t* db, uintptr_t addr, in
         for(int i=0; i<db->sep_size; ++i) {
             uint32_t x64_offs = db->sep[i].x64_offs;
             uint32_t nat_offs = db->sep[i].nat_offs;
-            if(addJumpTableIfDefault64(db->x64_addr+x64_offs, (db->always_test)?db->jmpnext:(db->block+nat_offs)))
+            void* sep_target = (db->always_test || db->sep_nolink) ? db->jmpnext : (db->block + nat_offs);
```
ptitSeb: same question here. I'm not sure I understand. If we are here, the block is valid. So the Main Entry Point & all Secondary Entry Point should be good

ptitSeb:
Are you sure just the handling in custommem.c/MmaplistAddBlock is not enough (even without db->sep_nolink)?

ksco:
Thanks for the review. Restoring hidden dynablock* headers for SEP fixes Steam issue already, actually. I might have made some mistakes when testing. SEP offsets are perfectly safe to use! I have force-pushed and updated the PR title.

src/custommem.c
ptitSeb: I think this line also need to setup jmpnext unconditionnaly, like for the MEP.

ptitSeb:
> Thanks for the review. Restoring hidden dynablock* headers for SEP fixes Steam issue already, actually. I might have made some mistakes when testing. SEP offsets are perfectly safe to use! I have force-pushed and updated the PR title.

Yeah, the SEP+DynaCache has not be tested at all. My fault, I should have done a small test campaign, so expect some more issue there :'(

ptitSeb:
Looks good now, thanks! And sorry for the chaotic review...

PR #3824, opened 2026-05-06
[WRAPPER] Add gnutls_free wrapped function for libgnutls

zengdage:
`gnutls_free` is a function pointer. If it were defined as an ordinary function, it would trigger a signal 11 error in `CheckExec`.


ptitSeb:
Looks correct to me.

Can I ask in what case you needed that change?

zengdage:
> Looks correct to me.
> 
> Can I ask in what case you needed that change?

When running Feishu, a communication software developed by Alibaba, on the RISC-V platform using Box64, an error occurs:

`Emit Signal 11 at IP=0x34b68c404353cc(??? / ?) / addr=0x34b68c404353cc, code=0xecec.`

After analysis, the root cause is that gnutls_free, a global function pointer variable, is directly invoked as an ordinary function. This results in a failure in CheckExec.

ptitSeb:
> > Looks correct to me.
> > Can I ask in what case you needed that change?
> 
> When running Feishu, a communication software developed by Alibaba, on the RISC-V platform using Box64, an error occurs:
> 
> `Emit Signal 11 at IP=0x34b68c404353cc(??? / ?) / addr=0x34b68c404353cc, code=0xecec.`
> 
> After analysis, the root cause is that gnutls_free, a global function pointer variable, is directly invoked as an ordinary function. This results in a failure in CheckExec.

I see. Thanks for answering, and nice debug!

PR #3834, opened 2026-05-09
[LA64_DYNAREC] Introduced a dynarec version of the UpdateFlags helper function

ksco:


ptitSeb:
Nice, this should bring some small speedup.


Side remark: no CPUTYPE handling yet?

ksco:
> Side remark: no CPUTYPE handling yet?

yeah because it's not supported by LoongArch DynaRec (yet).

PR #3836, opened 2026-05-09
[DOCS] Mark certain options as fragile or legacy

ksco:


ptitSeb:
I think `BOX64_DYNAREC_PAUSE` should also be tagged as fragile or legacy.

The idea there would be to remove DynaCache setting from this section before the end of the version tho.

ksco:
You will move PAUSE?

ptitSeb:
> You will move PAUSE?

Later, it's not urgent.

PR #3845, opened 2026-05-11
[CORE] Fix unexpected modification of GetNativeName's return value

zengdage:
When I enable `BOX64_ROLLING_LOG`, I observe an error that triggers a software exception in `x64Int3`. The direct cause of the error is that the value of `R_RSI` or `R_RDI` is passed to the `snprintf` function as a string pointer (e.g., line 183 in src/emu/x64int3.c), but the value of `R_RSI` or `R_RDI` at this time is not a valid string pointer.

Initially, I thought the values of `R_RSI` or `R_RDI` were incorrect, but after analysis, I found that when the error occurs, the actual function pointer, `a` is `my___tls_get_addr` instead of `fopen`, so the values of `R_RSI` and `R_RDI` are correct, and the symbol name is wrong.

Furthermore, I discovered that the value of the symbol name changes during the execution of the `x64Int3` function: it shifts from `__tls_get_addr` to `fopen` or other values. The root cause of this change is that a static array is used to store the symbol name in `GetNativeName`, leading to data tampering in a multi-threaded environment.

I changed the `static char buff` to `static __thread char native_name` to avoid multi-threaded data tampering.

### DIFF
--- a/src/os/os_linux.c
+++ b/src/os/os_linux.c
@@ -106,9 +106,10 @@ const char* GetBridgeName(void* p)
     return getBridgeName(p);
 }
 
+static __thread char native_name[500] = { 0 };
+
 const char* GetNativeName(void* p, int lib)
 {
-    static char buff[500] = { 0 };
     {
         const char* n = GetBridgeName(p);
         if (n)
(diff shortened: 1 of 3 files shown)


ptitSeb:
That was bothering me for some time, but not enough so I start debugging it. Thanks.

PR #3849, opened 2026-05-11
[INTERP][DYNAREC] Fixed XSAVE/XRSTOR opcodes

ksco:


ptitSeb:
Mmmm, are you sure it's a fix? I think I checked this behaviour and that was the correct handling to check rex.w (wich is 0 on 32bit).

ksco:
The XSAVE layout is mostly the same as FXSAVE layout, and here is the FXSAVE layout for rex.w == 0:  https://www.felixcloutier.com/x86/fxsave#tbl-3-47

ksco:
> that was the correct handling to check rex.w (wich is 0 on 32bit).

The issue is on the 64-bit side; rew.x or not has no effects on XMM registers.

ptitSeb:
If you check with Zydis, a rex.w version is XSAVE64 (`48 0f ae 60 00 00 00`), while the regular is XSAVE (without the 48 prefix)

ksco:
Yes, and they are identical regarding the XMM register spill/restore.

ptitSeb:
There might be an issue with FXSAVE/FXRESTR actualy, as it also seems to go to 64 variant only with REX.W set.

ptitSeb:
image
there is comment in Intel doc about rex.w effect

ksco:
image

ksco:
rex.w or not only effects x87 stuff, not XMM registers, I'm 100% sure this is a legit fix.

ptitSeb:
Ok, read the doc, looks good indeed.
I still remember having issue with Wow64 stuffs with this... Maybe we need a non rex.w version of the 64bits save/restore to stay compatible with 32bits version of the opcode (for the x87 part)?

ksco:
Maybe, as the layout is indeed different.

ksco:
I added a test that would fail without this patch.

ptitSeb:
Thnaks, but that is not my concern. I agree that the 64bit version of XSAVE/XRSTOR need to save/restore all 16 XMM regs, that this PR fixes. But my issue is that the 64bits "short" version of XSAVE/XRSTOR cannot be used with 32bits version of XRSTOR/XSAVE without potentialy breaking x87 state.

ptitSeb:
Mmmm, after rechecking, the x87 structure difference between 32bit & 64bit seems non-relevant to what box64 save/restore.

PR #3852, opened 2026-05-12
[DYNAREC] Reflected x87 state before FXAM

ksco:
FXAM can be used to determine if the x87 stack is empty, purge the x87 cache before FXAM so it classifies the reflected x87 state.

### DIFF
--- a/src/dynarec/arm64/dynarec_arm64_d9.c
+++ b/src/dynarec/arm64/dynarec_arm64_d9.c
@@ -130,7 +130,7 @@ uintptr_t dynarec64_D9(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
             break;
         case 0xE5:
             INST_NAME("FXAM");
-            #if 1
+            x87_purgecache(dyn, ninst, 0, x1, x2, x3);
             i1 = x87_get_current_cache(dyn, ninst, 0, NEON_CACHE_ST_D);
             // value put in x14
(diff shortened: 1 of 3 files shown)


src/dynarec/arm64/dynarec_arm64_d9.c
```diff
@@ -130,7 +130,7 @@ uintptr_t dynarec64_D9(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
             break;
         case 0xE5:
             INST_NAME("FXAM");
-            #if 1
+            x87_purgecache(dyn, ninst, 0, x1, x2, x3);
```
ptitSeb, line 133: Why are you purging the cache? the next line is there to handle the case were the value is in the cache or not (`i1==-1` if it's not in the cache) specificaly to avoid purging the cache, as it kills performances (and this opcode is used quite often in legacy 32bit games)

ksco: Oh I didn't recognize that the code below addresses the same issue. I think the real issue here is `i1` only indicates the compilation time state, not the runtime. So if FXAM is used to examine the empty state in a loop with stack popping, this will break.

ptitSeb: But, why this opcode would be different to any other x87 opcode in the same loop?
Do you have some test that shows an issue with this one?

PR #3856, opened 2026-05-12
dynarec/arm64: reduce AVX dispatch and cache transform overhead

necat101:
dynarec/arm64: reduce AVX dynarec dispatch and cache transform overhead

This trims ARM64 dynarec compilation overhead for AVX-heavy blocks without
changing AVX instruction semantics.

The AVX map/prefix dispatcher now uses a compact static dispatch table instead
of a long ordered branch chain. This path is reached for every AVX instruction
in each dynarec pass, so avoiding repeated conditionals helps low-end ARM64
devices during cache compilation.

The FPU cache transform path also now builds small target-cache presence masks
for XMM/YMM/MM/x87 entries. These masks avoid repeated linear scans of the
unchanged jump-target cache while deciding which cached values need to be
unloaded. YMM zero tracking was also tightened so zeroed upper halves that are
known unneeded at the branch target do not force an unnecessary transform.


ptitSeb:
Interesting stuffs here, thanks for the PR.

Our of curiosity, did you benchmark somethings with and without your change?

necat101:
unfortunately, the ARM64 device im working with is my S25+ running userLAnd ubuntu + VNC connected to my computer to test this, the only games i can test is COD bo1/2 and VRChat (which fails obviously due to EAC), i have much newer titles but my phone only has like 90gb of space left and nowhere near enough to test MK11, MK1, GTA 5 etc.. these optimizations are all completely experimental and all tests ran were CLI smoke tests with me physically testing the instructions as they get cached one by one. also, sorry for the late response, i was asleep, and i am currently at work when I had just seen you merged my PR! when i get home ill send the tests i had ran, and answer any additional questions you may have about this! :3

necat101:
The expected win is mostly in dynarec compilation overhead rather than steady-state generated-code speed:

- the AVX dispatch table removes the repeated prefix/map branch chain hit for every AVX instruction across the dynarec passes
- the cache-mask change avoids repeated linear scans of the target NEON cache during cache transforms
- the YMM-zero tweak can avoid a cache transform when the target already marks that upper state unneeded

So I expect this to help most with AVX-heavy block compilation / cache-build stutter on lower-end ARM64, not necessarily show up as a large FPS gain once blocks are already compiled.

ptitSeb:
I agree, but I was wondering if you saw some performances increase in some of the AVX tests of box64, that just compile the block for a 1 time execution for example.

necat101:
I benchmarked it after your question. I built two clean Release ARM64 binaries: the parent of the dynarec-cache optimization commit and the commit with the change applied.

For avx_intrinsics, both pass. I also ran tests/test30 directly through box64 with:

BOX64_DYNAREC=1
BOX64_DYNACACHE=0
BOX64_AVX=2
BOX64_DYNAREC_FASTNAN=0
BOX64_DYNAREC_FASTROUND=0
BOX64_LOG=0
BOX64_NOBANNER=1

That used 5 warmups and 100 measured alternating before/after runs, with stdout compared against ref30.txt every run.

Results were slightly better on this machine:
before mean 0.257366s, median 0.257570s
after  mean 0.246677s, median 0.244043s

I noticed a marginal ~4% improvement on my machine, but results may vary

ptitSeb:
Nice results. Thanks for the benchmark!

necat101:
also while benchmarking this, I had identified where i could optimize some non-AVX instructions with the same style of lookup table i did with AVX! im about to run some tests now before i make that PR! :3

necat101:
> Nice results. Thanks for the benchmark!

your very welcome! thank you very much again for this amazing open source project! im very sorry i got drunk the other day and linked my dads yt channel on the wrong PR >*<

ptitSeb:
Don't worry, it's ok.

PR #3861, opened 2026-05-13
Optimize ARM64 dynarec cache compilation for non-AVX paths

necat101:
This PR reduces cache compilation overhead on ARM64 dynarec paths used by older/non-AVX titles without changing guest-code compatibility behavior.

Changes include:

- Replaced repeated x64 prefix conditional checks with a prefix lookup table.
- Preserved exact 32-bit behavior for REX-byte handling.
- Skipped YMM cache scans, push/pop work, reset work, and purge checks when the block does not use YMM.
- Avoided arch metadata scans for blocks with no x87/MMX/XMM/YMM state.
- Kept dynarec dump/test/trace checks evaluated per instruction, preserving debugging behavior.

Benchmarks

Compared optimized tree against a detached baseline build from `HEAD`.

Environment:

BOX64_DYNAREC=1
BOX64_DYNACACHE=0
BOX64_AVX=0
BOX64_DYNAREC_FASTROUND=0
BOX64_DYNAREC_FASTNAN=0
```

Workloads run through Box64 itself:

fpu, sse_asm, sse_intrinsics, sse4_2, x87, x87cache, mmx, lock
```

Aggregate cold-launch compile-suite results over 12 paired runs:

| Build | Median | Mean |
| Baseline | 1.955278s | 1.909630s |
| Optimized | 1.945228s | 1.877103s |

Result:

Median gain: 0.51%
Mean gain:   1.70%

Per-test medians showed small gains on `fpu`, `sse_intrinsics`, `sse4_2`, and `x87`, with very short tests remaining within timing noise. `benchfloat` was also run under the emulator with AVX disabled and was effectively flat after warmup, as expected for a workload after cache compilation successfully completes.

Validation

- Full ARM64 build passed.
- `ctest`: 34/34 passed.
- Focused non-AVX tests passed.
- AVX sanity tests still passed.
- `BOX64_DYNAREC_DUMP=1` smoke test emitted dynarec dump output and completed.
- `BOX64_DYNAREC_TEST=1 BOX64_DYNAREC_TEST_NODUMP=1` smoke test completed.
- `box64 --test` CLI path still reaches the expected unit-test mode.


ptitSeb:
I guess the interpreter could use the same REX lookup table.

necat101:
> I guess the interpreter could use the same REX lookup table.

i think so too! (sorry for the late response, i just had a family emergency pop up irl)

ptitSeb:
> > I guess the interpreter could use the same REX lookup table.
> 
> i think so too! (sorry for the late response, i just had a family emergency pop up irl)

That's not a problem, github is not supposed to be instant messaging...

Would you add Interpreter support in this PR or do you prefer a sperate one?

necat101:
> > > I guess the interpreter could use the same REX lookup table.
> > 
> > 
> > i think so too! (sorry for the late response, i just had a family emergency pop up irl)
> 
> That's not a problem, github is not supposed to be instant messaging...
> 
> Would you add Interpreter support in this PR or do you prefer a sperate one?



> > > I guess the interpreter could use the same REX lookup table.
> > 
> > 
> > i think so too! (sorry for the late response, i just had a family emergency pop up irl)
> 
> That's not a problem, github is not supposed to be instant messaging...
> 
> Would you add Interpreter support in this PR or do you prefer a sperate one?

i prefer a separate one! just in case we need to revert any one change at a time and not potentially discard both when 1 or the other could be beneficial!

(I am very paranoid about version control lol)

PR #3862, opened 2026-05-13
Share x64 prefix lookup between interpreter and dynarec

necat101:
Ok, a quick finish up on the originally intended optimization (sorry for the wait and separate PR my brother's car needed quick work :3 )

This PR extends the x64 prefix lookup optimization to the interpreter while keeping dynarec behavior unchanged.

Changes:

Added a shared x64_prefix_kind[256] table for x64 prefix classification.
Moved prefix kind definitions to [x64run_private.h](http://localhost:5900/codex-local-browse/home/userland/box64/x64run_private.h).
Updated the interpreter prefix decode loop to use the shared lookup table.
Updated the dynarec native pass to use the same shared table instead of a private copy.
The decode semantics are preserved:

LOCK, [REP/REPNZ](http://localhost:5900/codex-local-browse/home/userland/box64/REP/REPNZ), segment overrides, 66, and 67 prefixes keep the same side effects.
[0x40..0x4f](http://localhost:5900/codex-local-browse/home/userland/box64/0x40..0x4f) are still treated as REX only outside 32-bit mode.
Debug/dynarec test behavior is unchanged.
Validation
cmake --build build -j4 passed.
ctest --test-dir build --output-on-failure: 34/34 passed.
Interpreter-only smoke tests passed with BOX64_DYNAREC=0.
BOX64_DYNAREC_TEST=1 BOX64_DYNAREC_TEST_NODUMP=1 smoke test passed.


necat101:
whoops! i made a mistake! sorry im under a lil bit of stress rn, ill be back with the fix!

ptitSeb:
Yeah, something was strange in this PR. Also, it contained a lot of arm64 dynarec files for something about interpreter.

ptitSeb:
Take you time, there is no stress for this.

necat101:
> Take you time, there is no stress for this.

tysm :3 i am  mainly stressed about family stuffs right now. this bug came from my VNC session bugging out and i lost connection for a little bit while my agent went unsupervised for a few mins while i reset my router. (I use agents to automate checks and to help me brainstorm while i write pseudocode, and I almost NEVER let it write stuff completely on its own!)

PR #3863, opened 2026-05-13
Share x64 prefix lookup between interpreter and dynarec

necat101:
This PR extends the x64 prefix lookup optimization to the interpreter while keeping dynarec behavior unchanged.

Changes:

Added a shared x64_prefix_kind[256] table for x64 prefix classification.
Moved prefix kind definitions to [x64run_private.h](http://localhost:5900/codex-local-browse/home/userland/box64/x64run_private.h).
Updated the interpreter prefix decode loop to use the shared lookup table.
Updated the dynarec native pass to use the same shared table instead of a private copy.
The decode semantics are preserved:

LOCK, REP/REPNZ, segment overrides, 66, and 67 prefixes keep the same side effects.
0x40..0x4f are still treated as REX only outside 32-bit mode.
Debug/dynarec test behavior is unchanged.
Validation
cmake --build build -j4 passed.
ctest --test-dir build --output-on-failure: 34/34 passed.
Interpreter-only smoke tests passed with BOX64_DYNAREC=0.
BOX64_DYNAREC_TEST=1 BOX64_DYNAREC_TEST_NODUMP=1 smoke test passed.

sorry for that mistake fellas! i had a cached binary on my system (a hacked together ubuntu chroot environment on my phone  over VNC to my PC) and i made a mistake with the linkage of the x64_prefix_kind table!


ksco:
The commit history is a mess, please rebase ;)

necat101:
sorry for the late response, its been a busy work week so far.. ill be off tomorrow and can rebase :3

ksco:
Closing for long-term inactivity. Feel free to open a new one.

PR #3873, opened 2026-05-20
[INTERP] Fixed 66 prefix decoding

ksco:
rex.w cancels "66", but not for 66 0f type of prefix. This was addressed in DynaRec, but not in the interpreter, not sure why.


ptitSeb:
In that case, there is a lot of code that can be removed in `run66.c` (the code that handles rex.w cases)

PR #3874, opened 2026-05-20
[LA64_DYNAREC] Optimized AES opcodes using vpaes algorithm

ksco:
Below is a simple benchmark using a password-protected 7z self-extracting Windows program.

Before
0dfb2d005a2552d4a8e7412369cf4904

After
31829c7f4cf264c32de948025910714c


ptitSeb:
Wow, fancy! Also, impressive speedup!

PR #3876, opened 2026-05-20
[DYNAREC] Save temporary registers on the stack before calling PrintT…

zengdage:
…race

The PrintTrace function may modify temporary registers, so we need to push them onto the stack before execution and restore them upon return.

For example, in the RV64 implementation, register `t3` stores the comparison result. As its value may be overwritten by PrintTrace, the subsequent `jz` instruction will use invalid data directly.

```
[BOX64] 0x3f0000239b: 48 85 C0 test rax, rax
[BOX64] 0x3ff7af34d4: 53 emitted opcodes, inst=2, barrier=0 state=3/1(0), set=3F/80, use=0, need=0/80, fuse=1/0, sm=0(0/0), sew@entry=7, sew@exit=7, pred=1
        03f00e13        ADDI            t3, zero, 0x3f(63)
        45ccaa23        SW              t3, emu_s9, 0x454(1108)
        01087e33        AND             t3, rax_a6, rax_a6
        47ccb423        SD              t3, emu_s9, 0x468(1128)
[BOX64] New Instruction x64:0x3f0000239e, native:0x3ff7af35a8
[BOX64] TRACE ----
        01f80b37        LUI             rip_s6, 0x1f80000(33030144)
        001b0b1b        ADDIW           rip_s6, rip_s6, 0x1(1)
        00db1b13        SLLI            rip_s6, rip_s6, 0xd(13)
        39eb0b13        ADDI            rip_s6, rip_s6, 0x39e(926)
        000b0313        MV              t1, rip_s6
        018cbc23        SD              rbx_s8, emu_s9, 0x18(24)
        029cb023        SD              rsp_s1, emu_s9, 0x20(32)
        028cb423        SD              rbp_s0, emu_s9, 0x28(40)
        05acb823        SD              r10_s10, emu_s9, 0x50(80)
        05bcbc23        SD              r11_s11, emu_s9, 0x58(88)
        072cb023        SD              r12_s2, emu_s9, 0x60(96)
        073cb423        SD              r13_s3, emu_s9, 0x68(104)
        074cb823        SD              r14_s4, emu_s9, 0x70(112)
        075cbc23        SD              r15_s5, emu_s9, 0x78(120)
        0010039b        ADDIW           t2, zero, 0x1(1)
        ffffffb7        LUI             t6, 0xfffff000(-4096)
        7dff8f9b        ADDIW           t6, t6, 0x7df(2015)
        01fbffb3        AND             t6, flags_s7, t6
[...]

PR #3880, opened 2026-05-21
[DYNAREC] Modify natflags register save logic to now save all scratch registers

zengdage:
Fix scratch register corruption caused by non-consecutive flags producer and consumer when BOX64_DYNAREC_TRACE is enabled, which introduced by #3876.

For example, the `test rax, rax` flags producer stores its flag calculation operands in scratch register `t3`. The next `mov r14, rax` instruction does not use scratch registers, but its associated trace code can still overwrite t3's value. This means we need to reference the flags consumer that is `jz 0x0000003F0000ABC0` to identify which registers require saving.

```
[BOX64] 0x3f00009f68: 48 85 C0 test rax, rax
[BOX64] 0x3ff7afaef0: 53 emitted opcodes, inst=14, barrier=0 state=3/1(0), set=3F/80, use=0, need=0/80, fuse=1/0, sm=0(0/0), sew@entry=7, sew@exit=7, pred=13
	03f00e13	ADDI            t3, zero, 0x3f(63)
	45ccaa23	SW              t3, emu_s9, 0x454(1108)
	01087e33	AND             t3, rax_a6, rax_a6
	47ccb423	SD              t3, emu_s9, 0x468(1128)
[BOX64] New Instruction x64:0x3f00009f6b, native:0x3ff7afafc4
[BOX64] TRACE ----
[BOX64]  n1:0 n2:0 ----
	01f80b37	LUI             rip_s6, 0x1f80000(33030144)
	005b0b1b	ADDIW           rip_s6, rip_s6, 0x5(5)
	00db1b13	SLLI            rip_s6, rip_s6, 0xd(13)
	f6bb0b13	ADDI            rip_s6, rip_s6, 0xffffff6b(-149)
...............................................................
...............................................................
...............................................................
	048cb783	LD              r9_a5, emu_s9, 0x48(72)
	000cb803	LD              rax_a6, emu_s9, 0x0(0)
	088cbb03	LD              rip_s6, emu_s9, 0x88(136)
	080cbb83	LD              flags_s7, emu_s9, 0x80(128)
	fdfbfb93	ANDI            flags_s7, flags_s7, 0xffffffdf(-33)
	006bdf93	SRLI            t6, flags_s7, 0x6(6)
	020fff93	ANDI            t6, t6, 0x20(32)
	01fbebb3	OR              flags_s7, flags_s7, t6
[BOX64] ----------
[BOX64] 0x3f00009f6b: 49 89 C6 mov r14, rax
[BOX64] 0x3ff7afafc4: 54 emitted opcodes, inst=15, barrier=0 state=0/3(0), set=0/0, use=0, need=80/80, fuse=0/1, sm=0(0/0), sew@entry=7, sew@exit=7, pred=14
	00080a13	MV              r14_s4, rax_a6
[BOX64] New Instruction x64:0x3f00009f6e, native:0x3ff7afb09c
[BOX64] TRACE ----
[BOX64]  n1:28 n2:0 ----
	ff010113	ADDI            sp, sp, 0xfffffff0(-16)
	01c13023	SD              t3, sp, 0x0(0)
	01f80b37	LUI             rip_s6, 0x1f80000(33030144)
	01fbebb3	OR              flags_s7, flags_s7, t6
...............................................................
...............................................................
...............................................................
	00013e03	LD              t3, sp, 0x0(0)
	01010113	ADDI            sp, sp, 0x10(16)
[BOX64] ----------
[BOX64] 0x3f00009f6e: 0F 84 4C 0C 00 00 jz 0x0000003F0000ABC0
[BOX64] 0x3ff7afb09c: 67 emitted opcodes, inst=16, barrier=2 state=0/3(0), set=0/0, use=0, need=80/80, fuse=1/0, sm=0(0/0), sew@entry=7, sew@exit=7, pred=15, jmp=out
	020e1463	BNE             t3, zero, 0x28(40) # +10i(0x3ff7afb1a8)
	00000013	NOP

```

### DIFF
--- a/src/dynarec/dynarec_native_pass.c
+++ b/src/dynarec/dynarec_native_pass.c
@@ -257,11 +257,15 @@ uintptr_t native_pass(dynarec_native_t* dyn, uintptr_t addr, int alternate, int 
         if((trace_end == 0)
             || ((ip >= trace_start) && (ip < trace_end)))  {
                 MESSAGE(LOG_DUMP, "TRACE ----\n");
-                if (BOX64ENV(dynarec_nativeflags)) SAVE_ACTIVE_SCRATCH_REGISTERS;
+                #if defined (SPILL_NF_REGISTERS)
+                if (BOX64ENV(dynarec_nativeflags)) SPILL_NF_REGISTERS;
+                #endif
                 fpu_reflectcache(dyn, ninst, x1, x2, x3);
                 GO_TRACE(PrintTrace, 1, x5);
                 fpu_unreflectcache(dyn, ninst, x1, x2, x3);
-                if (BOX64ENV(dynarec_nativeflags)) LOAD_ACTIVE_SCRATCH_REGISTERS;
+                #if defined (RESTORE_NF_REGISTERS)
+                if (BOX64ENV(dynarec_nativeflags)) RESTORE_NF_REGISTERS;
+                #endif
                 MESSAGE(LOG_DUMP, "----------\n");
             }
         }
--- a/src/dynarec/arm64/dynarec_arm64_private.h
+++ b/src/dynarec/arm64/dynarec_arm64_private.h
@@ -225,10 +225,6 @@ int Table64(dynarec_arm_t *dyn, uint64_t val, int pass);  // add a value to tabl
 
 void CreateJmpNext(void* addr, void* next);
 
-// TODO: Save and restore the temp register.
-#define SAVE_ACTIVE_SCRATCH_REGISTERS do{} while(0);
-#define LOAD_ACTIVE_SCRATCH_REGISTERS do{} while(0);
-
 #define GO_TRACE(A, B, s0)      \
     GETIP(addr);                \
     MOVx_REG(x1, xRIP);         \
--- a/src/dynarec/la64/dynarec_la64_private.h
+++ b/src/dynarec/la64/dynarec_la64_private.h
@@ -200,9 +200,33 @@ int Table64(dynarec_la64_t *dyn, uint64_t val, int pass);  // add a value to tab
 
 void CreateJmpNext(void* addr, void* next);
 
-// TODO: Save and restore the temp register.
-#define SAVE_ACTIVE_SCRATCH_REGISTERS do{} while(0);
-#define LOAD_ACTIVE_SCRATCH_REGISTERS do{} while(0);
+// While we could theoretically traverse forward to find the flags-consuming x86
+// instruction and get the exact scratch registers to save, this is too complicated.
+// So we went with the simpler approach of saving all scratch registers, this won't
+// add noticeable performance overhead in trace mode.
+#define SPILL_NF_REGISTERS         \
+    do {                           \
+       ADDI_D(xSP, xSP, -64);      \
+       ST_D(x1, xSP, 0 * 8);       \
+       ST_D(x2, xSP, 1 * 8);       \
+       ST_D(x3, xSP, 2 * 8);       \
+       ST_D(x4, xSP, 3 * 8);       \
+       ST_D(x5, xSP, 4 * 8);       \
+       ST_D(x6, xSP, 5 * 8);       \
+       ST_D(x7, xSP, 6 * 8);       \
+    } while(0);
+
+#define RESTORE_NF_REGISTERS       \
+    do {                           \
+       LD_D(x1, xSP, 0 * 8);       \
+       LD_D(x2, xSP, 1 * 8);       \
+       LD_D(x3, xSP, 2 * 8);       \
+       LD_D(x4, xSP, 3 * 8);       \
+       LD_D(x5, xSP, 4 * 8);       \
+       LD_D(x6, xSP, 5 * 8);       \
+       LD_D(x7, xSP, 6 * 8);       \
+       ADDI_D(xSP, xSP, 64);       \
+    } while(0);
 
 #define GO_TRACE(A, B, s0)         \
     GETIP(addr, s0);               \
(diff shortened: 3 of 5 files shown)


zengdage:
@ptitSeb @ksco Sorry, the scratch register usage detection method I shared last night has a flaw: for non-consecutive flags producer and consumer, the trace code for all intermediate x86-64 instructions between them must also save scratch registers to avoid data corruption. My fix traverses forward to find the target flags consumer, then fetches the exact scratch registers that need to be saved.
Please help to review again, thanks.

ksco:
This is getting complicated, and I don't like where it's going. On second thought, I think we should simply disable nativeflags if trace is effective. It does not make much sense to have trace and nativeflags both enabled anyway, right?

ptitSeb:
Ok, disabling NativeFlags if Dynarec Trace is enabled looks fair enough to me.
But that's something to keep in mind when debugging, as the issue being debugged could be a NativeFlag one...

ksco:
> But that's something to keep in mind when debugging, as the issue being debugged could be a NativeFlag one...

You will have the wrong flags state if nativeflags is enabled; that's also why we disabled nativeflags in cosim.

zengdage:
Can we just simplify this by saving all scratch registers? Trace is for debugging, performance isn't a priority. Adding ~10 extra instructions won't make any difference.

ksco:
> Can we just simplify this by saving all scratch registers? Trace is for debugging, performance isn't a priority. Adding ~10 extra instructions won't make any difference.

And why do you need to enable nativeflags for trace?

zengdage:
> > Can we just simplify this by saving all scratch registers? Trace is for debugging, performance isn't a priority. Adding ~10 extra instructions won't make any difference.
> 
> And why do you need to enable nativeflags for trace?

Disabling nativeflags is also ok for me.  I just think that directly preserving all scratch registers is a very straightforward fix for this issue, and it will not cause any noticeable impact in trace mode.

ptitSeb:
Well, disable NativeFlags or Save all the registers, I don't mind any of those, as they both have their pros and cons in my point of view.

ksco:
> Well, disable NativeFlags or Save all the registers, I don't mind any of those, as they both have their pros and cons in my point of view.

But what are the cons of disabling nativeflags in trace mode? Trace mode is very similar to cosim, but it is for human manual analysis, so it makes sense to disable it IMO.

ptitSeb:
> > Well, disable NativeFlags or Save all the registers, I don't mind any of those, as they both have their pros and cons in my point of view.
> 
> But what are the cons of disabling nativeflags in trace mode? Trace mode is very similar to cosim, but it is for human manual analysis, so it makes sense to disable it IMO.

The cons are, mostly, disabling native flags prevent you from using cosim & trace to debug native flags issues. you will use a different code path in trace than without trace.

ksco:
Well, you can't use cosim & trace to debug native**flags** issues, nativeflags is meant to skip those FLAGS computations...

ptitSeb:
> Well, you can't use cosim & trace to debug native**flags** issues, nativeflags is meant to skip those FLAGS computations...

The x64 opcode flow is still respected. Trace don't give access to native stuffs, just the x64 part. The flow on jump or conditionnal moved from a trace can be used to debug issue with native flags: you can spot an incoherent branch taken for example.

Again, I'm not against any of the change. If you prefer to disable native flags for trace, I'm fine with that.

ksco:
> Again, I'm not against any of the change. If you prefer to disable native flags for trace, I'm fine with that.

Bah, me neither. @zengdage If you decide to do the scratch registers spill/restore, please refer to this closed PR for suggested macro names and add changes for other backends too: https://github.com/ptitSeb/box64/pull/3879/changes

zengdage:
> > Again, I'm not against any of the change. If you prefer to disable native flags for trace, I'm fine with that.
> 
> Bah, me neither. @zengdage If you decide to do the scratch registers spill/restore, please refer to this closed PR for suggested macro names and add changes for other backends too: https://github.com/ptitSeb/box64/pull/3879/changes

Done. Modify nat flag register spill logic to now save all scratch registers and add implementation for LA64 and PPC64LE too.

ksco (approved):
LGTM, but please update your PR and commit title/desc to reflect the changes.

src/dynarec/dynarec_native_pass.c
```diff
@@ -257,11 +257,15 @@ uintptr_t native_pass(dynarec_native_t* dyn, uintptr_t addr, int alternate, int
         if((trace_end == 0)
             || ((ip >= trace_start) && (ip < trace_end)))  {
                 MESSAGE(LOG_DUMP, "TRACE ----\n");
-                if (BOX64ENV(dynarec_nativeflags)) SAVE_ACTIVE_SCRATCH_REGISTERS;
+                #if defined (RV64) || defined(LA64) || defined(PPC64LE)
```
ptitSeb:
```suggestion
                #if defined (SPILL_NF_REGISTERS)
```

src/dynarec/dynarec_native_pass.c
```diff
@@ -257,11 +257,15 @@ uintptr_t native_pass(dynarec_native_t* dyn, uintptr_t addr, int alternate, int
...
+                if (BOX64ENV(dynarec_nativeflags)) SPILL_NF_REGISTERS;
+                #endif
                 fpu_reflectcache(dyn, ninst, x1, x2, x3);
                 GO_TRACE(PrintTrace, 1, x5);
                 fpu_unreflectcache(dyn, ninst, x1, x2, x3);
-                if (BOX64ENV(dynarec_nativeflags)) LOAD_ACTIVE_SCRATCH_REGISTERS;
+                #if defined (RV64) || defined(LA64) || defined(PPC64LE)
```
ptitSeb:
```suggestion
                #if defined (RESTORE_NF_REGISTERS)
```

ptitSeb:
I suggested some change for simplicity

src/dynarec/dynarec_native_pass.c
zengdage: Done.

src/dynarec/dynarec_native_pass.c
zengdage: Done

PR #3882, opened 2026-05-22
[DYNAREC] Reworked invalidate dynablocks management

ksco:


src/include/box64context.h
```diff
@@ -183,6 +183,10 @@ typedef struct box64context_s {
     #endif
     uintptr_t           max_db_size;    // the biggest (in x86_64 instructions bytes) built dynablock
     rbtree_t*             db_sizes;
+    #define DB_ZOMBIE_SIZE 512
```
ptitSeb: Isn't 512 a bit too much? how about 64 or 128 istead, should be more than enough?

ksco: Sure, I'll change to 64.

PR #3883, opened 2026-05-22
[RV64_DYNAREC] Fixed ANDNPS/ANDNPD/PANDN register reuse conflict in vector path

1362525207:
Fix register reuse conflict in SSE ANDNPS/ANDNPD/PANDN for RV64 vector dynarec.

When Gx and Ex refer to the same register, VXOR_VI clobbers the source value while computing the NOT, causing VAND_VV to operate on incorrect data.

This patch detects the aliasing case and copies the source to a scratch register before applying NOT.

Affected instructions:
- ANDNPS in dynarec_rv64_0f_vector.c
- ANDNPD in dynarec_rv64_660f_vector.c
- PANDN in dynarec_rv64_660f_vector.c

Validation:
- ctest: 34/34 passed on RV64 hardware.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_0f_vector.c
+++ b/src/dynarec/rv64/dynarec_rv64_0f_vector.c
@@ -569,6 +569,10 @@ uintptr_t dynarec64_0F_vector(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip,
             SET_ELEMENT_WIDTH(x1, VECTOR_SEW32, 1);
             GETEX_vector(q0, 0, 0, VECTOR_SEW32);
             GETGX_vector(v0, 1, VECTOR_SEW32);
+            if (v0 == q0) {
+                q0 = fpu_get_scratch(dyn);
+                VMV_V_V(q0, v0);
+            }
             VXOR_VI(v0, v0, 0x1f, VECTOR_UNMASKED);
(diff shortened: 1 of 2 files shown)


ksco (approved):

src/dynarec/rv64/dynarec_rv64_0f_vector.c
```diff
@@ -569,6 +569,10 @@ uintptr_t dynarec64_0F_vector(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip,
             SET_ELEMENT_WIDTH(x1, VECTOR_SEW32, 1);
             GETEX_vector(q0, 0, 0, VECTOR_SEW32);
             GETGX_vector(v0, 1, VECTOR_SEW32);
+            if (v0 == q0) {
```
ptitSeb, line 572: I don't understand what you are doing here. q0 was EX before. Now it(s a scratch register ... and EX is never used and just lost?

ksco: it copied the register to prevent clobbering

ptitSeb: Mmmm, ok, fine, it's only for `q0 == v0`, so q0 content can be lost.

PR #3884, opened 2026-05-23
[RV64_DYNAREC] Fixed PCMPISTRM opcode scratch register conflict

ksco:


src/dynarec/rv64/dynarec_rv64_660f38.c
```diff
@@ -689,40 +689,6 @@ uintptr_t dynarec64_660F38(dynarec_rv64_t* dyn, uintptr_t addr, uint8_t opcode,
                         SW(x3, gback, gdoffset + i * 4);
                     }
                     break;
-                case 0x61:
```
ptitSeb, line 692: You removed PCMPESTRI ?

ksco: It should be in the 0x3A and already is, this was a mistake.

ptitSeb: Ah ok, fine.

PR #3890, opened 2026-05-24
[DYNAREC] Optimized bigblock add_next performance with hash table

ksco:
I noticed that when I run `wine explorer` with Wine 11.8 without DynaCache on a Snapdragon 8cx Gen3, there is a hotspot on `add_next`:

```diff
-   22.63%    22.62%  explorer.exe     box64  [.] add_next
     0.01%     0.01%  wine64           box64  [.] add_next
     0.00%     0.00%  wineserver       box64  [.] add_next
```

After applying this optimization, the hotspot disappeared:

```diff
+    0.94%     0.57%  explorer.exe     box64  [.] add_next
     0.02%     0.01%  wine64           box64  [.] add_next
     0.00%     0.00%  conhost.exe      box64  [.] add_next
```

The measured startup time of `wine explorer` dropped from 21s to 16s, which is now quite close to the 14s achieved with DynaCache enabled.

### DIFF
--- a/src/dynarec/dynarec_native.c
+++ b/src/dynarec/dynarec_native.c
@@ -68,11 +68,10 @@ void printf_x64_instruction(dynarec_native_t* dyn, zydis_dec_t* dec, instruction
 void add_next(dynarec_native_t *dyn, uintptr_t addr) {
     if (!BOX64DRENV(dynarec_bigblock))
         return;
-    khint_t k = kh_get(nextset, khnextset, addr);
-    if(k != kh_end(khnextset))
-        return;
     int ret;
     kh_put(nextset, khnextset, addr, &ret);
+    if(!ret)
+        return;
     // add slots
     if(dyn->next_sz == dyn->next_cap) {
         printf_log(LOG_NONE, "Warning, overallocating next\n");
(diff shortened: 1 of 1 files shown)


src/dynarec/dynarec_native.c
```diff
@@ -65,16 +68,11 @@ void printf_x64_instruction(dynarec_native_t* dyn, zydis_dec_t* dec, instruction
...
-    for(int i=0; i<dyn->next_sz; ++i)
-        if(!dyn->next[i]) {
-            dyn->next[i] = addr;
-            return;
-        }
+    khint_t k = kh_get(nextset, khnextset, addr);
+    if(k != kh_end(khnextset))
+        return;
+    int ret;
+    kh_put(nextset, khnextset, addr, &ret);
```
ptitSeb, line 72: There is a way to just use kh_put here, instead of kh_get+kh_put, and check ret to see if kh_put actualy add a key or not (ret==0 eans key aready exist so can exit here)

ptitSeb:
Interesting, but I wonder if an rbtree woud be better, as there is a notion of order in the added addresses?

ksco:
I'll give rbtree a try.

ksco:
nah I don't think an rbtree is useful here, a sorted array is more suitable. but anyway, `get_closest_next()` was not the hotspot here, so that doesn't matter.

src/dynarec/dynarec_native.c
ksco: done

PR #3892, opened 2026-05-25
[SIGNAL] Added partial deferred signal support

ksco:
The cold startup time of `wine explorer` on my 3B6000 dropped from 21s to 13s.

### DIFF
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -16,6 +16,7 @@
 #include "callback.h"
 #include "x64trace.h"
 #include "custommem.h"
+#include "signals.h"
 #include "khash.h"
 #include "threads.h"
 #include "rbtree.h"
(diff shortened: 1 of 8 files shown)


ptitSeb:
There is an issue with WOW64 build:
```
In file included from /home/runner/work/box64/box64/src/dynarec/arm64/dynarec_arm64_helper.c:11:
/home/runner/work/box64/box64/wine/wow64/../../src/emu/x64emu_private.h:143:5: error: unknown type name 'siginfo_t'; did you mean 'sysinfo_t'?
  143 |     siginfo_t   deferred_siginfo[MAX_SIGNAL+1];
      |     ^~~~~~~~~
      |     sysinfo_t
/home/runner/work/box64/box64/wine/wow64/../../src/include/sysinfo.h:21:3: note: 'sysinfo_t' declared here
   21 | } sysinfo_t;
      |   ^
1 error generated.
```

ptitSeb:
To go further with this change, at some point we'll have to check if can avoid the `get_emu_thread()` function, which might be a bit taxing. Having an `x64emu_t*` along all function  call is not always possible (well, possible yes, but usefull or practical, not always), and the cost of the function call needs to be evaluated, but that's something to keep in mind.

PR #3895, opened 2026-05-26
[SYSCALL] Ignore PR_SET_SYSCALL_USER_DISPATCH

ksco:
Used by Wine since 11.5: https://www.winehq.org/news/2026032001, but this seems not supported by LoongArch. I think it's supported by Arm, needs more testing.

### DIFF
--- a/src/emu/x64syscall.c
+++ b/src/emu/x64syscall.c
@@ -19,6 +19,7 @@
 #include <sys/wait.h>
 #include <sys/utsname.h>
 #include <sys/resource.h>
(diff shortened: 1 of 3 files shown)


ptitSeb:
LGTM.

We might need to modify how syscall are traped in wine, and adapt the resulting SIGSYS `siginfo_t` to closer match https://man7.org/linux/man-pages/man2/PR_SET_SYSCALL_USER_DISPATCH.2const.html when user dispatch is enabled (even if, after checking wine sources, that doesn't seems important for now)

PR #3906, opened 2026-05-29
[SYSINFO] Fixed readFreq for LoongArch64

ksco:
We cannot use the "real" value from sysinfo because RDTSC opcode uses a "fake" rdtime.d.

### DIFF
--- a/src/os/freq_linux.c
+++ b/src/os/freq_linux.c
@@ -69,20 +69,51 @@ static inline uint64_t readFreq()
     return val;
 }
 #elif defined(LA64)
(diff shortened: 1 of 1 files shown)


ptitSeb:
What do you mean by "fake"? To me, it looks like the frequency of the internal timer was not read correctly, and can be different from the main cpu frequency after all.

ksco:
Because it's not the freq, it's just a timer.

ksco:
We can also turn the timer to real freq by multiplying something, but I'm not really motivated to do that.

PR #3907, opened 2026-05-29
[ENV] Enabled AVX by default on LoongArch

ksco:


ptitSeb:
doc need to be changed also, it's mentionned that AVX=2 by default only for Arm64

PR #3908, opened 2026-05-29
[ENV] Disable dynacache if dynarec is disabled

ksco:


src/tools/env.c
```diff
@@ -282,6 +282,8 @@ static void applyCustomRules()
 #ifndef _WIN32
     if (box64env.addlibs) AddNewLibs(box64env.addlibs);
 #endif
+
+    box64env.dynacache = box64env.dynarec && box64env.dynacache;
```
ptitSeb, line 286: but dynacache can have a value of "2", it's not just On/Off

PR #3914, opened 2026-05-29
fix clang warnings: check linker flags

peppergrayxyz:
building for neoverse-n1 (-DADLINK) with clang warns about unsupported linker flags:
```
clang: warning: optimization flag '-fuse-linker-plugin' is not supported [-Wignored-optimization-argument]
clang: warning: argument unused during compilation: '-fuse-ld=gold' [-Wunused-command-line-argument]
```

add check for ld.gold and only use it if available


ksco:
Please don't do this. Merge it into 1 PR.

PR #3915, opened 2026-05-29
fix clang warnings

peppergrayxyz:
building using clang causes some warnings:

**check linker flags**
neoverse-n1 (-DADLINK) warns about unsupported linker flags:
```
clang: warning: optimization flag '-fuse-linker-plugin' is not supported [-Wignored-optimization-argument]
clang: warning: argument unused during compilation: '-fuse-ld=gold' [-Wunused-command-line-argument]
```
add check for ld.gold and only use it if available

**use portable inttypes**
warning about format specifier mismatch:
```
src/os/sysinfo.c:173:26: warning: format specifies type 'unsigned long long' but the argument has type 'uint64_t' (aka 'unsigned long') [-Wformat]
  173 |     sprintf(str, "%llu", info->frequency);
      |                   ~~~~   ^~~~~~~~~~~~~~~
      |                   %lu
src/os/sysinfo.c:193:129: warning: format specifies type 'int' but the argument has type 'uint64_t' (aka 'unsigned long') [-Wformat]
  193 |             snprintf(branding, sizeof(branding), BOX64_BUILD_INFO_STRING_SHORT " on %.*s @%04d MHz", 28, box64_sysinfo.cpuname, box64_sysinfo.frequency / 1000000);
      |                                                                                           ~~~~                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                                                           %04lu
```

use inttypes format macros instead of format specifiers directly

**print missing argument**
warning about an unused parameter:
```
src/emu/x64int3.c:208:117: warning: data argument not used by format string [-Wformat-extra-args]
  208 |                     snprintf(buff, 256, "%04d|%p: Calling %s(\"%s\")", tid, *(void**)(R_RSP), s, (tmp)?tmp:"(nil)", (tmp2)?tmp2:"nil");
      |                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                               ^
```

**use array indexing**
warning about the way the signature is formated:
```
src/wrapped/wrappedvulkan_private.h:80:1: warning: adding 'int' to a string does not append to the string [-Wstring-plus-int]
src/libtools/vulkanoverlay.c:133:94: note: expanded from macro 'GO_vFpff'
  133 | #define GO_vFpff(A, B, N)      static void my_##A##_##N(void* a, float b, float c) {F3(A, #B +2, N);}
      |                                                                                     ~~~~~~~~~^~~~~~
src/wrapped/wrappedvulkan_private.h:80:1: note: use array indexing to silence this warning
```

follow compiler suggestion and replace `#B +2` with `&(#B)[2]`

**use proper type in function signature**
warning about type mismatch:

```
src/wrapped/wrappedgmp.c:158:38: warning: passing 'const char *' to parameter of type 'void *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
  158 |     return my->__gmp_vasprintf(strp, fmt, VARARGS);
      |                                      ^~~
  165 |     return my->__gmp_vfprintf(stream, fmt, VARARGS);
      |                                       ^~~
  176 |     return my->__gmp_vasprintf(strp, fmt, VARARGS);
      |                                      ^~~
  187 |     return my->__gmp_vfprintf(stream, fmt, VARARGS);
      |                                       ^~~
```

modifiy definition to use the actual type instead of void.

**remove extraneous parentheses**
warning about extraneous parentheses:
```
src/wrapped32/wrappeddbus.c:541:18: warning: equality comparison with extraneous parentheses [-Wparentheses-equality]
  541 |         if((type == (int)'s')) {
      |             ~~~~~^~~~~~~~~~~
src/wrapped32/wrappeddbus.c:541:18: note: remove extraneous parentheses around the comparison to silence this warning
```


ksco:
Thank you for merging this!

CMakeLists.txt
```diff
@@ -158,6 +158,10 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
 endif()
 
+include(CheckLinkerFlag)
+enable_language(CXX)
```
ptitSeb: No, Why would CXX be needed? There is no C++ in the project, I don't want a C++ compiler to sudenly be needed..

ptitSeb (changes requested):
Remove the needs of a C++ compiler.

CMakeLists.txt
peppergrayxyz: sorry, force of habit! I changed it to use the c compiler!

peppergrayxyz:
I changed `CXX` to `C` and I added these as well:

**excess elements in initializer**
building using clang warns about excess elements in initializer

```
src/wrapped/wrappercallback.h:16:53: warning: excess elements in struct initializer [-Wexcess-initializers]
   16 | static TYPENAME(LIBNAME) TYPENAME2(my_, LIBNAME) = {0};
      |                                                     ^
```

**remove excessive argument**
building using clans warns about too many arguments:

```
src/wrapped32/wrappedopenal.c:104:41: warning: too many arguments in call to 'fillALProcWrapper32'
  104 |         fillALProcWrapper32(emu->context);
      |         ~~~~~~~~~~~~~~~~~~~             ^
src/wrapped32/wrappedopenal.c:104:28: warning: passing arguments to 'fillALProcWrapper32' without a prototype is deprecated in all versions of C and is not supported in C23 [-Wdeprecated-non-prototype]
  104 |         fillALProcWrapper32(emu->context);
      |                            ^
```

Remove the argument, because the function doesn't take any

src/wrapped/wrappercallback.h
```diff
@@ -13,7 +13,7 @@ typedef struct TYPENAME2(LIBNAME, _my_s) {
 } TYPENAME(LIBNAME);
 
 static library_t* my_lib = NULL;
-static TYPENAME(LIBNAME) TYPENAME2(my_, LIBNAME) = {0};
+static TYPENAME(LIBNAME) TYPENAME2(my_, LIBNAME) = {};
```
ptitSeb: Isn't that a C23 feature? There are people that use old compiler, pretty sure that will not work there.

peppergrayxyz: Oh, true... I just checked with clang and gcc but hadn't older compilers in mind. Is there actually any reason for the initializer? Since this is static anyway, it could be removed altogether

ptitSeb: If I remember correctly, this warning appear only on a couple of file, and I think a solution is to remove the include file, because the structure is empty, the whole include file is not usefull.

peppergrayxyz: I updated the PR to only removing the redundant initializer for now (don't want to open another potential can of worms and I don't understand well enough what is thing does)

Issue #3916, opened 2026-05-29
several clang warnings are bugs


peppergrayxyz:
There are clang warnings, that point to bugs:

### src/tools/env.c:

**Warning:**
```
src/tools/env.c:786:13: warning: using the result of an assignment as a condition without parentheses [-Wparentheses]
  786 |         if(p=lowercase_filename+strlen(lowercase_filename)-strlen(" (deleted)"))
      |            ~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/tools/env.c:786:13: note: place parentheses around the assignment to silence this warning
  786 |         if(p=lowercase_filename+strlen(lowercase_filename)-strlen(" (deleted)"))
      |             ^
      |            (                                                                   )
src/tools/env.c:786:13: note: use '==' to turn this assignment into an equality comparison
  786 |         if(p=lowercase_filename+strlen(lowercase_filename)-strlen(" (deleted)"))
      |             ^
      |             ==
```
**Code:**

L785-789 in function RecordEnvMappings
```C
  // memfd, first remove the (deleted) at the end
  char* p = strstr(lowercase_filename, " (deleted)");
  if(p=lowercase_filename+strlen(lowercase_filename)-strlen(" (deleted)"))
      *p = 0;
  // add the "/fd" at the end to differenciate between memfd
  char* new_name = box_calloc(1, strlen(lowercase_filename)+100);
```

**Expected behaviour:**

The function removes the suffix if present

**Actual behaviour:**

The function always removes the last 10 bytes.
Also, the +100 bytes buffer on line 788 seems excessive for `%d`. INT_MAX is 19 chars on 64bit.

### wrappedlibc.c:
 
**Warning:**
```
src/wrapped/wrappedlibc.c:2046:46: warning: format specifies type 'char *' but the argument has type 'char (*)[4096]' [-Wformat]
 2046 |         if(sscanf(path, "/proc/%d/%s", &pid, &p)==2)
      |                                   ~~         ^~
src/wrapped/wrappedlibc.c:2047:16: warning: address of array 'p' will always evaluate to 'true' [-Wpointer-bool-conversion]
 2047 |             if(p && !strcmp(p, w))
      |                ^ ~~
 ```

**Code:**

L241-2051
```C
if(sscanf(path, "/proc/%d/%s", &pid, &p)==2)
    if(p && !strcmp(p, w))
[...]

PR #3917, opened 2026-05-29
fix clang warnings that are bugs

peppergrayxyz:
I split this PR from the previous, because these are not just fixes to warnings, but to behaviour, because the warning is caused by bugs (see also https://github.com/ptitSeb/box64/issues/3916)

- **remove suffix only if present**
the suffix is always removed, not only when present due to an assignment
that should be a comparison
- **fix isProcAny check**
the pid should only be returned, when the path matches w, but is always
returned. Pass p as is (is already a pointer)
- **add missing parameter**
pid is missing in path construction

### DIFF
--- a/src/wrapped/wrappedlibc.c
+++ b/src/wrapped/wrappedlibc.c
@@ -2043,8 +2043,8 @@ static int isProcAny(const char *path, const char* w)
     if(strncmp(path, "/proc/", 6)==0) {
         int pid;
         char p[4096] ={0};
-        if(sscanf(path, "/proc/%d/%s", &pid, &p)==2)
-            if(p && !strcmp(p, w))
+        if(sscanf(path, "/proc/%d/%s", &pid, p)==2)
+            if(strcmp(p, w) == 0)
                 return pid;
     }
     return -1;
--- a/src/wrapped/wrappedlibc.c
+++ b/src/wrapped/wrappedlibc.c
@@ -2125,7 +2125,7 @@ EXPORT ssize_t my_readlink(x64emu_t* emu, void* path, void* buf, size_t sz)
                 if(filename[0]=='.') {
                     // relative path, need to grap cwd and cannonicalise the path
                     char cwd_name[strlen(path)+4];
-                    sprintf(cwd_name, "/proc/%d/cwd");
+                    sprintf(cwd_name, "/proc/%d/cwd", pid);
                     char cwd[MAX_PATH] = {0};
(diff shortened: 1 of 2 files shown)


src/wrapped/wrappedlibc.c
```diff
@@ -2043,8 +2043,8 @@ static int isProcAny(const char *path, const char* w)
     if(strncmp(path, "/proc/", 6)==0) {
         int pid;
         char p[4096] ={0};
-        if(sscanf(path, "/proc/%d/%s", &pid, &p)==2)
-            if(p && !strcmp(p, w))
+        if(sscanf(path, "/proc/%d/%s", &pid, p)==2)
+            if(strcmp(p, w) == 0)
```
ptitSeb, line 2047: The change from `!strcmp(...)` tp `strcmp(...) == 0` smells a lot like A.I. A human would have just removed the leading `p && `...

(the comments of the PR/Tickets are also very verbose)

Just a reminder, we ask everyone to declare if they use A.I for a PR, and not to let A.I. do all the comunication (creating tickets), as we are human handling the tickets.

Also also, it's always nice to see wich agent are use to do stuffs.

peppergrayxyz: The form `strcmp(...) == 0` is used on the lines just below (L2059-64), so I thought I'll keep the style consistent:

```C
  if(strcmp((const char*)path, tmp)==0)
      return 1;
  // check if self PID ....
  pid_t pid = getpid();
  sprintf(tmp, "/proc/%d/%s", pid, w);
  if(strcmp((const char*)path, tmp)==0)
```

I used to get feedback on my PRs that I'm too minimalistic, so I started being more verbose. Also I'm on arm64 Gentoo/musl/libcxx which is usually not supported by projects, so I try to make it easy for maintainers in the hopes to get them more likely upstreamed.

Sorry for not acting human enough 😅

ptitSeb: Ah well, sorry, I ted to be more and more paranoiac...

Oh, musl, interesting...

peppergrayxyz: No worries. Crazy times...

I'm not familiar with the source base, so I'm curious if I can get musl to work or hit a roadblock. But so far it looks good and Android doesn't use glibc either, so it shouldn't be I'm cautiously optimistic 🤞

ptitSeb: Well, there is no specific musl support, but having a build that does run musl x64 binry should be possible without too much trouble... Running glibc x64 binary is another story tho....

PR #3920, opened 2026-06-03
[SYSCALL] Fix mmap for io_uring

zengdage:
The pr is try to fix the issue which introduced by mmap for io_uring fd. Now mmap doesn't support non-NULL address for io_uring fd in linux-6.18. For example, the box64 will run the following code fail.

```
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sys/syscall.h>
#include <sys/mman.h>
#include <fcntl.h>
#include <string.h>
#include <stdint.h>
#include <errno.h>
#include <linux/io_uring.h>

int main() {
    struct io_uring_params params = {0};
    int ring_fd;

    ring_fd = syscall(__NR_io_uring_setup, 32, &params);
    if (ring_fd < 0) {
        perror("io_uring_setup failed");
        return 1;
    }

    size_t total_size = params.sq_off.array + params.sq_entries * sizeof(uint32_t) +
                        params.cq_off.cqes + params.cq_entries * sizeof(struct io_uring_cqe);

    total_size = (total_size + 4095) & ~4095;

    void *ring_ptr = mmap(NULL, total_size,
                          PROT_READ|PROT_WRITE, MAP_SHARED,
                          ring_fd, 0);
    if (ring_ptr == MAP_FAILED) {
        perror("ring mmap failed");
        close(ring_fd);
        return 1;
    }

    munmap(ring_ptr, total_size);
    close(ring_fd);
    return 0;
}

```

### DIFF
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -3333,6 +3333,12 @@ EXPORT void* box_mmap(void *addr, size_t length, int prot, int flags, int fd, ss
     }
     #endif
     void* ret = InternalMmap(addr, length, prot, new_flags, fd, offset);
+    // io_uring doesn't support non-NULL address.
+    // The optimal approach is to detect whether an fd is an io_uring instance,
+    // but this is overly complex. So we simply retry the mmap call with the
+    // original address here.
+    if (ret == MAP_FAILED && old_addr == NULL && fd >= 0)
+        ret = InternalMmap(old_addr, length, prot, new_flags, fd, offset);
 #if !defined(NOALIGN)
     if((ret!=MAP_FAILED) && (flags&MAP_32BIT) &&
       (((uintptr_t)ret>0xffffffffLL) || ((box64_wine) && ((uintptr_t)ret&0xffff) && (ret!=addr)))) {


ptitSeb:
I understand the fix, but it's too generic in it's current form and will break numerous other things that rely on mmap to fail in certain cases (like allocating too large amount of memory for a 32bits software).

My suggestion would be to move up the test to line 3336 instead, and to add a check if fd>=0 (or evenjust >0 I guess would be acceptable) to reduce the number of case it would fail.

Anothor solution would be to maintain a list of openned uring, but that's certainly more complex to do.

zengdage:
> I understand the fix, but it's too generic in it's current form and will break numerous other things that rely on mmap to fail in certain cases (like allocating too large amount of memory for a 32bits software).
> 
> My suggestion would be to move up the test to line 3336 instead, and to add a check if fd>=0 (or evenjust >0 I guess would be acceptable) to reduce the number of case it would fail.
> 
> Anothor solution would be to maintain a list of openned uring, but that's certainly more complex to do.

Done. Please help review again. 
Checking if an fd is an io_uring fd is somewhat complicated, so I think we can wait and see for now. Of course, if it is strongly required to do so, I can also try to implement it.
It is just that based on my understanding, there is no difference between the first mmap call and the second mmap call except for the address. Excluding mmap failures caused solely by address factors, if the first mmap execution fails, the second mmap execution will also fail with overwhelming probability.

ptitSeb:
Ok, thanks. I guess this is fine for now.

PR #3940, opened 2026-06-11
[WRAPPER] Add symbols for Anaconda-Navigator

yzewei:


src/wrapped/wrappedlibresolv_private.h
```diff
@@ -87,18 +87,18 @@ GO(__res_nameinquery, iFpiipp)
 #ifdef STATICBUILD
 //GO(__res_nmkquery, 
 #else
-GO(__res_nmkquery, iFpipiipippi)
+GO2(__res_nmkquery, iFpipiipippi, res_nmkquery)
```
ptitSeb, line 90: Pretty sure that will break stuffs for system using old glibc, prior to 2.34 or something like that.

yzewei: Use GOM instead of GO2 for processing.

PR #3945, opened 2026-06-12
Add null check for dynablock in arch_unaligned

Alion-King:
I find a error iwhen testing examples “signal03 signal05” in LTP。
the resolution has existed in arm64 and ppc。


ptitSeb:
LGTM

ksco:
Lets merge it then.

PR #3951, opened 2026-06-15
[SIGNAL] Set no_tf when resuming from a signal for Syringe.exe

ksco:
Wine implements NtSetContextThread by send SIGUSR1 to child and resume it after -- where TF might be set, we need to skip the first SIGTRAP.

### DIFF
--- a/src/libtools/signals.c
+++ b/src/libtools/signals.c
@@ -565,7 +565,9 @@ int my_sigactionhandler_oldcode_64(x64emu_t* emu, int32_t sig, int simple, sigin
 
     if(memcmp(sigcontext, &sigcontext_copy, sizeof(x64_ucontext_t))) {
         #if defined(DYNAREC)
-        if(db) {
+        if(db || emu->jmpbuf)
+            mctx2emu(emu, &sigcontext->uc_mcontext);
+        if(db && !ACCESS_FLAG(F_TF)) {
             // if signal was inside a dynablock, just mirror all the new regs in the right place to simple run native_next
             mctx2emu(emu, &sigcontext->uc_mcontext);
             copyEmu2USignalCTXreg(p, emu, native_next);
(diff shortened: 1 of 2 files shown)


src/libtools/signals.c
```diff
@@ -568,6 +568,7 @@ int my_sigactionhandler_oldcode_64(x64emu_t* emu, int32_t sig, int simple, sigin
         if(db) {
             // if signal was inside a dynablock, just mirror all the new regs in the right place to simple run native_next
             mctx2emu(emu, &sigcontext->uc_mcontext);
+            if(ACCESS_FLAG(F_TF)) emu->flags.no_tf = 1;
```
ptitSeb: This will not work here, as it will resume execution in dynarec directly. 
mctx2emu needs to be done before and TF tested to not resume in dynarec if it's set.
```C
        #if defined(DYNAREC)
        if(db || emu->jmpbuf)
            mctx2emu(emu, &sigcontext->uc_mcontext);
        if(db && !ACCESS_FLAG(F_TF)) {
            // if signal was inside a dynablock, just mirror all the new regs in the right place to simple run native_next
            mctx2emu(emu, &sigcontext->uc_mcontext);
            copyEmu2USignalCTXreg(p, emu, native_next);
            printf_log((sig==10)?LOG_DEBUG:log_minimum, "Context has been changed in Sigactionhanlder, jumping to native_next from DynaBlock at %p, RSP=%p\n", (void*)R_RIP, (void*)R_RSP);
            return 1;
        }
        #endif
        if(emu->jmpbuf) {
            #ifndef DYNAREC
            mctx2emu(emu, &sigcontext->uc_mcontext);
            #endif
            if((skip==1) && (emu->ip.q[0]!=sigcontext->uc_mcontext.gregs[X64_RIP]) && !ACCESS_FLAG(F_TF))
                skip = 3;   // if it jumps elsewhere, it can resume with dynarec...
            printf_log((sig==10)?LOG_DEBUG:log_minimum, "Context has been changed in Sigactionhanlder, doing siglongjmp to resume emu at %p, RSP=%p (resume with %s)\n", (void*)R_RIP, (void*)R_RSP, (skip==3)?"Dynarec":"Interp");
```
something like that instead

ksco: yeah, indeed, need to force the interpreter…

ptitSeb: will you change this PR or do you prefer I merge this one as-is and push another PR to fix those cases?

ksco: I'll address this soon.

ksco: But we do need `if (ACCESS_FLAG(F_TF) && skip == 1) emu->flags.no_tf = 1;`, right?

ptitSeb: yeah sure
my changes were just to make sure Interpreter was used

src/libtools/signals.c
```diff
@@ -574,9 +576,9 @@ int my_sigactionhandler_oldcode_64(x64emu_t* emu, int32_t sig, int simple, sigin
         }
         #endif
         if(emu->jmpbuf) {
-            if((skip==1) && (emu->ip.q[0]!=sigcontext->uc_mcontext.gregs[X64_RIP]))
+            if((skip==1) && (emu->ip.q[0]!=sigcontext->uc_mcontext.gregs[X64_RIP]) && !ACCESS_FLAG(F_TF))
                 skip = 3;   // if it jumps elsewhere, it can resume with dynarec...
-            mctx2emu(emu, &sigcontext->uc_mcontext);
```
ptitSeb, line 579: mctx2emu call for non-Dynarec code seems to be missing

ptitSeb: and it should be moved before the if, as it contain update x64emu_t data

PR #3963, opened 2026-06-19
[CPUID] Fixed a regression introduced in f05bdb1f9

ksco:
This is really dumb, I'm sorry. Fixed FF VII Rebirth. Thanks @phorcys for finding this.


ptitSeb:
Ooof, I see! That might help a few things indeed.

PR #3975, opened 2026-06-22
[LA64_DYNAREC] More work on ROL/ROR opcodes OF emulation

ksco:


src/dynarec/la64/dynarec_la64_00.c
```diff
@@ -2258,11 +2258,16 @@ uintptr_t dynarec64_00(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                     INST_NAME("ROL Eb, Ib");
                     u8 = geted_ib(dyn, addr, ninst, nextop) & 0x1f;
                     if (u8) {
-                        SETFLAGS((u8 > 1) ? X_CF : (X_OF | X_CF), SF_SUBSET, NAT_FLAGS_FUSION); // removed PENDING on purpose
-                    GETEB(x1, 1);
-                    u8 = F8 & 0x1f;
-                    emit_rol8c(dyn, ninst, ed, u8, x4, x5, x6);
-                    EBBACK();
+                        if (MODREG && u8 > 1) {
+                            READFLAGS(X_OF);
```
ptitSeb: I'm unsure why you would want to read OF flag here. Why not just completly ignore OF flags as it's not in the SETFLAGS, and will just be get from previous opcode if needed.
Using the READFLAGS will force computation of OF even if it's not consumed everywhere.

ksco: Yeah, indeed. It's some debug leftovers...

src/dynarec/la64/dynarec_la64_00.c
```diff
@@ -3981,8 +3995,6 @@ uintptr_t dynarec64_00(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                     SKIPTEST(x1);
                     if (!BOX64DRENV(dynarec_safeflags)) {
                         SETFLAGS(X_ALL, SF_SET_NODF, NAT_FLAGS_NOFUSION);
-                    } else if (BOX64ENV(cputype)) {
```
ptitSeb, line 3984: why removing some cputype handling?

ksco: Because it's not supported, and setting cputype=1 to make some partial effects is not a good idea.

ptitSeb: Well, the main effect is the CPUID change, which are supported as it's a C function. Rest is undefined behaviour, which could be added as  a step by step, not necessarily full handling in 1 PR.
But I'm fine if you prefer to remove all handling for now.

ksco: yeah, I'll add full support at once when I find the time

PR #3994, opened 2026-06-29
[WRAPPER][BOX32] Fixed unexpected bridge reuse

ksco:
ftruncate and ftruncate64 point to the same native libc code on 64-bit systems, but on 32-bit x86 they are different functions. We reused bridges by native address, so a bridge made for ftruncate64 could also be used for ftruncate accidently, which breaks the latest legacy Linux Steam.

Added a new bridge key to include the wrapper too, so aliases with different guest APIs get separate bridges.

### DIFF
--- a/src/tools/bridge.c
+++ b/src/tools/bridge.c
@@ -28,6 +28,17 @@
 
 KHASH_MAP_INIT_INT128(bridgemap, uintptr_t)
 
+typedef struct bridgekey_s {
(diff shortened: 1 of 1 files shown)


ptitSeb:
m, There is something strange here: that's not how bridge are supposed to work. I need to check this back, hash and stuffs are not supposed to be needed.

PR #3998, opened 2026-06-30
[TRACE] Support configuring trace address list through BOX64_TRACE

zengdage:
Sometimes only specific individual addresses need to be traced, which do not form a continuous address range. If placed within a single address interval, excessive trace output will be generated. This patch is added to facilitate debugging.

Format: BOX64_TRACE=0x1000:0x2000:0x3000

Just trace the address 0x1000,0x2000 and 0x3000


ptitSeb:
A list of individual addresses instead f a list of range looks strange and not too usefull for me. But I guess list of range can be added later.

docs/USAGE.md
```diff
@@ -764,6 +764,7 @@ Same as BOX64_TRACE but starts the trace immediately.
  * 1: Enable trace output. Trace starts before the initialization of dependencies.
  * symbolname: Enable trace output for `symbolname` only. Trace starts before the initialization of dependencies.
  * 0xXXXXXXX-0xYYYYYYY: Enable trace output for the range of address (inclusive-exclusive). Trace starts before the initialization of dependencies.
+ * 0xXXXXXXX:0xYYYYYYY:0xZZZZZZZ: Enable trace output only for the specified address list. Trace starts before the initialization of dependencies.
```
ptitSeb, line 767: Please, modify `usage.json` instead (and `usage_cn.json`if you can write chinese) then use `docs/gen/gen.py` to regenerate `USAGE.md` (and other docs).

ptitSeb (changes requested):
Do not change USAGE.md directly, it's generated.

zengdage:
> A list of individual addresses instead f a list of range looks strange and not too usefull for me. But I guess list of range can be added later.

Thanks. I will try to support for a list of address range later.

PR #3999, opened 2026-07-01
[CORE] Add MallocHookRun to call startMallocHook before RunElfInit

zengdage:
The `startMallocHook` function is only invoked inside `RunElfInit`, and `RunElfInit` is only called for the main ELF binary within `__libc_start_main`.

In the `RunElfInit` function, `startMallocHook` is currently invoked only after all initialization routines of dynamic libraries have been executed. However, if the malloc hook is defined within a dependent library and that library’s initialization routine invokes the malloc hook interface, activation of the malloc hook will be delayed, leading to unexpected runtime exceptions.

Therefore, I add a call to `MallocHookRun` so that the malloc hook can be available before the execution of library init functions.

Before:
```
__libc_start_main
    |
    ---> RunElfInit
             |
             ----> libA.so Init
             |          |
             |          ---> _Znam
             |                 |
             |                 ---> Native actual_malloc
             ---> startMallocHook
```

After:
```
__libc_start_main
    |
    ---> MallocHookRun
    |        |
    |        ---> startMallocHook
    |
    ---> RunElfInit
             |
             ----> libA.so Init
                       |
                       ---> _Znam
                              |
                              ---> X86-64 real_malloc
```

### DIFF
--- a/src/elfs/elfloader.c
+++ b/src/elfs/elfloader.c
@@ -48,6 +48,12 @@
 #include "../tools/bridge_private.h"
 #include "x64tls.h"
 
+#ifndef STATICBUILD
+void startMallocHook();
+#else
+void startMallocHook() {}
+#endif
+
 void* my__IO_2_1_stderr_ = (void*)1;
 void* my__IO_2_1_stdin_  = (void*)2;
 void* my__IO_2_1_stdout_ = (void*)3;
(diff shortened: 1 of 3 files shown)


zengdage:
I found this issue while analyzing startup crashes of x86-64 Linux WeChat (An IM app developed by Tencent) running on RISC-V Linux.

ptitSeb:
Ok, make sense, thanks.

PR #4000, opened 2026-07-01
[CORE] Add ALLOC guard for malloc_usable_size

zengdage:
Add ALLOC check before invoking real_malloc_usable_size to avoid invalid memory access.

### DIFF
--- a/src/mallochook.c
+++ b/src/mallochook.c
@@ -380,7 +380,7 @@ EXPORT void cfree(void* p)
 
 EXPORT size_t malloc_usable_size(void* p)
 {
-    if(malloc_hack_2 && real_malloc_usable_size) {
+    if(malloc_hack_2 && ALLOC && real_malloc_usable_size) {
         if(getMmapped((uintptr_t)p))
             return RunFunctionFmtNoAlt(real_malloc_usable_size, "p", p);
     }


src/mallochook.c
```diff
@@ -380,7 +380,7 @@ EXPORT void cfree(void* p)
 
 EXPORT size_t malloc_usable_size(void* p)
 {
-    if(malloc_hack_2 && real_malloc_usable_size) {
+    if(malloc_hack_2 && ALLOC && real_malloc_usable_size) {
```
ptitSeb, line 383: I'm not sure just ALLOC makes sense here. Because FREE & ALLOC are separated, even when ALLOC==0, you still get blocks that are custom allocated by the app (some inline stuffs happening), so blocks can exist  that would need the real_malloc_usable _size use. If you really want to add a test, then use `(ALLOC || FREE)`

Did you add this test to fix an issue?

zengdage: > I'm not sure just ALLOC makes sense here. Because FREE & ALLOC are separated, even when ALLOC==0, you still get blocks that are custom allocated by the app (some inline stuffs happening), so blocks can exist that would need the real_malloc_usable _size use. If you really want to add a test, then use `(ALLOC || FREE)`
> 
> Did you add this test to fix an issue?

Sorry, I add the ALLOC check while debugging WeChat, as I encountered a segmentation fault within malloc_usable_size in libc. 

After comparing the implementation of malloc, I naively assumed that since malloc uses the ALLOC flag, malloc_usable_size would also require it. At that time, I failed to find the root cause of the crash and my understanding of this logic was incomplete; I only added this condition because the segmentation fault went away afterward. 

I just attempted to reproduce the original crash, but I was unable to trigger it again. 

I have been focusing on debugging WeChat recently and have prepared several experimental patches that allow WeChat to launch and run stably. Maybe we can shelve this patch for the time being and go over my other patches first.

ptitSeb: Yeah, I don't think this is the correct approach: if a memory address is not identified correctly, the root cause of the misidentification should be addressed instead.

PR #4004, opened 2026-07-01
[CORE] Separate the malloc function for box64 itself and the emulated app

zengdage:
When debugging WeChat and its embedded Electron application, I encountered several memory access errors. After comparative analysis against native x86-64 execution, I found that certain memory blocks are allocated via the app’s self-implemented malloc on x86-64, whereas under Box64 emulation, these allocations fall back to the libc native malloc. This difference triggers runtime crashes in WeChat and its Electron app.

This patch fully isolates Box64’s internal malloc from the emulated program’s custom malloc implementation when the environment variable BOX64_MALLOC_HACK=2 is set, resolving the above crash issue.

### DIFF
--- a/src/mallochook.c
+++ b/src/mallochook.c
@@ -144,7 +144,7 @@ uint32_t getProtection(uintptr_t addr);
 // mmap history
 static int malloc_hack_2 = 0;
 
-#define ALLOC 0
+#define ALLOC 1
 #define FREE 1
 
 char* box_strdup(const char* s) {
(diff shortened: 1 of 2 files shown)


ptitSeb:
Nope. By doing this, you will break every box32 application. I cannot accept this PR

zengdage:
> Nope. By doing this, you will break every box32 application. I cannot accept this PR

My apologies, I haven’t yet figured out why this change breaks box32. Could you help explain this to me?

ksco:
@zengdage Can you check if #4196 fixes said issue?

zengdage:
> @zengdage Can you check if #4196 fixes said issue?

#4196 does not fix this issue. #4004  still needs to be applied for Electron apps such as WeChat Mini Programs to run.

PR #4013, opened 2026-07-03
[INTERP] Added unaligned handling for more opcodes

ksco:


src/emu/x64run66f0.c
```diff
@@ -252,20 +267,22 @@ uintptr_t Run66F0(x64emu_t *emu, rex_t rex, uintptr_t addr)
                     GETEW(0);
                     GETGW;
 #if defined(DYNAREC) && !defined(TEST_INTERPRETER)
-                    if(((uintptr_t)ED)&1) {
```
ptitSeb, line 255: Are you sure this is an improvement? Using a lock instead of the partial atomic move seems a regression to me.

ksco: Old code is not really unaligned. I guess I could fix it but I decided to align with other similar opcode (to simply use a lock).

ptitSeb: Ah, well, old code is buggy, as it was supposed to use `_b` instead if  `_h`...

ksco: You prefer I fix the bug directly?

ptitSeb: Yeah, let me fix the old code first.

PR #4016, opened 2026-07-08
[LA64_DYNAREC] Added upper 32bits zero tracking

ksco:


src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -42,6 +43,14 @@ uintptr_t geted(dynarec_la64_t* dyn, uintptr_t addr, int ninst, uint8_t nextop,
...
         return geted16(dyn, addr, ninst, nextop, ed, hint, scratch, fixaddress, rex, i12);
 
+    if ((nextop & 7) == 4) {
+        uint8_t sib = PK(0);
+        if (((sib & 7) != 5) || ((nextop & 0xC0) != 0)) UP32_READ(TO_NAT((sib & 7) + (rex.b << 3)));
+        if ((((sib >> 3) & 7) + (rex.x << 3)) != 4) UP32_READ(TO_NAT(((sib >> 3) & 7) + (rex.x << 3)));
+    } else if (((nextop & 7) != 5) || ((nextop & 0xC0) != 0)) {
+        UP32_READ(TO_NAT((nextop & 7) + (rex.b << 3)));
+    }
+
```
ptitSeb, line 53: It's fine, but I'm not a big a fan of UP32 having it's own logic (instead of UP32 as-needed inside each cases of geted)

ksco: Okay, I'll change it, but I won't change it in this PR.

ptitSeb:
Signal handling is missing: if a signal occurs, there is no way to actualize pending UP32 opcoration to rebuild correct context for emulated signal handlers.

ksco:
> Signal handling is missing: if a signal occurs, there is no way to actualize pending UP32 opcoration to rebuild correct context for emulated signal handlers.

Is there any similar code exists that I can refer to?

ptitSeb:
> > Signal handling is missing: if a signal occurs, there is no way to actualize pending UP32 opcoration to rebuild correct context for emulated signal handlers.
> 
> Is there any similar code exists that I can refer to?

That's the ARCH thingy basicaly (arm64 can rebuild flags based on native flags or emulated one for example)

ksco:
Okay, I haven't done the ARCH thing for LoongArch...

src/dynarec/la64/dynarec_la64_functions.c
```diff
@@ -930,3 +931,98 @@ void propagateFpuBarrier(dynarec_la64_t* dyn)
...
+    int n = dyn->size;
+    if (n <= 0)
+        return;
+
+    uint16_t* in = (uint16_t*)alloca((size_t)n * sizeof(uint16_t));
+    uint16_t* out = (uint16_t*)alloca((size_t)n * sizeof(uint16_t));
+    int* pred_count = (int*)alloca((size_t)n * sizeof(int));
+    memset(in, 0, (size_t)n * sizeof(uint16_t));
+    memset(out, 0, (size_t)n * sizeof(uint16_t));
+    memset(pred_count, 0, (size_t)n * sizeof(int));
```
ptitSeb: isn't pred already computed at this point? why would you need to compute a new set?

PR #4017, opened 2026-07-09
[DYNAREC] Made callret=2 the default

ksco:


configurator/model.py
```diff
@@ -109,7 +109,6 @@ class ValueState:
         ("BOX64_DYNAREC_STRONGMEM", "1"),
     ),
     "fast": (
-        ("BOX64_DYNAREC_CALLRET", "1"),
```
ptitSeb, line 112: Why removong callret=1 from fast and festest profile? callret=1 should be a bit faster (and a bit more compact), so it should still be a good idea to keep it there?

ksco: Sure

khanh-it:
> 2: Try to optimize CALL/RET, skipping the jump table when possible, adding code to handle return to dirty/modified block. Does not work on WowBox64. [Default]

Then, what if I run a proton/wine game without setting BOX64_DYNAREC_CALLRET?

KostasGangstar-AI:
I'm using WowBox64 in Winlator Ludashi, if callret=2 isn't available in WowBox64 does this mean I have to leave it to 0 like the default was before this PR merged in my emulator's settings

ksco:
No, it stays 0 on WowBox64.

PR #4041, opened 2026-07-13
[RV64_DYNAREC] Added !fastnan path for more opcodes

ksco:


src/dynarec/rv64/dynarec_rv64_f30f_vector.c
```diff
@@ -313,6 +313,7 @@ uintptr_t dynarec64_F30F_vector(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t i
             VFMUL_VV(v0, v0, v1, VECTOR_MASKED);
             break;
         case 0x5A:
+            if (!BOX64ENV(dynarec_fastnan)) return 0;
```
ptitSeb, line 316: Wow, no fastnan, no dynarec... ok.

ksco: No, it will fallback to the DynaRec scalar path because the vector implementation will be as complex as the scalar version anyway…

ptitSeb: Ah ok. Fair enough

PR #4044, opened 2026-07-15
[LA64_DYNAREC] Added basic arch state gathering support and enabled up32

ksco:


ptitSeb:
Some unaligned handling sounds good! That help perfs a lot when it's needed...
But the updateUp32Liveness function is getting more and more complex...

ksco:
> But the updateUp32Liveness function is getting more and more complex...

I hope it will not become a hotspot

src/dynarec/la64/dynarec_la64_arch.c
```diff
@@ -0,0 +1,142 @@
...
+    (void)p;
+    if(!db) return;
+    if(!db->arch_size || !db->arch) return;
+    int ninst = getX64AddressInst(db, x64pc);
+    if(ninst < 0 || ninst >= db->isize) return;
+    arch_arch_t* arch = arch_for_ninst(db, ninst);
+    uint16_t up32 = arch->up32;
+    while(up32) {
+        int r = __builtin_ctz(up32);
+        up32 &= up32 - 1;
```
ptitSeb, line 139: are you sure this will change up32 the correct way? why not using a simpler `up32>>=r+1` ?

ksco: `up32 &= up32 - 1;` is used to remove the least significant 0, same as how the x86 opcode BLSR is implemented. `up32>>=r+1` will break the `r` value in next iteration of the while loop.

ksco: Sorry I meant the least significant 1

ptitSeb: Ok,. Yeah the shift thing will break r, aggreed.

Ok, I more thoroughly checked, it looks good.

src/libtools/sigtools.c
```diff
@@ -469,6 +471,59 @@ int sigbus_specialcases(siginfo_t* info, void * ucntx, void* pc, void* _fpsimd,
...
+            uint64_t *fpregs = NULL;
+            while (info->magic && !fpregs) {
+                if (info->magic == FPU_CTX_MAGIC || info->magic == LSX_CTX_MAGIC || info->magic == LASX_CTX_MAGIC)
+                    fpregs = (uint64_t*)((uintptr_t)info + sizeof(struct sctx_info));
+                else
+                    info = (struct sctx_info*)((uintptr_t)info + info->size);
+            }
+            if (!fpregs)
+                return 0;
+            value = fpregs[val];
```
xiangzhai, line 515: Hi @ksco 

ABI 1.0 (it is able to build cmake 3.19.0):

```
diff --git a/src/libtools/sigtools.c b/src/libtools/sigtools.c
index 4825498c7..8528aea29 100644
--- a/src/libtools/sigtools.c
+++ b/src/libtools/sigtools.c
@@ -502,6 +502,7 @@ int sigbus_specialcases(siginfo_t* info, void * ucntx, void* pc, void* _fpsimd,
         if(is32bits) addr = (uint8_t*)(((uintptr_t)addr)&0xffffffff);
         uint64_t value;
         if (is_fp) {
+            #ifndef LA64_ABI_1
             struct sctx_info *info = (struct sctx_info*)p->uc_mcontext.__extcontext;
             uint64_t *fpregs = NULL;
             while (info->magic && !fpregs) {
@@ -513,6 +514,9 @@ int sigbus_specialcases(siginfo_t* info, void * ucntx, void* pc, void* _fpsimd,
             if (!fpregs)
                 return 0;
             value = fpregs[val];
+            #else
+            value = p->uc_mcontext.__fpregs->__val64[val];
+            #endif
         } else {
             value = p->uc_mcontext.__gregs[val];
         }
```

Thanks,
Leslie Zhai

ksco: Feel free to do a pull request.

PR #4049, opened 2026-07-16
[DYNAREC] Added BOX64_DYNAREC_DISASM for Zydis disassembly in dynarec dump

Mikura-Kyouka:


ptitSeb:
Your change would make Zydis to initialize all the time, wich is something we do not want. It should initialize only when needed.

ksco:
> wich is something we do not want.

Why?

ptitSeb:
> > wich is something we do not want.
> 
> Why?

We don't need to load stuffs we don't use. Why having more memory consumed for nothing?

ksco:
Only if you enabled `HAVE_TRACE` when compiling, which non-devs should not enable.

ptitSeb:
Still, I don't want lib loaded if we don't use it. Adding a check of dump or trace enable is easy enough, I don't see why we shouldn't keep the test.

Mikura-Kyouka:
My intention is that I think it is not intuitive to use `BOX64_TRACE=0-1` to enable Zydis loading. Maybe a new option can be added for control, and the new option will be automatically turned on when `BOX64_TRACE` is on.

ptitSeb:
> My intention is that I think it is not intuitive to use `BOX64_TRACE=0-1` to enable Zydis loading. Maybe a new option can be added for control, and the new option will be automatically turned on when `BOX64_TRACE` is on.

I understand, and I aggree but I think there is a need for a env. var. option for that.

src/emu/x64run.c
```diff
@@ -78,7 +78,7 @@ int Run(x64emu_t *emu, int step)
 #if defined(HAVE_TRACE)
         __builtin_prefetch((void*)addr, 0, 0); 
         emu->prev2_ip = emu->old_ip;
-        if(my_context->dec && IsTraceAddr(addr))
+        if(my_context->dec && BOX64ENV(trace) && strcmp(BOX64ENV(trace), "0") && IsTraceAddr(addr))
```
ptitSeb: try to avoid the use of `strcmp` in the main loop of the interpreter.

src/include/box64context.h
```diff
@@ -100,7 +100,6 @@ typedef struct box64context_s {
 
     path_collection_t   box64_emulated_libs;    // Collection of libs that should not be wrapped
 
-    int                 x64trace;
```
ptitSeb, line 103: why removing this variable? you could use it instead to know when instruction trace is enabled (instead of when zydis is used for dump purpose for example)

ptitSeb:
Ok, it's nice, but it's either missing something, or the title of the is wrong now.

src/emu/x64run.c
Mikura-Kyouka: Fixed.

src/include/box64context.h
Mikura-Kyouka: Restored.

Mikura-Kyouka:
> Ok, it's nice, but it's either missing something, or the title of the is wrong now.

Sorry I forgot to change the title. Now it is the same as the commit message.

docs/gen/usage.json
```diff
@@ -476,6 +476,25 @@
       }
     ]
   },
+  {
+    "name": "BOX64_DYNAREC_DUMP_ZYDIS",
```
ksco: Please don't include "Zydis" in the option name; we could change the disassembler to something else if needed.

ksco (changes requested):

docs/gen/usage.json
Mikura-Kyouka: What about `BOX64_DYNAREC_DUMP_DISASM`?

ksco: It's used everywhere, not just in dumps: trace, signal, missing opcodes... I think `BOX64_DYNAREC_DISASM` is better.

Mikura-Kyouka: Now using `BOX64_DYNAREC_DISASM`.

ksco:
You should update the commit message as well as PR title to reflect the name change.

Mikura-Kyouka:
> You should update the commit message as well as PR title to reflect the name change.

Sorry, I forgot again... Fixed.

docs/gen/usage.json
```diff
@@ -476,6 +476,25 @@
       }
     ]
   },
+  {
+    "name": "BOX64_DYNAREC_DISASM",
+    "description": "Loads Zydis to display x86 disassembly in BOX64_DYNAREC_DUMP output.",
```
ksco: As I said, it's not just for dumping. Maybe: `Try to initialize libZydis for more readable x86 disassembly in dump, trace, crash report, etc.`

Mikura-Kyouka: Fixed.

docs/gen/usage.json
```diff
@@ -476,6 +476,25 @@
...
+    "configurator": false,
+    "options": [
+      {
+        "key": "0",
+        "description": "Do not load Zydis for dynarec dump.",
+        "default": true
+      },
+      {
+        "key": "1",
+        "description": "Load Zydis for dynarec dump to show x86 disassembly.",
```
ksco: Update the options accordingly.

ksco (changes requested):

docs/gen/usage.json
Mikura-Kyouka: Updated.

Mikura-Kyouka:
Sorry for missing changes again and again...

ksco (approved):

ptitSeb (approved):

PR #4051, opened 2026-07-17
[ENV] Disable dynacache when dynarec_trace and trace is enabled

zengdage:
If dynacache is enabled, the trace code generated by dynarec_trace will be placed into the cache block. If another trace address is switched to, no further trace code will be generated.


ptitSeb:
Would be nice to also disable dynacache if dynacache_dump is enabled.

ptitSeb:
Thanks.

PR #4054, opened 2026-07-17
[LA64_DYNAREC] Added push/pop rsp merge optimization

ksco:


ptitSeb:
I really don't like that `RSP_TRACK` macro that gets in many places... I'm pretty sure there are other way to do this, using PASS0 or PASS1 macro magic to set the rsp state to none by default and override this value by other macros.

ksco:
> I really don't like that `RSP_TRACK` macro that gets in many places... I'm pretty sure there are other way to do this, using PASS0 or PASS1 macro magic to set the rsp state to none by default and override this value by other macros.

Sure, but this is a whitelist, so we need to be explicit. A blacklist is quite complex, with not much gain.

ptitSeb:
> > I really don't like that `RSP_TRACK` macro that gets in many places... I'm pretty sure there are other way to do this, using PASS0 or PASS1 macro magic to set the rsp state to none by default and override this value by other macros.
> 
> Sure, but this is a whitelist, so we need to be explicit. A blacklist is quite complex, with not much gain.

Exactly! That's why default value should be NOPE.
 the PASS1 `NEWINST` can have `dyn->insts[ninst].rsp_class = RSP_CLASS_NOP;` added to it for example, while the `PUSH1mz` macro would set `dyn->insts[ninst].rsp_class = RSP_CLASS_PUSH`; again in PASS1 only (for example).
If you need to put `RSP_TRACK` to all the opcode that are not whitelisted, that's not really a whitelist, that's a blacklist imho.

ksco:
`RSP_TRACK` is used to tell DynaRec this opcode is not a barrier -- it can sit in between multiple push/pop opcodes.

ptitSeb:
> `RSP_TRACK` is used to tell DynaRec this opcode is not a barrier -- it can sit in between multiple push/pop opcodes.

I understand. But I still think it whould be the default behaviour of an opcode (so without the need to use that macro), as you are doing a whitelist stuffs. I'm sure you can acheive the exact same result without this macro just by using PASS1 stuffs

ksco:
I use the manual whitelist because I don't want the `NOP` to be the default behaviour; I want BARRIER to be the default behaviour. Marking most opcodes as NOP will pull in too many things I need to deal with, and those opcodes are not the typical ones that appear in the prologue/epilogue. I know I can make the RSP_TRACK implicit, but that really just turns this into a much bigger PR because of the need to identify every barrier.

Issue #4052, opened 2026-07-17
Jump into an RWX heap page raises endless SIGSEGV (code=0xecec, internal prot=3) while kernel maps say rwx — Palworld dedicated server + UE4SS hooks


Shirowwww:
I'm running the x86_64 Palworld dedicated server (UE 5.1) on ARM64 (Oracle Ampere A1, Neoverse-N1, 4 cores, 4KB pages, kernel 6.17, Debian trixie container) with UE4SS injected through `BOX64_LD_PRELOAD`. UE4SS is a modding framework that installs runtime function hooks (PolyHook2 x64 detours), so during startup it allocates trampolines, patches prologues, and generally writes into executable pages a lot.

About 1 boot in 3, ~25s after the mods finish loading, the guest jumps into a single 4KB `rwxp` page carved out of `[heap]` and box64 raises Signal 11 with `code=0xecec` and internal `prot=3`. But `/proc/<pid>/maps`, captured while the process was still alive, says that page is `rwx`. So box64's internal protection tracking disagrees with the kernel, and CheckExec refuses to run code the kernel would happily execute.

The game's crash handler then re-raises in a loop (`sigaction` + `rt_tgsigqueueinfo`, visible in the rolling log), box64 re-emits at the same RIP, and it never recovers — tens of thousands of `Emit Signal 11` per second until OOM. With `BOX64_SHOWBT=1` the chain reaches SIG_DFL instead and aborts after 2 emissions, which is how I got clean captures.

I caught it 5 times. The signature is always the same:

```
0137|Emit Signal 11 at IP=0x4fa39090(??? / ?) / addr=0x4fa39090, code=0xecec
FSBASE=0x4f913cb0
RAX=0000000000000000 RCX=0000000100000001 RDX=0000000000000000 RBX=000000006d164450
RBP=0000000000000000 RSI=0000000000000000 RDI=000000006d164450  R8=0000000000000000
R10=0000000000000000 R11=cccccccccccccccd R12=000000007a1fc6a0 R13=0000000000000008
RIP=000000004fa39090
Signal 11: si_addr=0x4fa39090, TRAPNO=14, ERR=21, RIP=0x4fa39090, prot=3, mmapped:1
```

maps at crash time:

```
4f875000-4fa39000 rw-p 00000000 00:00 0    [heap]
4fa39000-4fa3a000 rwxp 00000000 00:00 0    [heap]   <-- RIP lands here
[...]

PR #4055, opened 2026-07-17
[CUSTOMMEM] only track the newly added heap when brk changed

Shirowwww:
Proper fix for #4052, replacing the CheckExec workaround in #4053.

`find31bitBlockNearHint` / `find47bitBlockNearHint` re-tracked the whole heap from `pbrk` as RW every time the program break moved. `setProtection` only writes the tracking, not the kernel, so every page the guest had made executable inside the heap silently lost PROT_EXEC in the tracking while the kernel still mapped it rwx. `CheckExec` then saw prot=3 against a kernel 0x7 and raised 0xecec forever at that RIP.

Hit it with UE4SS on a Palworld dedicated server: it puts hook trampolines on heap pages, the heap grows, the trampolines lose their exec bit in the tracking, and the server dies in a signal storm about one boot in three.

Now only the range actually added since last time is tracked, so existing heap pages keep their protection.

Tested on ARM64 (Neoverse-N1), Palworld dedicated server + UE4SS, 8 boots: no storm. I kept the CheckExec resync from #4053 in the test binary purely as a canary that logs when tracking and kernel disagree — with the same binary before this fix it fired on 5 of 8 boots, after it never fired once.

#4053 can be closed if you take this one.

### DIFF
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -2831,11 +2831,8 @@ int memExist(uintptr_t addr)
 #define MEDIUM (void*)0x40000000
 #define HIGH   (void*)0x60000000
 
-// The guest can grow its heap with brk(), which box64 doesn't see as an mmap, so
-// the new memory is tracked here instead. Only track what was actually added:
-// re-tracking the whole heap from pbrk as RW would clobber PROT_EXEC on pages the
-// guest had made executable inside the heap (JIT/hook trampolines), leaving the
-// tracking out of sync with the kernel until CheckExec faults on them.
+// only track what brk added: re-tracking the whole heap would lose PROT_EXEC on
+// pages the guest made executable in it (like hook trampolines)
 static void catchup_brk_protection(void)
 {
     if(pbrk && cur_brk && *cur_brk!=old_brk) {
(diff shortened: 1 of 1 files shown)


src/custommem.c
```diff
@@ -2831,13 +2831,25 @@ int memExist(uintptr_t addr)
 #define MEDIUM (void*)0x40000000
 #define HIGH   (void*)0x60000000
 
-void* find31bitBlockNearHint(void* hint_, size_t size, uintptr_t mask)
+// The guest can grow its heap with brk(), which box64 doesn't see as an mmap, so
+// the new memory is tracked here instead. Only track what was actually added:
+// re-tracking the whole heap from pbrk as RW would clobber PROT_EXEC on pages the
+// guest had made executable inside the heap (JIT/hook trampolines), leaving the
+// tracking out of sync with the kernel until CheckExec faults on them.
+static void catchup_brk_protection(void)
```
ptitSeb, line 2836: Typical AI over-commenting things?

Shirowwww: Ha, busted. Yes I used AI to help me dig through box64, I don't know this codebase and I was stuck on that storm for a while, it saved me a lot of time getting around the memory tracking code.

The finding and the testing are mine though, I went through the fix by hand before sending it and I understand what it does. Sorry for the essay, trimmed it down to 2 lines.

ptitSeb: No problem. That's still a nice finding. I'm fine with AI, but I don't like the way it over comment everything...

ptitSeb:
Thanks.

PR #4062, opened 2026-07-18
[CORE] Don't consume guest t/test arguments

Sped0n:
## Problem

Since #3789, Box64 scans the complete command line for `-t` and `--test` before identifying the guest executable.

Which means, guest-owned arguments can incorrectly activate Box64's unit-test runner:

```
box64 ./guest -t value
```

Instead of running the guest, Box64 exits with:

```
[BOX64] Failed to parse JSON configuration.
```

## Minimal Reproduction

> [!NOTE]
> LLM assistance was used to generate these commands.

Tested on a aarch64 host (macOS 26):

```bash
# 0.4.2 (broken)
REV=7eeb50164 docker run --rm --platform linux/arm64 \
  -e REV \
  -v "$PWD:/src:ro" \
  ubuntu:24.04 \
  bash -lc '
    set -eux
    apt-get update
    DEBIAN_FRONTEND=noninteractive apt-get install -y \
      --no-install-recommends build-essential cmake git python3
    git clone /src /work
    cd /work
    git checkout "$REV"
    cmake -S . -B build \
      -DARM64=ON \
      -DCMAKE_BUILD_TYPE=Release \
      -DNO_LIB_INSTALL=ON
    cmake --build build -j6
    BOX64_LOG=0 BOX64_NOBANNER=1 \
      ./build/box64 ./tests/test04 -t value
  '

# 0.4.0
REV=dae0917c4 docker run --rm --platform linux/arm64 \
  -e REV \
  -v "$PWD:/src:ro" \
  ubuntu:24.04 \
  bash -lc '
    set -eux
    apt-get update
    DEBIAN_FRONTEND=noninteractive apt-get install -y \
      --no-install-recommends build-essential cmake git python3
    git clone /src /work
    cd /work
    git checkout "$REV"
    cmake -S . -B build \
      -DARM64=ON \
      -DCMAKE_BUILD_TYPE=Release \
      -DNO_LIB_INSTALL=ON
    cmake --build build -j6
    BOX64_LOG=0 BOX64_NOBANNER=1 \
      ./build/box64 ./tests/test04 -t value
  '
```

## Changes Overview

The unit-test parser now owns both command-line detection and parsing, and it returns `-1` when it encounters a non-test operand before `-t` or `--test`. 

This establishes the guest executable as the argument-ownership boundary without duplicating the unit-test option grammar in `src/core.c`.

Also unit-test options can stay unordered, so both forms continue to work:

```text
box64 -n 2 -t test.asm
box64 -t test.asm -n 2
```

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -837,14 +837,7 @@ int initialize(int argc, const char **argv, char** env, x64emu_t** emulator, elf
         exit(0);
     }
 
-    int is_test = 0;
-    for (int i = 1; i < argc; i++) {
-        if (!strcmp(argv[i], "--test") || !strcmp(argv[i], "-t")) {
-            is_test = 1;
-            break;
-        }
-    }
-    if (is_test) {
+    if (argc >= 3 && (!strcmp(argv[1], "--test") || !strcmp(argv[1], "-t"))) {
         box64_unittest_mode = 1;
         exit(unittest(argc, argv));
     }


Sped0n:
cc @ksco 

Perhaps a regression test is needed?

ksco:
Nice finding, thanks. I think a simpler way is to force the `-t/--test` to be the second parameter to enter test mode.

Sped0n:
Can, will revert to the behavior before #3789.

ksco (approved):

PR #4068, opened 2026-07-20
[DYNACACHE] Restricted last_mmap to executable images

ksco:


ptitSeb:
CI is red on Android & TERMX with 
image

ksco:
Sorry

src/wrapped/wrappedlibc.c
```diff
@@ -3614,9 +3628,13 @@ EXPORT void* my_mmap64(x64emu_t* emu, void *addr, size_t length, int prot, int f
     if(ret!=MAP_FAILED) {
         if (emu && !(flags & MAP_ANONYMOUS) && (fd > 0)) {
             // the last_mmap will allow mmap created by wine, even those that have hole, to be fully tracked as one single mmap
-            if((ret>=last_mmap_addr[0]) && ret+length<(last_mmap_addr[0]+last_mmap_len[0]))
+            // check if the file is actually a PE file to be safe.
+            int in_last_mmap_0 = (ret >= last_mmap_addr[0]) && ret + length < (last_mmap_addr[0] + last_mmap_len[0]);
+            int in_last_mmap_1 = (ret >= last_mmap_addr[1]) && ret + length < (last_mmap_addr[1] + last_mmap_len[1]);
+            int is_executable = (in_last_mmap_0 || in_last_mmap_1) && is_elf_or_pe(fd);
```
ptitSeb, line 3634: is there a risk of slowing down mmap64 with many reading of header?
Also, maybe the last_mmap should also track `fd` to make sure the map is from the same one (and that might also help not doing `is_elf_of_pe(fd);` for each mmap64 with an file descriptor.

ksco: > is there a risk of slowing down mmap64 with many reading of header?

If `(in_last_mmap_0 || in_last_mmap_1)` is true, the slow function `RecordEnvMappings` will be called anyway. To me, it does not hurt to have 2 extra `pread` calls on top of that.

ksco: tracking fd is not really safe as it can be reused.

PR #4079, opened 2026-07-21
[RV64_DYNAREC] Reduce instruction count for 64-bit addition carry flag

zengdage:
For the risc-v instruction `add rd, rs1, rs2`, if an overflow occurs, only the lower XLEN bits of the sum are retained, and the resulting value will be numerically smaller than both `rs1` and `rs2`.

The combination of  risc-v  `add` and `sltu` instructions is commonly used to compute the carry flag of a 64-bit addition.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_emit_math.c
+++ b/src/dynarec/rv64/dynarec_rv64_emit_math.c
@@ -32,26 +32,16 @@ void emit_add32(dynarec_rv64_t* dyn, int ninst, rex_t rex, int s1, int s2, int s
     }
     IFX (X_CF) {
         if (rex.w) {
-            ZEXTW2(s5, s1);
-            if (cpuext.zba) // lo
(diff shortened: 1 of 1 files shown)


ksco (approved):
Interesting, thanks!

ksco (changes requested):
You can move `SET_FLAGS_NEZ(s5, F_CF, s4);` to the else block and use `OR(xFlags, xFlags, s5);` in the if block to save 1 more instuction.

zengdage:
> You can move `SET_FLAGS_NEZ(s5, F_CF, s4);` to the else block and use `OR(xFlags, xFlags, s5);` in the if block to save 1 more instuction.

Done.

ksco:
The code is pretty obvious, so I think it's better to just remove those comments.

ksco (approved):

PR #4083, opened 2026-07-21
[DYNAREC] Fixed CALL FAR/RET with callret=2

ksco:


src/dynarec/arm64/dynarec_arm64_00.c
ptitSeb, line 4680: Please move MARK outside of the `if` (like just before or after the `CALLRET_RET` macro)

src/dynarec/la64/dynarec_la64_00.c
ptitSeb, line 4423: Please move MARK outside of the `if` (like just before or after the `CALLRET_RET` macro)

PR #4085, opened 2026-07-21
[DRAFT] custommem: probe for 48-bit address space when the maps scan finds none

### LINKED ISSUE #4084: custommem: 48-bit address space misdetected as 39-bit when /proc/self/maps has no high mapping
### Summary

`loadProtectionFromMap()` only sets `have48bits` when `/proc/self/maps` **already** contains a mapping above `0x7fff00000000`. In freshly started processes and in many container environments no such high mapping exists yet, even though the kernel fully supports a 48-bit address space. box64 then misclassifies the host as 39-bit and packs allocations low.

### Repro / evidence

On an aarch64 Ubuntu container with a 48-bit-capable kernel, box64 logs:

```
[BOX64] Didn't detect 48bits of address space, considering it's 39bits
```

…yet the kernel clearly supports 48-bit addresses — a direct probe is honored:

```
mmap(0x7fff00000000, ...) -> 0x7fff00000000
mmap(NULL, ...)           -> 0x0000ffff8a770000
```

The heuristic simply never observed an existing high mapping at startup, so it never set the flag.

### Root cause

`src/custommem.c`, in the maps scan:

```c
if(s>0x7fff00000000LL)
    have48bits = 1;
```

`have48bits` is only turned on if such a mapping is *already present*; nothing probes the actual capability. A process that hasn't yet mapped anything high is misclassified.

### Suggested fix

If the scan didn't reveal a high mapping, actively probe: try to reserve a page at `0x7fff00000000`. If the kernel places it there, the address space is at least 48-bit; on a genuine 39-bit system the reservation lands low and `have48bits` stays `0`, so behaviour there is unchanged. PR #4085.

### Honest scope note

I found this while chasing a **separate** Chromium/PartitionAlloc startup crash in an emulated Electron app (Signal Desktop's amd64 build). I initially suspected the 48-bit misdetection was the cause — it is **not**. After building the probe fix, box64 correctly reports 48-bit, but that Electron app still crashes in `PartitionAlloc::Free` (a distinct allocator-provenance issue where box64-allocated pointers are handed to PA's `free`). So this report/PR is purely about the **detection heuristic being incorrect** and making it robust — not a fix for that crash. I didn't want to overstate it.

dmitry-brazhenko:
Fixes #4084.

`loadProtectionFromMap()` only set `have48bits` when `/proc/self/maps` already had a mapping above `0x7fff00000000`. In fresh processes / containers there is none yet even on a 48-bit-capable kernel, so box64 wrongly assumed 39-bit and packed allocations low.

Instead, probe: if the scan found nothing high, try to reserve a page at `0x7fff00000000`. If the kernel places it there we have 48-bit; on a real 39-bit system it lands low and `have48bits` stays 0 (unchanged).

### DIFF
--- a/src/custommem.c
+++ b/src/custommem.c
@@ -2766,6 +2766,15 @@ void loadProtectionFromMap()
                 have48bits = 1;
         }
     }
+    if(!have48bits && !box64_is32bits) {
+        void* probe = InternalMmap((void*)0x7fff00000000LL, box64_pagesize, PROT_NONE,
+                                   MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0);
+        if(probe!=MAP_FAILED) {
+            if((uintptr_t)probe>=0x7fff00000000LL)
+                have48bits = 1;
+            InternalMunmap(probe, box64_pagesize);
+        }
+    }
     static int shown48bits = 0;
     if(!shown48bits) {
         shown48bits = 1;


src/custommem.c
```diff
@@ -2766,6 +2766,23 @@ void loadProtectionFromMap()
                 have48bits = 1;
         }
     }
+    // The scan above only sets have48bits if an existing mapping already lives
```
ptitSeb: please remove this AI generated comment, it's not usefull. The code is self-explanatory

dmitry-brazhenko: Done, removed it. Thanks for the review!

ptitSeb:
Not a [DRAFT] anymore I suppose?

dmitry-brazhenko:
Still in progress, I will ping you a little later

dmitry-brazhenko:
Consolidated into #4087 (kept as a draft while we validate). Your comment is addressed there — the AI-generated comment is removed. Thanks!

PR #4086, opened 2026-07-21
malloc_hack: redirect captured function pointers and direct calls

dmitry-brazhenko:
`getAlternate` (how `BOX64_MALLOC_HACK` reroutes malloc/free) was only applied on some paths. When a program takes the **address** of the overridden malloc/free without a symbol call — a data relocation (`RELATIVE`/`GLOB_DAT`/`64`) or an arm64 direct `CALL` — the raw address was kept and the redirect bypassed.

Chromium/Electron (PartitionAlloc) capture `free`'s address into their allocator tables and also call it directly, so they mix box64- and app-allocated pointers under `malloc_hack` and crash in `free()`.

Apply `getAlternate` on those paths too:
- `RelocateElfRELA`: `RELATIVE`, `GLOB_DAT`, `JUMP_SLOT`, `64` — matching what `PltResolver64` already does for lazy binding.
- arm64 `0xE8 CALL` — matching rv64/la64/ppc64le, which already do this.

**Result:** an unmodified amd64 Signal Desktop (Electron) now boots under box64 on aarch64 instead of crashing in `PartitionAlloc::Free`.

Note: no unit test — the behaviour only appears with a real overriding allocator under `malloc_hack`; a synthetic override perturbs libc startup. Validated by running the app; the change only extends existing `getAlternate` coverage.

### DIFF
--- a/src/dynarec/arm64/dynarec_arm64_00.c
+++ b/src/dynarec/arm64/dynarec_arm64_00.c
@@ -3916,6 +3916,7 @@ uintptr_t dynarec64_00(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
                         j64 = (uint32_t)(addr+i32);
                     else
                         j64 = addr+i32;
+                        j64 = (uintptr_t)getAlternate((void*)j64);
                     jump_to_next(dyn, j64, 0, ninst, rex.is32bits);
                     CALLRET_RET(can_continue);
                     MARK;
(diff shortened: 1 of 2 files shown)


src/dynarec/arm64/dynarec_arm64_00.c
```diff
@@ -3916,6 +3916,7 @@ uintptr_t dynarec64_00(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
                         j64 = (uint32_t)(addr+i32);
                     else
                         j64 = addr+i32;
+                        j64 = (uintptr_t)getAlternate((void*)j64);
```
ptitSeb, line 3919: Bad alignement, it should not be align to the else branch but the else itself.

dmitry-brazhenko:
Consolidated into #4087 together with the 48-bit fix, per review. Closing this one.

PR #4089, opened 2026-07-22
[CORE] Add instruction translation statistics tool

zengdage:
1. Add statistics tool to track the average number of native instructions generated for x86-64 instruction, helping identify potential bottlenecks in instruction translation.

2. Add `BOX64_DYNAREC_INST_STATS` to control it, default as 0.

3. Add `dyn->stats` condition checks at `INST_NAME` and `INST_EPILOGUE` in pass2.h, performance impact is limited by dcache and branch prediction.

```
Total unique x86-64 instructions: 55

[BOX64] Address          Size     Nat Insts    Count    Avg Nat    Inst Ratio   Name
[BOX64] ---------------- -------- ------------ -------- ---------- ------------ --------------------------------------
[BOX64] 0x100001140      4        0            13       0          0.000000     NOP / ENDBR32 / ENDBR64
[BOX64] 0x100001144      2        74           32       2          2.312500     XOR Ed, Gd
[BOX64] 0x100001146      3        28           28       1          1.000000     MOV Ed, Gd
[BOX64] 0x100001149      1        22           11       2          2.000000     POP reg
[BOX64] 0x10000114d      4        1            1        1          1.000000     AND Ed, Ib
[BOX64] 0x100001151      1        24           12       2          2.000000     PUSH reg
[BOX64] 0x100001153      3        1            1        1          1.000000     XOR Ed, Gd
[BOX64] 0x100001158      7        126          40       3          3.150000     LEA Gd, Ed
[BOX64] 0x10000115f      6        28           1        28         28.000000    CALL Ed
[BOX64] 0x30020000       19       172          4        43         43.000000    Special Box64 instruction
[BOX64] 0x30020013       1        322          14       23         23.000000    RET
[BOX64] 0x100001004      4        7            7        1          1.000000     SUB Ed, Ib
[BOX64] 0x100001008      7        16           4        4          4.000000     MOV Gd, Ed
[BOX64] 0x10000100f      3        12           3        4          4.000000     TEST Ed, Gd
[BOX64] 0x100001012      2        28           14       2          2.000000     JZ ib
[BOX64] 0x100001014      2        24           1        24         24.000000    CALL Ed
```


ksco (changes requested):
Sorry but how is this useful?

zengdage:
> Sorry but how is this useful?

This tool counts the approximate number of native instructions generated for each type of x86‑64 instruction. If a particular x86‑64 instruction results in a high native instruction count, we can target it for software optimization, or even implement dedicated CPU instructions to accelerate binary translation.


For example,
```
[BOX64] 0x100001074      7        75           3        25.000000    JMP Ed`
```
We can observe that roughly 25 native instructions are required to emulate a single JMP instruction. We can then conduct an in-depth analysis to see whether there is room for optimization in the JMP translation path to reduce the instruction count, or explore defining a new dedicated CPU instruction to accelerate this operation.

ksco:
If you really find these stats useful (which I don't), you can write a simple Python script and compute this table from the output of `BOX64_DYNAREC_DUMP=3`.

zengdage:
> BOX64_DYNAREC_DUMP=3

Yes. we could achieve the same effect by enabling `BOX64_DYNAREC_DUMP` and processing the output with scripts. However, when running large applications, `BOX64_DYNAREC_DUMP` generates an enormous volume of logs and noticeably slows down program execution.

ksco:
`Box64_DYNAREC_DUMP` is versatile enough and not that slow; I have written many one-time scripts to extract some information from the dump of huge 3A games in the past, never been an issue. Adding new environment variables for every insight that can be derived from the dump will bloat the codebase. Box64 already has too many environment variables IMO.

ptitSeb:
I'm afraid I, too, don't see the advantage of this stats. Most opcodes have huge variability because of flags or other local optimisation, and having the size of the generated code is not really representative of anything IMO.
If you are writting a paper or something and needs that kind of stats, I can understand, but I don't think other developper would benefit from this statistic.

zengdage:
> `Box64_DYNAREC_DUMP` is versatile enough and not that slow; I have written many one-time scripts to extract some information from the dump of huge 3A games in the past, never been an issue. Adding new environment variables for every insight that can be derived from the dump will bloat the codebase. Box64 already has too many environment variables IMO.

Get it. close this or.

PR #4092, opened 2026-07-22
Removed useless files

ksco:


ptitSeb:
Those files should be left, there are here to be picked up by the various AI coding agent to make sure the set of rules are used (unless those agents evolved and are ignoring those files now?)

ksco:
They all recognise AGENTS.md

Ari4ka (approved):

PR #4099, opened 2026-07-23
[RV64_DYNAREC] Use risc-v zicond to eliminate some branchs for setting flag

zengdage:
When setting flags, using `Zicond `eliminates the branch at the cost of one extra instruction.

### DIFF
--- a/src/dynarec/rv64/dynarec_rv64_helper.h
+++ b/src/dynarec/rv64/dynarec_rv64_helper.h
@@ -903,6 +903,10 @@
         if (cpuext.xtheadcondmov) {         \
             ORI(scratch, xFlags, 1 << F);   \
             TH_MVNEZ(xFlags, scratch, reg); \
+        } else if (cpuext.zicond) {         \
+            ADDI(scratch, xZR, 1 << F);     \
+            CZERO_EQZ(scratch, scratch, reg); \
+            OR(xFlags, xFlags, scratch);    \
         } else {                            \
             BEQZ(reg, 8);                   \
             ORI(xFlags, xFlags, 1 << F);    \
@@ -914,6 +918,10 @@
         if (cpuext.xtheadcondmov) {         \
             ORI(scratch, xFlags, 1 << F);   \
             TH_MVEQZ(xFlags, scratch, reg); \
+        } else if (cpuext.zicond) {         \
+            ADDI(scratch, xZR, 1 << F);     \
+            CZERO_NEZ(scratch, scratch, reg); \
+            OR(xFlags, xFlags, scratch);    \
         } else {                            \
             BNEZ(reg, 8);                   \
             ORI(xFlags, xFlags, 1 << F);    \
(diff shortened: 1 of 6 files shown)


ksco (changes requested):
I'm not sure about this, it's 3 instructions. Please make a benchmark to confirm this is actually faster.

zengdage:
> I'm not sure about this, it's 3 instructions. Please make a benchmark to confirm this is actually faster.

Sorry,  I can't find a suitable x86 application to test this scenario, since it is strongly tied to branch mispredictions, and performance behavior can vary widely across different cases.

I can only run tests from the perspective of the risc-v isa in `SpacemiT K3` board. 

1. When the branch misprediction rate is high (in my test, branch miss 25%), Zicond can speedup `2×` compared to using beqz.  
2. When the branch misprediction rate is close to 0.0%, their execution times are almost equal.

branch miss 25%
```
czero.eqz elapsed time: 1433485.78 us
beqz elapsed time: 4622089.78 us
```

`rand01` always return 0, and branch miss become 0%
```
czero.eqz elapsed time: 956166.24 us
beqz elapsed time: 955717.83 us
```

compile command: `gcc test_czero.c -o test_czero -O2 -DZICOND`

```
#include <stdlib.h>
#include <string.h>
#include <stdio.h>
#include <time.h>

#define TEST_SIZE 1024*1024
#define TEST_NUM 1000

#ifdef ZICOND
#define NAME "czero.eqz"
#else
#define NAME "beqz"
#endif

static inline int rand01(void)
{
    return rand() > (RAND_MAX / 2);
}


int main(int argc, char **argv)
{
    int i;
    unsigned char temp;
    struct timespec start, end;
    unsigned char *data = malloc(TEST_SIZE);
    memset(data, 0, TEST_SIZE);

    srand((unsigned)time(NULL));

    for(i = 0; i < TEST_SIZE; i++)
        data[i] = rand01();

    for(i = 0; i < TEST_SIZE; i++) {
        temp = *(volatile unsigned char *)(data + i);
        asm volatile("" : : "r"(temp) : "memory");
    }

    clock_gettime(CLOCK_MONOTONIC, &start);

    asm volatile(
        "li     t4, 0\n"
        "mv     t5, %[num]\n"
        "li     t6, 3\n"
        "L0:\n"
        "mv     t1, %[base]\n"
        "add    t2, t1, %[size]\n"
        "L1:\n"
        "lbu    t0, 0(t1)\n"
#ifdef ZICOND
        "addi   t3, zero, 4\n"
        "czero.eqz t3, t3, t0\n"
        "or     t6, t6, t3\n"
#else
        "beqz   t0, L2\n"
        "ori    t6, t6, 4\n"
#endif
        "L2:\n"
        "addi   t1, t1, 1\n"
        "bltu   t1, t2, L1\n"
        "addi   t4, t4, 1\n"
        "bltu   t4, t5, L0\n"
        :
        : [base] "r"(data), [size] "r"(TEST_SIZE), [num] "r"(TEST_NUM)
        : "t0", "t1", "t2", "t3", "t4", "t5", "t6", "memory");

    clock_gettime(CLOCK_MONOTONIC, &end);

    long seconds = end.tv_sec - start.tv_sec;
    long nanos = end.tv_nsec - start.tv_nsec;
    double elapsed_us = seconds * 1e6 + nanos / 1e3;

    printf("%s elapsed time: %.2f us\n", NAME, elapsed_us);
    return 0;
}
```

ksco (approved):
LGTM.

PR #4100, opened 2026-07-23
[RV64_DYNAREC] Remove redundant SMREAD

zengdage:
Frankly, I am not sure whether this` SMREAD` should be removed. 

Based on my understanding, `GETEX` already performs `SMREAD` in the` !MODREG` block, while `GETGX` only contains store operations. So I think it can be removed.


ptitSeb:
I agree, it can be removed. It's because the ARM64 version of this opcode doesn't make use of those GETEX macro and so needed the SMREAD.

ksco:
Sorry for the delay, I forgot about this.

PR #4107, opened 2026-07-24
[INTERPRETER] Improve x87 long double classification

yzewei:
Keep the existing raw 80-bit x87 shadow aligned through FLD ST(i), FXCH, FST/FSTP, FCMOV, FABS, and FCHS, and use it for FCOMI/FUCOMI/FUCOMIP when the shadow is still valid.

This avoids falling back to the double representation for long double classification, where large finite values such as LDBL_MAX may collapse to inf and break isinf/isnan/isfinite-style checks.

# Reproduce

```
@aosc-de793919 [ ChatTTS@main ] ! 
  BOX64_DYNAREC=0 BOX64_LOG=0 PYTHONFAULTHANDLER=1 \
  box64 python3 -X faulthandler - <<'PY'
import numpy as np
                  
for name, val in [
    ("+inf", np.inf),
    ("-inf", -np.inf),
    ("nan", np.nan),
]:                                   
    v = np.array([val], dtype="g")[0]
    print(name, v.tobytes().hex(),
          "isinf=", bool(np.isinf(v)),
          "isfinite=", bool(np.isfinite(v)),
          "isnan=", bool(np.isnan(v)),
          flush=True)
PY
done
===== BOX64_DYNAREC=0 =====
[BOX64] Box64 loongarch64 v0.4.3 9c0214dd3 with Dynarec built on Jul 22 2026 21:37:10
+inf 0000000000000080ff7f1600ff7f0000 isinf= False isfinite= True isnan= False
-inf 0000000000000080ffff1600ff7f0000 isinf= False isfinite= True isnan= False
nan 00000000000000c0ff7f1600ff7f0000 isinf= False isfinite= False isnan= True
@aosc-de793919 [ ChatTTS@main ] $ BOX64_DYNAREC=0 BOX64_LOG=0 PYTHONFAULTHANDLER=1   box64 python3 -X faulthandler - <<'PY'
import numpy as np

for name, val in [
    ("+inf", np.inf),
    ("-inf", -np.inf),
    ("nan", np.nan),
]:
    v = np.array([val], dtype="g")[0]
    print(name, v.tobytes().hex(),
          "isinf=", bool(np.isinf(v)),
          "isfinite=", bool(np.isfinite(v)),
          "isnan=", bool(np.isnan(v)),
          flush=True)
PY
 done
===== BOX64_DYNAREC=0 =====
[BOX64] Box64 loongarch64 v0.4.3 40a841df8 with Dynarec built on Jul 23 2026 21:41:27
+inf 0000000000000080ff7f1600ff7f0000 isinf= True isfinite= False isnan= False
-inf 0000000000000080ffff1600ff7f0000 isinf= True isfinite= False isnan= False
nan 00000000000000c0ff7f1600ff7f0000 isinf= False isfinite= False isnan= True

```

### DIFF
--- a/src/emu/x87emu_private.h
+++ b/src/emu/x87emu_private.h
@@ -24,6 +24,37 @@ typedef struct x64emu_s x64emu_t;
 #define STld(a)  emu->fpu_ld[(emu->top+(a))&7]
 #define STll(a)  emu->fpu_ll[(emu->top+(a))&7]
 
+// Use a NaN payload as invalid marker so a valid +0.0 reference stays usable.
+#define FPU_LD80_INVALID_REF UINT64_C(0x7ff8deadbeefc0de)
+static inline void fpu_ld80_clear(x64emu_t* emu, int i)
+{
+    STld(i).uref = FPU_LD80_INVALID_REF;
+}
+
+static inline int fpu_ld80_raw_valid(x64emu_t* emu, int i)
+{
+    return STld(i).uref != FPU_LD80_INVALID_REF && ST(i).q==STld(i).uref;
+}
+
+static inline void fpu_ld80_copy(x64emu_t* emu, int dst, int src)
+{
+    ST(dst).q = ST(src).q;
+    if(fpu_ld80_raw_valid(emu, src))
+        STld(dst) = STld(src);
+    else
+        fpu_ld80_clear(emu, dst);
+}
+
+static inline void fpu_ld80_swap(x64emu_t* emu, int a, int b)
+{
+    uint64_t q = ST(a).q;
+    fpu_ld_t tmp = STld(a);
+    ST(a).q = ST(b).q;
+    ST(b).q = q;
+    STld(a) = STld(b);
+    STld(b) = tmp;
+}
+
 static inline void fpu_do_push(x64emu_t* emu)
 {
     int newtop = (emu->top-1)&7;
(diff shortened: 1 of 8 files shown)


src/emu/x87emu_private.h
```diff
@@ -24,6 +24,35 @@ typedef struct x64emu_s x64emu_t;
...
 #define STll(a)  emu->fpu_ll[(emu->top+(a))&7]
 
+static inline void fpu_ld80_clear(x64emu_t* emu, int i)
+{
+    STld(i).uref = 0;
+}
+
+static inline int fpu_ld80_raw_valid(x64emu_t* emu, int i)
+{
+    return STld(i).uref && ST(i).q==STld(i).uref;
```
ptitSeb: That way you make 80bits subnormal might gets undetected!

I would suggest to use a NAN with a random payload (via some macro constant) as a signature instead of 0

yzewei: it's right. Thanks!

src/emu/x87emu_private.h
```diff
@@ -89,6 +122,79 @@ static inline void fpu_fcom(x64emu_t* emu, double b)
...
 
+#ifndef HAVE_LD80BITS
+// 1 => a > b
+// 0 => a == b
+// -1 => a < b
+// 2 => unordered
+static inline int fpu_ld80_raw_cmp(const longdouble_t* a, const longdouble_t* b)
+{
+    uint16_t a_upper = a->l.upper, b_upper = b->l.upper;
+    uint64_t a_lower = a->l.lower, b_lower = b->l.lower;
```
ptitSeb, line 133: I don't think a_upper & b_upper are usefull.

PR #4114, opened 2026-07-25
[LA64_DYNAREC] Optimized jump safe macros

ksco:


src/dynarec/la64/dynarec_la64_0f.c
```diff
@@ -1449,14 +1449,14 @@ uintptr_t dynarec64_0F(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
         /* out of the block */                                                              \
         i32 = dyn->insts[ninst].epilog - (dyn->native_size);                                \
         if (COMIS_FUSED()) {                                                                \
-            COMIS_BRANCH_NOT_TAKEN(i32);                                                    \
+            COMIS_BRANCH_NOT_TAKEN(i32, dyn->size);                                         \
```
ptitSeb: I don't understand why `_safe` jump are used here, as the jump is internal to the opcode

ptitSeb: but for jump to the end of the opcode itself, I think `ninst` is better than `dyn->size` as this one is updated only on pass0 and will be a constant for all other passes.

ksco: > I don't understand why _safe jump are used here, as the jump is internal to the opcode

Seems unnecessary indeed, but I guess it does not matter anymore.

PR #4136, opened 2026-07-27
[SYSCALL] Add more syscalls for Magisk

### LINKED ISSUE #1497: Unsupported Syscall 0x10Dh (269)
Trying to run Magisk on Linux, and it works fine on WSL in Windows but not on my Raspberry Pi with Box64. This is the error I keep getting: `Error: Unsupported Syscall 0x10Dh (269)`.

Anything I can do to solve this?

LukeShortCloud:
95, 115, and 269.

Resolves #1497


LukeShortCloud:
Magisk finally works with these new system call forwarders.

PR #4139, opened 2026-07-28
[LA64_DYNAREC] Speed up autocrc validation using exact match

ksco:
This is about 3x faster than the scalar CRC approach with more memory consumption for always_test blocks. Given that always_test blocks should be rare, the trade-off is likely worthwhile.


ptitSeb:
I have a hard time believing that comparing entire sources (so reading 2 blocks of the size) is faster than doing a crc (so reading only 1 block), no talking about the memory consumption.
Also, don't forget that on 16k pagsesize, many more page will have the NEVERCLEAN attribute.

ksco:
I noticed the bottleneck on a 16K page-size kernel running a benchmark. With this patch, that benchmark runs 3x faster. I also wrote a dedicated test to compare the 2 approach:

Scalar CRC: 1.0x
LSX: 4.03x faster
LASX: 7.72x faster

ksco:
> Also, don't forget that on 16k pagsesize, many more page will have the NEVERCLEAN attribute.

I try to avoid introducing a new env var on this. Do you think it's necessary?

ptitSeb:
I think it woul dbe better to come with an improved CRC algorythm instead. I really don't like the memory copy approach

ksco:
I'll think about this.

ptitSeb:
> I'll think about this.

Also, it would be worth just checking if the current autocrc works 100 fine  and not generate false negative and force block to tbe rebuilt

ksco:
Nah, it's working as expected.

ksco:
Abandon this in favour of #4178.

PR #4142, opened 2026-07-28
[LA64_DYNAREC] Add optional x87 raw long double compare

yzewei:
Improved x87 long double comparisons on LA64 dynarec.

Fixed long double classification issues, such as NumPy's `isinf/isfinite` operations on `np.longdouble` values.

# before
```
BOX64_DYNAREC=1 BOX64_LOG=0 PYTHONFAULTHANDLER=1  box64  python3 -X faulthandler - <<'PY'
import numpy as np                                                      
for name, val in [("+inf", np.inf), ("-inf", -np.inf), ("nan", np.nan)]:
    v = np.array([val], dtype="g")[0]
    print(name, bool(np.isinf(v)), bool(np.isfinite(v)), bool(np.isnan(v)))
PY
[BOX64] Box64 loongarch64 v0.4.3 with Dynarec built on Jul 28 2026 07:23:13
+inf False True False
-inf False True False
nan False False True

```

# after
```
BOX64_DYNAREC=1 BOX64_LOG=0 PYTHONFAULTHANDLER=1  box64  python3 -X faulthandler - <<'PY'
import numpy as np                                                      
for name, val in [("+inf", np.inf), ("-inf", -np.inf), ("nan", np.nan)]:
    v = np.array([val], dtype="g")[0]
    print(name, bool(np.isinf(v)), bool(np.isfinite(v)), bool(np.isnan(v)))
PY
[BOX64] Box64 loongarch64 v0.4.3 with Dynarec built on Jul 28 2026 08:23:30
+inf True False False
-inf True False False
nan False False True

```

### DIFF
--- a/src/dynarec/la64/dynarec_la64_d9.c
+++ b/src/dynarec/la64/dynarec_la64_d9.c
@@ -66,6 +66,9 @@ uintptr_t dynarec64_D9(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 // swap the cache value, not the double value itself :p
                 x87_get_st(dyn, ninst, x1, x2, nextop & 7, X87_ST(nextop & 7));
                 x87_get_st(dyn, ninst, x1, x2, 0, X87_ST0);
+                if (!BOX64ENV(dynarec_fastround)) {
+                    x87_swap_fpu_ld(dyn, ninst, x1, x2, x3, x4, 0, nextop & 7);
+                }
                 x87_swapreg(dyn, ninst, x1, x2, 0, nextop & 7);
                 // should set C1 to 0
                 break;
@@ -92,7 +95,14 @@ uintptr_t dynarec64_D9(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 if (ST_IS_F(0)) {
                     FNEG_S(v1, v1);
                 } else {
+                    if (!BOX64ENV(dynarec_fastround)) {
+                        MOVFR2GR_D(x2, v1);    // old ST0.q
+                    }
                     FNEG_D(v1, v1);
+                    if (!BOX64ENV(dynarec_fastround)) {
+                        MOVFR2GR_D(x3, v1);    // new ST0.q
+                        x87_update_fpu_ld_sign(dyn, ninst, x1, x4, x5, x2, x3, 0, 1);
+                    }
                 }
                 break;
             case 0xE1:
@@ -101,7 +111,14 @@ uintptr_t dynarec64_D9(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 if (ST_IS_F(0)) {
                     FABS_S(v1, v1);
                 } else {
+                    if (!BOX64ENV(dynarec_fastround)) {
+                        MOVFR2GR_D(x2, v1);    // old ST0.q
+                    }
                     FABS_D(v1, v1);
+                    if (!BOX64ENV(dynarec_fastround)) {
+                        MOVFR2GR_D(x3, v1);    // new ST0.q
+                        x87_update_fpu_ld_sign(dyn, ninst, x1, x4, x5, x2, x3, 0, 0);
+                    }
                 }
                 break;
 
--- a/src/dynarec/la64/dynarec_la64_db.c
+++ b/src/dynarec/la64/dynarec_la64_db.c
@@ -123,22 +123,39 @@ uintptr_t dynarec64_DB(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 SETFLAGS(X_ALL, SF_SET_NODF, NAT_FLAGS_NOFUSION);
                 v1 = x87_get_st(dyn, ninst, x1, x2, 0, X87_COMBINE(0, nextop & 7));
                 v2 = x87_get_st(dyn, ninst, x1, x2, nextop & 7, X87_COMBINE(0, nextop & 7));
+                if (!BOX64ENV(dynarec_fastround)) {
+                    MOVFR2GR_D(x6, v1);
+                    MOVFR2GR_D(x7, v2);
+                    x87_fcomi_fpu_ld(dyn, ninst, x1, x2, x3, x4, x5, x6, x7, nextop & 7);
+                    BNEZ_MARK(x1);
+                }
                 if (ST_IS_F(0)) {
                     FCOMIS(v1, v2, x1, x2);
                 } else {
                     FCOMID(v1, v2, x1, x2);
                 }
-
+                if (!BOX64ENV(dynarec_fastround)) {
+                    MARK;
+                }
                 break;
             case 0xF0 ... 0xF7:
                 INST_NAME("FCOMI ST0, STx");
                 SETFLAGS(X_ALL, SF_SET_NODF, NAT_FLAGS_NOFUSION);
                 v1 = x87_get_st(dyn, ninst, x1, x2, 0, X87_COMBINE(0, nextop & 7));
                 v2 = x87_get_st(dyn, ninst, x1, x2, nextop & 7, X87_COMBINE(0, nextop & 7));
+                if (!BOX64ENV(dynarec_fastround)) {
+                    MOVFR2GR_D(x6, v1);
+                    MOVFR2GR_D(x7, v2);
+                    x87_fcomi_fpu_ld(dyn, ninst, x1, x2, x3, x4, x5, x6, x7, nextop & 7);
+                    BNEZ_MARK(x1);
+                }
                 if (ST_IS_F(0)) {
                     FCOMIS(v1, v2, x1, x2);
                 } else {
                     FCOMID(v1, v2, x1, x2);
+                }
+                if (!BOX64ENV(dynarec_fastround)) {
+                    MARK;
                 }
                 break;
             default:
--- a/src/dynarec/la64/dynarec_la64_df.c
+++ b/src/dynarec/la64/dynarec_la64_df.c
@@ -91,10 +91,19 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 SET_DFNONE();
                 v1 = x87_get_st(dyn, ninst, x1, x2, 0, X87_COMBINE(0, nextop & 7));
                 v2 = x87_get_st(dyn, ninst, x1, x2, nextop & 7, X87_COMBINE(0, nextop & 7));
+                if (!BOX64ENV(dynarec_fastround)) {
+                    MOVFR2GR_D(x6, v1);
+                    MOVFR2GR_D(x7, v2);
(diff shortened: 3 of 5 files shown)


src/dynarec/la64/dynarec_la64_d9.c
```diff
@@ -91,17 +92,25 @@ uintptr_t dynarec64_D9(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+                    MOVFR2GR_D(x3, v1);        // new ST0.q
+                    x87_update_fpu_ld_sign(dyn, ninst, x1, x4, x5, x2, x3, 0, 1);
                 }
                 break;
             case 0xE1:
                 INST_NAME("FABS");
                 v1 = x87_get_st(dyn, ninst, x1, x2, 0, X87_ST0);
                 if (ST_IS_F(0)) {
                     FABS_S(v1, v1);
+                    x87_clear_fpu_ld(dyn, ninst, x1, x2, 0);
```
ptitSeb: Are you sure you need the `clear_fpu_ld`? because the valid/invalid fpu_ld should work automaticaly here (like everywhere else actualy)

src/dynarec/la64/dynarec_la64_db.c
```diff
@@ -123,23 +123,32 @@ uintptr_t dynarec64_DB(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 SETFLAGS(X_ALL, SF_SET_NODF, NAT_FLAGS_NOFUSION);
                 v1 = x87_get_st(dyn, ninst, x1, x2, 0, X87_COMBINE(0, nextop & 7));
                 v2 = x87_get_st(dyn, ninst, x1, x2, nextop & 7, X87_COMBINE(0, nextop & 7));
+                MOVFR2GR_D(x6, v1);
+                MOVFR2GR_D(x7, v2);
+                x87_fcomi_fpu_ld(dyn, ninst, x1, x2, x3, x4, x5, x6, x7, nextop & 7);
```
ptitSeb: can we have this test optionnal, only hapenning for `BOX64_DYNAREC_FASTROUND=0` for example?

src/dynarec/la64/dynarec_la64_db.c
```diff
@@ -123,23 +123,32 @@ uintptr_t dynarec64_DB(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
...
+                MARK;
                 break;
             case 0xF0 ... 0xF7:
                 INST_NAME("FCOMI ST0, STx");
                 SETFLAGS(X_ALL, SF_SET_NODF, NAT_FLAGS_NOFUSION);
                 v1 = x87_get_st(dyn, ninst, x1, x2, 0, X87_COMBINE(0, nextop & 7));
                 v2 = x87_get_st(dyn, ninst, x1, x2, nextop & 7, X87_COMBINE(0, nextop & 7));
+                MOVFR2GR_D(x6, v1);
+                MOVFR2GR_D(x7, v2);
+                x87_fcomi_fpu_ld(dyn, ninst, x1, x2, x3, x4, x5, x6, x7, nextop & 7);
```
ptitSeb: same question here abotut having this optionnal?

src/dynarec/la64/dynarec_la64_df.c
```diff
@@ -91,11 +91,16 @@ uintptr_t dynarec64_DF(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 SET_DFNONE();
                 v1 = x87_get_st(dyn, ninst, x1, x2, 0, X87_COMBINE(0, nextop & 7));
                 v2 = x87_get_st(dyn, ninst, x1, x2, nextop & 7, X87_COMBINE(0, nextop & 7));
+                MOVFR2GR_D(x6, v1);
+                MOVFR2GR_D(x7, v2);
+                x87_fcomi_fpu_ld(dyn, ninst, x1, x2, x3, x4, x5, x6, x7, nextop & 7);
```
ptitSeb: same question

ptitSeb:
@ksco what do you think about those x87 tests: should they be optionnal or is it ok to have them all the time?

ksco:
Absolutely optional. I need to find some time to review this PR, but not now, for sure.

yzewei:
The change should take effect when FASTROUND=0. As for x87_clear_fpu_ld, the single branch is not responsible for maintaining ld, so it should indeed be removed.

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -878,6 +914,313 @@ void x87_free(dynarec_la64_t* dyn, int ninst, int s1, int s2, int s3, int st)
...
+    MAYUSE(ninst);
+    MAYUSE(addr);
+    MAYUSE(tmp);
+    MAYUSE(tmp2);
+    MAYUSE(oldq);
+    MAYUSE(newq);
+    MAYUSE(st);
+    MAYUSE(chs);
+
+#ifndef HAVE_LD80BITS
```
ksco: Why?

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -878,6 +914,313 @@ void x87_free(dynarec_la64_t* dyn, int ninst, int s1, int s2, int s3, int st)
...
+    MAYUSE(dyn);
+    MAYUSE(ninst);
+    MAYUSE(s1);
+    MAYUSE(s2);
+    MAYUSE(s3);
+    MAYUSE(s4);
+    MAYUSE(a);
+    MAYUSE(b);
+
+#ifndef HAVE_LD80BITS
```
ksco: Why?

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -27,6 +28,41 @@
...
+{
+    MAYUSE(ninst);
+
+    // addr = &emu->fpu_ld[(emu->top + st - dyn->lsx.x87stack) & 7]
+    LD_WU(addr, xEmu, offsetof(x64emu_t, top));
+    if (st - dyn->lsx.x87stack)
+        ADDI_D(addr, addr, st - dyn->lsx.x87stack);
+    ANDI(addr, addr, 7);
+
+    MOV32w(tmp, sizeof(fpu_ld_t));
```
ksco: This is a compile-time constant (64). Do we need to use `MOV32` and `MUL_D`?

src/dynarec/la64/dynarec_la64_helper.c
```diff
@@ -878,6 +914,313 @@ void x87_free(dynarec_la64_t* dyn, int ninst, int s1, int s2, int s3, int st)
...
+    MAYUSE(b);
+
+#ifndef HAVE_LD80BITS
+    if (a == b)
+        return;
+
+    ld80_get_addr(dyn, ninst, s1, s4, a);
+    ld80_get_addr(dyn, ninst, s2, s4, b);
+
+    for (int i = 0; i < sizeof(fpu_ld_t); i += 2) {
```
ksco, line 1212: I don't understand, why the step here is 2?

src/dynarec/la64/dynarec_la64_d9.c
```diff
@@ -66,6 +66,7 @@ uintptr_t dynarec64_D9(dynarec_la64_t* dyn, uintptr_t addr, uintptr_t ip, int ni
                 // swap the cache value, not the double value itself :p
                 x87_get_st(dyn, ninst, x1, x2, nextop & 7, X87_ST(nextop & 7));
                 x87_get_st(dyn, ninst, x1, x2, 0, X87_ST0);
+                x87_swap_fpu_ld(dyn, ninst, x1, x2, x3, x4, 0, nextop & 7);
```
ksco: Wouldn't this need to be optional too?

ksco (changes requested):
I haven't finished the review, but this already confuses me. Could you please explain how this works in detail and what it is for? The NumPy example did not convince me, because it's a simple example that just shows that Box64 does not support long double, which is known. Does adding the exact support for this crafted example help any real-world program?

(This question holds for the interpreter change #4107 too, since that was already merged, maybe @ptitSeb has opinions too?)

ptitSeb:
I wont merge this PR for current release anyway. This will wait for the next dev. cycle.

yzewei:
> I haven't finished the review, but this already confuses me. Could you please explain how this works in detail and what it is for? The NumPy example did not convince me, because it's a simple example that just shows that Box64 does not support long double, which is known. Does adding the exact support for this crafted example help any real-world program?
> 
> (This question holds for the interpreter change #4107 too, since that was already merged, maybe @ptitSeb has opinions too?)

The NumPy case does not require full long double arithmetic support, and this PR is not trying to add that.

The concrete issue is that, on targets without HAVE_LD80BITS, Box64 already keeps an emulated raw x87 80-bit shadow for some x87/long-double state. In the LA64 dynarec path, some x87 stack/sign operations update the cached ST double value, but the raw 80-bit shadow can remain stale. Later x87 compare/classification instructions may then observe the stale raw shadow.

This is visible in real NumPy APIs, not only in a crafted unit test. For example, longdouble classification/comparison paths such as np.isinf, np.isfinite, np.isnan, and longdouble division corner cases expect inf/NaN classification to be coherent.

So far, NumPy is the only real-world workload where I have verified this directly. But the underlying issue is not NumPy-specific: it is a consistency problem between the LA64 dynarec cached ST value and Box64's existing emulated raw x87 80-bit shadow.

I can update the PR description to make this scope clearer: it does not implement full long double support; it only keeps the existing raw x87 shadow coherent for the affected dynarec x87 operations.

src/dynarec/la64/dynarec_la64_helper.c
yzewei: These helper functions update the simulated raw 80-bit shadow layout when HAVE_LD80BITS is unavailable, for example:

fpu_ld_t.ld.l.lower / upper

fpu_ld_t.uref

When HAVE_LD80BITS is available, fpu_ld_t stores a raw long double type, and the shadow layout does not need to be manually updated.

src/dynarec/la64/dynarec_la64_helper.c
yzewei: I'm concerned that fpu_ld_t is a packed struct of #pragma pack(push, 1), and the uint64_t field inside is not guaranteed to be 8-byte aligned. Using half-word load/store operations can avoid unaligned 64-bit accesses and is also less expensive than copying byte by byte.

src/dynarec/la64/dynarec_la64_d9.c
yzewei: my bad.

src/dynarec/la64/dynarec_la64_helper.c
yzewei: SLLI_D(addr, addr, 6);

src/dynarec/la64/dynarec_la64_helper.c
ksco: It makes no sense to enable HAVE_LD80BITS on a LoongArch build.

yzewei: > It makes no sense to enable HAVE_LD80BITS on a LoongArch build.

=.=  That's true.

PR #4144, opened 2026-07-28
[SYSCALL] Add various simple syscalls

LukeShortCloud:


LukeShortCloud:
I've been using this website as a nice reference for all system calls on x86_64 and aarch64:

https://www.chromium.org/chromium-os/developer-library/reference/linux-constants/syscalls/

When comparing unimplemented system calls in Box64, I first skipped ones I couldn't reliably test (for example, `shutdown`) and some calls that are unlikely to be used by a real-world app (for example, `swapon`).

AI disclaimer: I used Claude Ops 5 to further refine down the list and to also create tests to verify that more complex wrapping wasn't needed.

[syscall_passthrough_test.py](https://github.com/user-attachments/files/30466341/syscall_passthrough_test.py)

ptitSeb:
LGTM, thanks!

PR #4154, opened 2026-07-30
[ARM64_DYNAREC] Reduce default safe-flag overhead with native flag carry

jserv:
## Summary

Two ARM64 dynarec changes that cut the flag-handling overhead box64 pays under the default safe-flags policy (`BOX64_DYNAREC_SAFEFLAGS=1`), without changing correctness for the common case.

1. TEST/CMP emission cleanup. Route `TEST Eb,imm8` (0xF6 /0) through `emit_test8c`, store `{x,w}ZR` into the deferred `op2` of `emit_cmp{32,16,8}_0` instead of a zeroed scratch, and add ZF/PEND-only fast paths to `emit_test{16,16c,8c}` that use unshifted `ANDS`/`ANDSw_mask`. Static-only cleanup, runtime unchanged.

2. Native flag carry across block boundaries. A flag producer that ends a block has `X_PEND` forced onto it, so the deferred `op1/op2/res` spill keeps full EFLAGS reconstructable for any successor or signal. When the producer's whole real flag need (`gen_flags & X_ALL`, restricted to ZF/SF/CF/OF) reaches its consumers purely through native ARM NZCV, as in the common loop-control `CMP` feeding a native `Jcc`, that spill is redundant for the branch. A guarded pass in `updateNativeFlags` (reusing `getNativeFlagsUsed` for the coverage proof) clears `X_PEND` for those producers, so `emit_cmp`/`emit_test` take the `SET_DFNONE` path. Controlled by a new `BOX64_DYNAREC_LOOPFLAGS` env var.

## Result

On [Ampere eMag 8180](https://connect-admin.amperecomputing.com/api/secure-file-download/download-regular?file=new-private-files/94/tech/eMAG8180_PB_v0.55_20200211.pdf&type=technical-document&doc_id=415), `benchfloat-O2` LINPACK, default safe flags, n=15 alternating, core-pinned, `BOX64_DYNACACHE=0`:

| LOOPFLAGS | median KFLOPS | MAD | range |
| :--- | ---: | ---: | :--- |
| 0 (off) | 637,780 | 230 | 636,542 .. 638,108 |
| 1 (on) | 766,577 | 389 | 761,139 .. 767,342 |

**+20.2%**, the two distributions fully disjoint. The optimized figure matches `SAFEFLAGS=0` (about 763k), so it recovers the entire default safe-flag gap on this workload.

The hot `daxpy_ur` loop-control `CMP` drops from 6 emitted ARM ops (`STR op1/op2/df/res` + `SUBS`) to 1 (`SUBS`); the `JNZ` still branches on native NZCV.

The TEST/CMP cleanup on its own is a static win only: `benchfloat-O2` native instruction count 85,587 to 85,070 (517 fewer, about 68% of the 755 instruction safe vs `SAFEFLAGS=0` gap), with no runtime change (the removed ops are cheap MOV/shift ALU the out-of-order core hides).

## Correctness

All under default safe flags, on the same eMag host:

* `ctest` 34/34, including `longjumpInSignals`
* `test05-O2` output bit-identical to baseline
* self-checking flag test passes under SAFEFLAGS 0/1/2 and against the `DYNAREC=0` interpreter
* dynamic CMP/Jcc cosim (`BOX64_DYNAREC_TEST`): zero divergence, correct output
* signal-time EFLAGS test: fault-time flags are byte-identical to baseline on every case

`ZF/SF/CF/OF` stay reconstructable at faults through `adjust_arch`'s NZCV overlay (kept in `need_nat_flags`). Only PF/AF turn approximate across the block exit, which the dynarec already is for native flags (only the interpreter is exact at async fault points), and which the x86 ABI (arithmetic flags not preserved across calls) makes safe. A mid-block native `CMP+Jcc` already emits no deferred spill (`gen_flags=X_ZF`, no `X_PEND`); this brings the same treatment to the block-boundary case.

## Notes

`BOX64_DYNAREC_LOOPFLAGS` defaults to `0`. The safety argument for clearing the boundary spill rests on x86 ABI flag-death across calls, so it is kept opt-in pending broader real-application and Wine validation.

### DIFF
--- a/src/dynarec/arm64/dynarec_arm64_00.c
+++ b/src/dynarec/arm64/dynarec_arm64_00.c
@@ -4068,8 +4068,7 @@ uintptr_t dynarec64_00(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip, int nin
                     SETFLAGS(X_ALL, SF_SET_PENDING);
                     GETEB(x1, 1);
                     u8 = F8;
-                    MOV32w(x2, u8);
-                    emit_test8(dyn, ninst, x1, x2, x3, x4, x5);
+                    emit_test8c(dyn, ninst, x1, u8, x2, x4, x5);
                     break;
                 case 2:
                     INST_NAME("NOT Eb");
--- a/src/dynarec/arm64/dynarec_arm64_emit_tests.c
+++ b/src/dynarec/arm64/dynarec_arm64_emit_tests.c
@@ -80,9 +80,8 @@ void emit_cmp32(dynarec_arm_t* dyn, int ninst, rex_t rex, int s1, int s2, int s3
 void emit_cmp32_0(dynarec_arm_t* dyn, int ninst, rex_t rex, int s1, int s3, int s4)
 {
     IFX_PENDOR0 {
-        MOV64xw(s4, 0);
         STRxw_U12(s1, xEmu, offsetof(x64emu_t, op1));
-        STRxw_U12(s4, xEmu, offsetof(x64emu_t, op2));
+        STRxw_U12(xZR, xEmu, offsetof(x64emu_t, op2));
         STRxw_U12(s1, xEmu, offsetof(x64emu_t, res));
         SET_DF(s4, rex.w?d_cmp64:d_cmp32);
     } else {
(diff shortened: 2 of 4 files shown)


ptitSeb:
So, can you split this PR in 2 (or 3 part), with the simple stuffs separated from the LOOP thingy (and the xZR stuff alone maybe, as this would be a 0 risk PR).

About the LOOP thing, I suppose you are using some form of Coding Agent? Because every explanation is ultra verbose and yet not really clear (sorry). If I understand correctly, the idea here is to ignore AF &PF flags when most other flags are already computed, as native, and X_PEND is also required: the idea is then to remove X_PEND and just ignore AF & PF to get faster loop (altho no loop check is involved here) by not storing Defered Flags informations. It's an interesting idea, but I want to point out that, for some opcode like COMI, PF is actually computed using native VF flags. Also, the name LOOP while it's not testing about loop is a bit surprising. Also also, there are documentation to change when you add a new env. var (`docs/gen/usage.json` then use `gen.py` to re-generates documentation)

Last point: I'm working a creating a new release of box64 (v0.4.4), so this PR will have to wait for the next developement cycle anyway.

PR #4158, opened 2026-07-31
[ELFLOADER] Use host page size instead of p_align for mapping ranges

ksco:
`p_align` only requires `p_vaddr` and `p_offset` to be congruent modulo `p_align`; segment starts need not be `p_align`-aligned. A `p_align` of 16 KiB on a host with 4 KiB pages is therefore perfectly legal. The loader, however, used `p_align` as the page mask when computing mapping and protection ranges, which makes the writable segment's range extended into pages belonging to executable code in my case.

### DIFF
--- a/src/elfs/elfloader.c
+++ b/src/elfs/elfloader.c
@@ -359,7 +359,7 @@ int AllocLoadElfMemory(box64context_t* context, elfheader_t* head, int mainbin)
                 uintptr_t paddr = head->multiblocks[n].paddr&~(box64_pagesize - 1);
                 size_t asize = head->multiblocks[n].asize+(head->multiblocks[n].paddr-paddr);
                 void* p = MAP_FAILED;
-                if(paddr==(paddr&~(box64_pagesize-1)) && (asize==ALIGN(asize))) {
+                if(asize == ALIGN(asize)) {
                     printf_dump(log_level, "Allocating 0x%zx (0x%zx) bytes @%p, will read 0x%zx @%p for Elf \"%s\"\n", asize, e->p_memsz, (void*)paddr, e->p_filesz, (void*)head->multiblocks[n].paddr, head->name);
                     p = InternalMmap(
                         (void*)paddr,
--- a/src/elfs/elfloader.c
+++ b/src/elfs/elfloader.c
@@ -322,12 +322,9 @@ int AllocLoadElfMemory(box64context_t* context, elfheader_t* head, int mainbin)
             head->multiblocks[n].align = e->p_align;
             // HACK: Mark all the code pages writable in unittest mode because some tests mix code and (writable) data...
             uint8_t prot = ((e->p_flags & PF_R)?PROT_READ:0)|(((e->p_flags & PF_W) || box64_unittest_mode)?PROT_WRITE:0)|((e->p_flags & PF_X)?PROT_EXEC:0);
-            // check if alignment is correct
-            uintptr_t balign = head->multiblocks[n].align-1;
-            if (balign < (box64_pagesize - 1)) balign = box64_pagesize - 1;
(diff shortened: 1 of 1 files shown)


src/elfs/elfloader.c
```diff
@@ -322,12 +322,9 @@ int AllocLoadElfMemory(box64context_t* context, elfheader_t* head, int mainbin)
...
             // HACK: Mark all the code pages writable in unittest mode because some tests mix code and (writable) data...
             uint8_t prot = ((e->p_flags & PF_R)?PROT_READ:0)|(((e->p_flags & PF_W) || box64_unittest_mode)?PROT_WRITE:0)|((e->p_flags & PF_X)?PROT_EXEC:0);
-            // check if alignment is correct
-            uintptr_t balign = head->multiblocks[n].align-1;
-            if (balign < (box64_pagesize - 1)) balign = box64_pagesize - 1;
-            head->multiblocks[n].asize = (e->p_memsz + (e->p_paddr & balign) + (box64_pagesize - 1)) & ~(box64_pagesize - 1);
+            head->multiblocks[n].asize = (e->p_memsz + (e->p_paddr & (box64_pagesize - 1)) + (box64_pagesize - 1)) & ~(box64_pagesize - 1);
             int try_mmap = 1;
-            if(e->p_paddr&balign)
+            if(e->p_paddr&(box64_pagesize - 1))
```
ptitSeb, line 327: I think the balign test should also stay in place

ksco: What?

ksco: You mean `if((e->p_paddr&(box64_pagesize - 1)) && (e->p_addr&blign))`?

src/elfs/elfloader.c
```diff
@@ -359,7 +356,7 @@ int AllocLoadElfMemory(box64context_t* context, elfheader_t* head, int mainbin)
                 }
             }
             if(!try_mmap) {
-                uintptr_t paddr = head->multiblocks[n].paddr&~balign;
+                uintptr_t paddr = head->multiblocks[n].paddr&~(box64_pagesize - 1);
```
ptitSeb, line 359: Why? there is a test 2 line 365 to check if the paddr is pagesize align, and use an alternate method if it's not. Doing like that will just break elf that rely on small page alignement.

src/elfs/elfloader.c
ptitSeb: Well, mmm, I think it's an `||` instead?

ksco: That's just redundant --- `balign = max(p_align-1, pagesize-1)`, so for `p_align >= pagesize` the low page bits are already covered.

ptitSeb: but then, why using pagesize instead of balign? testing alignement against balign will be more restrictive than pagesize?

src/elfs/elfloader.c
ksco: Okay, I have removed the useless guard at line 365.

src/elfs/elfloader.c
ksco: Check out the PR description; balign is only a modulo for p_vaddr and p_offset, it should never be used as alignment.

src/elfs/elfloader.c
ptitSeb: also, balign might be bigger than pagesize, even on 16K pagesize.

ksco: Yes, that's why we should not use it.

ptitSeb:
I guess similar changes would be needed in `AllocLoadElfMemory32` too

ksco:
Yes, I will do it.

ksco:
In another PR.

PR #4165, opened 2026-08-01
[RV64_DYNAREC] Refined many fastnan=0 handling

ksco:


src/dynarec/rv64/dynarec_rv64_f30f.c
```diff
@@ -208,18 +208,58 @@ uintptr_t dynarec64_F30F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip, int
...
         case 0x52:
             INST_NAME("RSQRTSS Gx, Ex");
             nextop = F8;
             GETEXSS(v0, 0);
             GETGXSS_empty(v1);
-            q0 = fpu_get_scratch(dyn);
+            q0 = fpu_get_scratch(dyn); // 1.0f
+            d0 = fpu_get_scratch(dyn); // 0.0f
             LUI(x3, 0x3F800); // 1.0f
             FMVWX(q0, x3);
```
ptitSeb, line 237: don't you need to initialize `d0` to 0.0f with a XOR or something like that?

ptitSeb: Ah no, it's done in the next if

PR #4166, opened 2026-08-01
[INTERP] Refined canonical NaN handling for RISC-V

ksco:


ptitSeb:
This feels strange to me. I don't think most platform needs this special case (like Arm64 or Loongarch), only RISC-V needs it, and yet, there is no `#ifdef` around the new code. Some code looks like approximation to me (like `|0x0040000` wich seems arbitrary) and I would prefer it to be risc-v only as other plantforms handles NaN similar to x86 iirc.

ksco:
It won't change the behaviour of other platforms, just make them slower. But sure, I can add RISC-V only guards.

ksco:
Done.

PR #4174, opened 2026-08-02
[SIGNAL] Preserve TF handling when resuming signals with Dynarec

Mikura-Kyouka:
When a signal handler restores a context with TF set, resume through the interpreter for one instruction so no_tf can suppress the initial spurious SIGTRAP.

This fixes the issue exposed after commit 38d64ba when running Syringe.exe from the Ares expansion platform.

### DIFF
--- a/src/libtools/signal32.c
+++ b/src/libtools/signal32.c
@@ -745,7 +745,9 @@ int my_sigactionhandler_oldcode_32(x64emu_t* emu, int32_t sig, int simple, sigin
             #undef GO
             // flags
             emu->eflags.x64=sigcontext->uc_mcontext.gregs[I386_EFL];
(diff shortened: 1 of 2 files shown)


ksco:
Why not just revert 38d64ba?

ksco:
Related PR: https://github.com/ptitSeb/box64/pull/3951

Basically, commit 38d64ba broke #3951.

ksco (approved):
PR looks legit to me, but maybe we should just revert the offending commit.

ptitSeb:
Well, commit 38d64baba2e3806dcae8a25949e1574027e4b20c should bring some slight speedup for regular signal handling, especialy for DRM that use a lot of segfault, so I prefer to use that PR instead of reversing the other.

PR #4178, opened 2026-08-03
[LA64_DYNAREC] Speed up crc validation

ksco:


ksco:
This is only slightly slower than the exact match method, which is good enough.

ptitSeb:
Well, the exact match will always be faster in case of mismatch...
But how this new one is faster? is it just loop unrolling?

ksco:
No, it's more than that. Every iteration of the original 1-way CRC loop has a data dependency on the previous iteration, and CRC is quite slow. A multi-way CRC loop will cancel most of those dependencies.

ptitSeb:
Oh, I see. The point is to avoid starving the pipeline with dependant load+crc pairs

ksco:
For 12 elements, it's like

```
1 -> 2 -> 3 -> 4 -> 5 -> 6 -> 7 -> 8 -> 9 -> 10 -> 11 -> 12
```

v.s.

```
1 -> 7
  2 -> 8
    3 -> 9
      4 -> 10
        5 -> 11
          6 -> 12
```

ksco:
I tested 4-way, 6-way, 8-way, 6 way seems to be the best, but I guess that depends on the hardware.

PR #4183, opened 2026-08-04
[PAGESIZE] Emulated 4k pagesize for better compatibility

ksco:


ptitSeb:
why not having this under a setting? there are many program that runs fine with larger pagesize, but might break with the emulation? or are you confident the emulation will just work and having it configurable will just make everything more complex for no reason?

ksco:
Well, I do doubt this will break any already working applications, considering x86 Linux is 4k pagesize only.

ksco:
It also does not make sense to return 16384 while our ELF loader tries its best to emulate a 4K "experience".

ptitSeb:
But box64 is not the only program that read system pagesize to adapt to it at runtime. Pretty sure Linux Unity3D games does that for example.

ksco:
> But box64 is not the only program that read system pagesize to adapt to it at runtime. Pretty sure Linux Unity3D games does that for example.

So you're worried that this will cancel out some of the good adaptions by the guest program?

ptitSeb:
Yeah, exactly

ksco:
Okay, but I don't know, adding an option for this sounds over-worried to me, unless there is such a case. x86 Linux is 4k pagesize only, so it always makes sense to return 4096.

PR #4185, opened 2026-08-04
[WRAPPER] Wrapped madvise for non4k pagesize systems

ksco:
For now, only MADV_DONTNEED is handled by fallbacking to a plain memset on unaligned cases.


src/emu/x64syscall.c
```diff
@@ -662,6 +663,11 @@ void EXPORT x64Syscall_linux(x64emu_t *emu)
             if(S_RAX==-1)
                 S_RAX = -errno;
             break;
+        case 28: // sys_madvise
```
ptitSeb: Why did you put 28 between 11 & 13???

PR #4187, opened 2026-08-04
[MMAP] Fixed wine mappings on 16K hosts

ksco:
Wine works!


ksco:
image
For the record.

PR #4190, opened 2026-08-05
[RV64_DYNAREC] Fix RDTSC and RDTSCP

zengdage:
Fix typo. Left shifts shall be used here to simulate high frequency from a low-frequency counter.


ksco (approved):
Oh did I forgot to port this fix to RV

PR #4192, opened 2026-08-05
[RV64_DYNAREC] Add qNaN handle for ROUNDPD

zengdage:


src/dynarec/rv64/dynarec_rv64_660f38.c
```diff
@@ -808,7 +808,15 @@ uintptr_t dynarec64_660F38(dynarec_rv64_t* dyn, uintptr_t addr, uint8_t opcode,
                     // i = 0
                     FLD(d0, wback, fixedaddress);
                     FEQD(x4, d0, d0);
-                    BNEZ(x4, 8);
+                    BNEZ(x4, 4 + 6 * 4);
```
ptitSeb: Are you sure about this fixed-sized jump while the following code is conditionnal and not always emitted?

ptitSeb: I would expect to see something like
```C
BNEZ(x4, BOX64ENV(dynarec_fastnan)?8:(4+6*4));
```
instead

zengdage: Sorry, this is a error. Now fixed.

ptitSeb (approved):

PR #4210, opened 2026-08-07
[DYNACACHE] Checked CRC when loading dynablocks from cache

ksco:


ksco:
I need to think about this.

ptitSeb:
Yeah, that should not be needed: loaded blocks are setup as dirty, so the test will be done on-demand.

ksco:
Yeah

ksco:
@ptitSeb Okay, now I finally understand this: cached native blocks can contain relocated direct links to other cached blocks, and such a link does not trigger a CRC test!

ksco:
Nah, that still goes into the jmpnext.

ptitSeb:
Yeah "direct" link still go through the jmptable, just with a direct link...

PR #4211, opened 2026-08-07
[RCFILE] Added dota2 profile

ksco:


ksco:
Screenshot_20260807_151353

Tested on 3B6000 16k pagesize kernel.

Issue #4214, opened 2026-08-07
Claude Code Review Report


qarmin:
I ran Claude Code to look for potential issues, including logic bugs, copy-paste errors, architecture problems, panics, and similar issues.

I don't have much experience with low-level C and assembly, so I wasn't able to properly verify most of these findings myself. However, I reported a similar set of findings to the FEX project, and the author fixed most of the issues I reported(https://github.com/FEX-Emu/FEX/issues/5678).

The full report - [box64_20260807.html](https://github.com/user-attachments/files/30839754/box64_20260807.html)

Here are a few example findings that I'm almost certain are actual bugs:

### CPY_4 `CRITICAL`

**Description:** `my32_XFreeFontSet` frees `set->names[i]` in **both** cleanup loops; the second loop iterates over
`set->fonts_size` but still frees from the `set->names` array. When a program called both
`XFontsOfFontSet(..., fonts, names)` variants, the same `set->names[i]` block is passed to
`box32_free()` twice (double free / heap corruption), the `set->fonts[i]` blocks are leaked, and if
`fonts_size > names_size` the loop reads past the end of `set->names`.

**Locations:**

https://github.com/ptitSeb/box64/blob/2f130fab1d6e1a4ee8a71dc60cfdfcc839ad192a/src/wrapped32/wrappedlibx11.c#L2569-L2579

**Fix:**
 ```c
     for(int i=0; i<set->fonts_size; ++i)
-        box32_free(set->names[i]);
+        box32_free(set->fonts[i]);
     box32_free(set->fonts);
```

### CPY_49 `HIGH`

**Description:** `my32_glBindBuffersRange` converts both the `offsets` and the `sizes` arrays from 32-bit `long_t` to native
64-bit `long`, but then passes the **unconverted** guest array `sizes` to the native entry point instead of
`sizes_l`. The native driver therefore reads `count` 64-bit `GLsizeiptr` values out of a buffer that only holds
`count` 32-bit values: every size is wrong (two adjacent 32-bit sizes merged into one 64-bit value) and the read
runs `4 * count` bytes past the end of the guest array. The very next functions in the same file
(`my32_glBindVertexBuffers`, `my32_glVertexArrayVertexBuffers`) pass their converted array correctly, which
makes this an obvious editing slip. The bug is duplicated verbatim in `wrappedlibglxnvidia.c`.

**Locations:**

[...]

PR #4215, opened 2026-08-07
Addressed some wrapping issues report in #4214

ksco:


ptitSeb:
That's a lot of small errors fixed there!

PR #4220, opened 2026-08-09
[ARM64_DYNAREC] Addressed more issues reported in #4214

ksco:


src/dynarec/arm64/dynarec_arm64_avx_66_0f.c
```diff
@@ -1464,8 +1464,9 @@ uintptr_t dynarec64_AVX_66_0F(dynarec_arm_t* dyn, uintptr_t addr, uintptr_t ip,
             MOVI_32(q1, 64);
             UQXTN_32(q0, v1);
             UMIN_32(q0, q0, q1);    // limit to 0 .. +64 values
-            NEG_32(q0, q0);         // neg to do shr
-            VDUPQ_32(q0, q0, 0);    // only the low 8bits will be used anyway
+            UXTL_32(q0, q0);        // zero-extend 32->64 per lane
+            VDUPQ_64(q0, q0, 0);    // x86 uses the scalar count from bits[63:0] for both lanes
```
ptitSeb: While this operation is also exect, it's 1 opcode longer than the original one, that was  also exact, because, as the comment mentionned, the `USHLQ_64` shift only takes the lower 8bits of the shift register and sign extend it to get the element shift. So the expsion to 64bits is actually not mendatory, the 8-63 bits will be ignored anyway.

ksco: I see, should I revert this?

ptitSeb: Well, I think prefer the previous one yes, as it's slightly faster, even if the comment is a bit dry

ptitSeb:
so many fixes

PR #4227, opened 2026-08-11
[LA64_DYNAREC] Inlined a few common used C fuctions

ksco:
50% faster in dhrystone benchmark.


ksco:
I can't debug this for now, so I'll just remove the test.

ptitSeb:
wow, the return of HLE ;). That might make generated code much bigger, as it's inlined... can't something fixed, like UpdateFlags, still fast but would not inflate the code size that much?

ksco:
> wow, the return of HLE ;). That might make generated code much bigger, as it's inlined... can't something fixed, like UpdateFlags, still fast but would not inflate the code size that much?

Indeed.

src/dynarec/la64/dynarec_la64_native.c
```diff
@@ -0,0 +1,141 @@
...
+    if (native_call_blocks[call])
+        return native_call_blocks[call]->block;
+
+    dynarec_la64_t helper = { 0 };
+    instruction_la64_t insts[1] = { 0 };
+    helper.insts = insts;
+    helper.need_dump = BOX64ENV(dynarec_dump) == 3 ? 0 : BOX64ENV(dynarec_dump);
+    helper.cap = 1;
+
+    la64_native_call_pass2(&helper, call);
```
ptitSeb, line 68: no pass0 & pass1 ?

ptitSeb: Ah, I guess for this code pass0 & pass1 that parse the x86 code are doing nothing, as there is no x86 code to parse.

PR #4228, opened 2026-08-11
Fix some UAF errors report in #4214

zengdage:
Both `my_gpgme_data_new_from_cbs` and `my_ber_sockbuf_add_io` are supplied with stack-allocated variables. These variables go out of live scope once the function returns and must not be accessed afterwards.

Set it in:

1. https://github.com/gpg/gpgme/blob/master/src/data-user.c#L99 
2. https://github.com/delphij/openldap/blob/master/libraries/liblber/sockbuf.c#L201

### DIFF
--- a/src/wrapped/wrappedlber.c
+++ b/src/wrapped/wrappedlber.c
@@ -178,21 +178,54 @@ static void* find_sbi_close_Fct(void* fct)
 
 #undef SUPER
 
+// sockbuf_io
+#define SUPER() \
+GO(0)           \
+GO(1)           \
+GO(2)           \
+GO(3)           \
+GO(4)
+
+#define GO(A)                                           \
+static my_sockbuf_io_t* my_sockbuf_io_ref_##A = NULL;   \
+static my_sockbuf_io_t my_sockbuf_io_struct_##A = {0};
+SUPER()
+#undef GO
+
+static void wrap_sockbuf_io(my_sockbuf_io_t* dst, my_sockbuf_io_t* src)
+{
+    #define GO(A) dst->A = find_##A##_Fct(src->A)
+    GO(sbi_setup);
+    GO(sbi_remove);
+    GO(sbi_ctrl);
+    GO(sbi_read);
+    GO(sbi_write);
+    GO(sbi_close);
+    #undef GO
+}
+
+static my_sockbuf_io_t* find_sockbuf_io_Struct(my_sockbuf_io_t* io)
+{
+    if(!io) return NULL;
+    #define GO(A) if(my_sockbuf_io_ref_##A == io) return &my_sockbuf_io_struct_##A;
+    SUPER()
+    #undef GO
+    #define GO(A) if(!my_sockbuf_io_ref_##A) {              \
+        wrap_sockbuf_io(&my_sockbuf_io_struct_##A, io);     \
+        my_sockbuf_io_ref_##A = io;                         \
+        return &my_sockbuf_io_struct_##A;                   \
+    }
+    SUPER()
+    #undef GO
+    printf_log(LOG_NONE, "Warning, no more slot for liblber sockbuf_io\n");
+    return NULL;
+}
+
+#undef SUPER
+
 EXPORT int my_ber_sockbuf_add_io(x64emu_t* emu, void* sb, my_sockbuf_io_t* io, int layer, void* arg)
 {
-    my_sockbuf_io_t my_io = {0};
-    if(io) {
-        #define GO(A) my_io.A = find_##A##_Fct(io->A)
-        GO(sbi_setup);
-        GO(sbi_remove);
-        GO(sbi_ctrl);
-        GO(sbi_read);
-        GO(sbi_write);
-        GO(sbi_close);
-        #undef GO
-    }
-    return my->ber_sockbuf_add_io(sb, io?&my_io:NULL, layer, arg);
+    return my->ber_sockbuf_add_io(sb, find_sockbuf_io_Struct(io), layer, arg);
 }
 
 #include "wrappedlib_init.h"
-
--- a/src/wrapped/wrappedlber.c
+++ b/src/wrapped/wrappedlber.c
@@ -204,7 +204,7 @@ static void wrap_sockbuf_io(my_sockbuf_io_t* dst, my_sockbuf_io_t* src)
     #undef GO
 }
 
-static my_sockbuf_io_t* find_sockbuf_io_Struct(my_sockbuf_io_t* io)
+static my_sockbuf_io_t* allocate_sockbuf_io_Struct(my_sockbuf_io_t* io)
 {
     if(!io) return NULL;
     #define GO(A) if(my_sockbuf_io_ref_##A == io) return &my_sockbuf_io_struct_##A;
@@ -221,11 +221,41 @@ static my_sockbuf_io_t* find_sockbuf_io_Struct(my_sockbuf_io_t* io)
     return NULL;
 }
 
+static my_sockbuf_io_t* find_sockbuf_io_Struct(my_sockbuf_io_t* io)
+{
+    if(!io) return NULL;
+    #define GO(A) if(my_sockbuf_io_ref_##A == io) return &my_sockbuf_io_struct_##A;
+    SUPER()
+    #undef GO
+    printf_log(LOG_NONE, "Warning, No available liblber sockbuf_io slot found.\n");
+    return NULL;
+}
+
+static void reset_sockbuf_io_Struct(my_sockbuf_io_t* io)
+{
+    if(!io) return;
+    #define GO(A) if(my_sockbuf_io_ref_##A == io) {         \
+        my_sockbuf_io_ref_##A = NULL;                       \
+        memset(&my_sockbuf_io_struct_##A, 0, sizeof(*io));  \
+        return;                                             \
+    }
+    SUPER()
+    #undef GO
+    printf_log(LOG_NONE, "Warning, No available liblber sockbuf_io slot found.\n");
+}
+
 #undef SUPER
 
 EXPORT int my_ber_sockbuf_add_io(x64emu_t* emu, void* sb, my_sockbuf_io_t* io, int layer, void* arg)
 {
-    return my->ber_sockbuf_add_io(sb, find_sockbuf_io_Struct(io), layer, arg);
+    return my->ber_sockbuf_add_io(sb, allocate_sockbuf_io_Struct(io), layer, arg);
+}
+
+EXPORT int my_ber_sockbuf_remove_io(x64emu_t* emu, void* sb, my_sockbuf_io_t* io, int layer)
+{
+    int ret = my->ber_sockbuf_remove_io(sb, find_sockbuf_io_Struct(io), layer);
+    reset_sockbuf_io_Struct(io);
+    return ret;
 }
 
 #include "wrappedlib_init.h"
--- a/src/wrapped/wrappedlber.c
+++ b/src/wrapped/wrappedlber.c
@@ -188,6 +188,7 @@ GO(4)
 
 #define GO(A)                                           \
 static my_sockbuf_io_t* my_sockbuf_io_ref_##A = NULL;   \
+static uint32_t my_sockbuf_io_ref_count_##A = 0;        \
 static my_sockbuf_io_t my_sockbuf_io_struct_##A = {0};
 SUPER()
 #undef GO
@@ -204,15 +205,20 @@ static void wrap_sockbuf_io(my_sockbuf_io_t* dst, my_sockbuf_io_t* src)
     #undef GO
 }
 
-static my_sockbuf_io_t* allocate_sockbuf_io_Struct(my_sockbuf_io_t* io)
+static my_sockbuf_io_t* find_sockbuf_io_Struct(my_sockbuf_io_t* io, int alloc)
 {
     if(!io) return NULL;
-    #define GO(A) if(my_sockbuf_io_ref_##A == io) return &my_sockbuf_io_struct_##A;
+    #define GO(A) if(my_sockbuf_io_ref_##A == io) {        \
+        if (alloc) my_sockbuf_io_ref_count_##A++;          \
+        return &my_sockbuf_io_struct_##A;                  \
+    }
     SUPER()
     #undef GO
+    if (!alloc) return io;
     #define GO(A) if(!my_sockbuf_io_ref_##A) {              \
         wrap_sockbuf_io(&my_sockbuf_io_struct_##A, io);     \
         my_sockbuf_io_ref_##A = io;                         \
+        my_sockbuf_io_ref_count_##A = 1;                    \
         return &my_sockbuf_io_struct_##A;                   \
     }
     SUPER()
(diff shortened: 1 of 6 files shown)


ptitSeb:
Some builds are red.
```
/home/runner/work/box64/box64/src/libtools/threads32.c:579:16: error: assignment to 'ptr_t' {aka 'unsigned int'} from 'void *' makes integer from pointer without a cast [-Wint-conversion]
  579 |     cond->cond = NULL;
```
`cond->cond` is 32bits pointer, so basicaly, an integer. Just use `cond->cond = 0;` there (there is no equivalent of `NULL` for `ptr_t` for now)

src/wrapped/wrappedlber.c
```diff
@@ -180,9 +180,11 @@ static void* find_sbi_close_Fct(void* fct)
 
 EXPORT int my_ber_sockbuf_add_io(x64emu_t* emu, void* sb, my_sockbuf_io_t* io, int layer, void* arg)
 {
-    my_sockbuf_io_t my_io = {0};
+    my_sockbuf_io_t* my_io = NULL;
     if(io) {
-        #define GO(A) my_io.A = find_##A##_Fct(io->A)
+        // TODO: memory leak, my_io should be free in ber_sockbuf_remove_io
+        my_io = (my_sockbuf_io_t*)box_calloc(1, sizeof(my_sockbuf_io_t));
```
ptitSeb: I don't want a solution that explicitely create a memory leak. Maybe a find_XXXX based on io address would be a better solution.

src/wrapped/wrappedlibgpgme.c
```diff
@@ -134,14 +134,16 @@ static void* find_release_Fct(void* fct)
...
 {
-    my_gpgme_data_cbs_t cbs_ = {0};
+    my_gpgme_data_cbs_t *cbs_ = NULL;
     if(cbs) {
-        cbs_.read = find_read_Fct(cbs->read);
-        cbs_.write = find_write_Fct(cbs->write);
-        cbs_.seek = find_seek_Fct(cbs->seek);
-        cbs_.release = find_release_Fct(cbs->release);
+        // TODO: memory leak, cbs_ should be free in my_gpgme_data_release.
+        cbs_ = box_calloc(1, sizeof(my_gpgme_data_cbs_t));
```
ptitSeb: Were is the free of this calloc? the Claude repport mentionned the possibility of using `gpgme_data_release` for that.

src/wrapped/wrappedlber.c
zengdage: I am still unfamiliar with how ` find_xxx` works; I will look into it first.

src/wrapped/wrappedlibgpgme.c
zengdage: As can be seen from [https://github.com/gpg/gpgme/blob/master/src/data-user.c#L99](https://github.com/gpg/gpgme/blob/master/src/data-user.c#L99), the `cbs` object is stored in `gpgme_data->data.user.cbs`. Therefore, the `struct gpgme_data `structure (defined at [https://github.com/gpg/gpgme/blob/master/src/data.h#L135](https://github.com/gpg/gpgme/blob/master/src/data.h#L135)) needs to be referenced to retrieve the stored `cbs`.

I have added the `my_gpgme_data_release` to free cbs.  Please help to review it again.

src/wrapped/wrappedlibgpgme.c
```diff
@@ -134,14 +134,63 @@ static void* find_release_Fct(void* fct)
...
+        } user;
+    } data;
+};
+
+
+EXPORT void my_gpgme_data_release(x64emu_t *emu, void* data)
+{
+    struct my_gpgme_data *gpgme_data = (struct my_gpgme_data *)data;
+    if (gpgme_data) {
+        my_gpgme_data_cbs_t *cbs_ = gpgme_data->data.user.cbs;
```
ptitSeb: According to the gpgme sources, the cbs struct is the 1st argument of the structure, which, in that case, don't even need to be defined.

Image

src/wrapped/wrappedlibgpgme.c
ptitSeb: There is something wrong here. The union is a simplified version of the actual structure (which is fine), but the issue is, those user cbs/handle fields might not be valid, for a data intialized with an fd for example. And I don't see any field that is setup to allow knowing wich field of the union is valid.

src/wrapped/wrappedlibgpgme.c
zengdage: There are two cbs objects inside `gpgme_data`. Looking at the source code of `gpgme_data_new_from_cbs,` it uses the latter one.

src/wrapped/wrappedlibgpgme.c
zengdage: > There is something wrong here. The union is a simplified version of the actual structure (which is fine), but the issue is, those user cbs/handle fields might not be valid, for a data intialized with an fd for example. And I don't see any field that is setup to allow knowing wich field of the union is valid.

Getting the `cbs` via `data.user.cbs` may be an incorrect approach. I think we need to create a map between `gpgme_data` and `cbs` to get the `cbs` object to be freed.  This seems like the correct direction. But this adds some complexity. What do you think?

ptitSeb: Either a map or the static find_XXXX approach. I really depend on the occurence of this call

src/wrapped/wrappedlber.c
zengdage: I reworked `my_ber_sockbuf_add_io` with the `find_xxx `  and preallocated 5 `my_sockbuf_io_t` structs. Please help to review. Thanks.

zengdage:
Test case as follow：

```
#include "portable.h"
#include <stdio.h>
#include <ac/stdlib.h>
#include <ac/string.h>
#include "lber-int.h"

static int setup_called;

static int
test_setup( Sockbuf_IO_Desc *sbiod, void *arg )
{
        setup_called = 1;
        sbiod->sbiod_pvt = arg;
        return 0;
}

static Sockbuf_IO test_io = {
        test_setup,
        NULL,
        NULL,
        NULL,
        NULL,
        NULL
};

int main( void )
{
        Sockbuf sb;
        int arg = 42;
        int rc;
        memset( &sb, 0, sizeof( sb ) );
        sb.sb_valid = LBER_VALID_SOCKBUF;

        Sockbuf_IO ios[6];
        for(int i = 0; i < sizeof(ios)/sizeof(ios[0]); i++) {
                ios[i] = test_io;
                rc = ber_sockbuf_add_io( &sb, &ios[i], LBER_SBIOD_LEVEL_PROVIDER, &arg );
                if ( rc != 0 || !setup_called || sb.sb_iod == NULL ||
                        sb.sb_iod->sbiod_level != LBER_SBIOD_LEVEL_PROVIDER ||
                        sb.sb_iod->sbiod_pvt != &arg )
                {
                        fprintf( stderr, "ber_sockbuf_add_io test failed\n" );
                        return EXIT_FAILURE;
                }
                printf(" %s %d sb.sb_iod->sbiod_io:%p\n", __func__, __LINE__, sb.sb_iod->sbiod_io);
        }
        puts( "ber_sockbuf_add_io test passed" );
        return EXIT_SUCCESS;
}
```

test result:
```
 main 50 sb.sb_iod->sbiod_io:0x36200e88
 main 50 sb.sb_iod->sbiod_io:0x36200dc8
 main 50 sb.sb_iod->sbiod_io:0x36200df8
 main 50 sb.sb_iod->sbiod_io:0x36200e28
 main 50 sb.sb_iod->sbiod_io:0x36200e58
[BOX64] Warning, no more slot for liblber sockbuf_io
ber_sockbuf_add_io test failed
```

zengdage:
I have a question about the `find_xxx` mechanism. It does not appear to handle thread safety when accessing these static variables. What considerations were made for this in Box64?

ptitSeb:
> I have a question about the `find_xxx` mechanism. It does not appear to handle thread safety when accessing these static variables. What considerations were made for this in Box64?

Until now, functions that use the find_XXXX mecanism were very unlikely to need to be thread safe, as they are mostly use to setup stuffs and used in sequence and not in parallel.
So instead of wasting cpu cycle and ressource with a local mutex, it was just designed to use minimm ressources and simple sequencial aspect.
Using a local mutex if needed can be done very easily (with a staticaly initialized mutex), but it would need to be done on a as-needed basis.

zengdage:
> > I have a question about the `find_xxx` mechanism. It does not appear to handle thread safety when accessing these static variables. What considerations were made for this in Box64?
> 
> Until now, functions that use the find_XXXX mecanism were very unlikely to need to be thread safe, as they are mostly use to setup stuffs and used in sequence and not in parallel. So instead of wasting cpu cycle and ressource with a local mutex, it was just designed to use minimm ressources and simple sequencial aspect. Using a local mutex if needed can be done very easily (with a staticaly initialized mutex), but it would need to be done on a as-needed basis.

Got it. Thanks.

ptitSeb:
Looking at lber doc (https://man7.org/linux/man-pages/man3/lber-sockbuf.3.html for example), I see there is a `ber_sockbuf_remove_io` to goes with ̀ber_sockbuf_add_io` so I guess the IA analysis was incomplete: that function also need to be wrapped using the same find_XXX I guess, to be sure it's removed correctly?

src/wrapped/wrappedlber.c
```diff
@@ -178,21 +178,84 @@ static void* find_sbi_close_Fct(void* fct)
...
+
+EXPORT int my_ber_sockbuf_add_io(x64emu_t* emu, void* sb, my_sockbuf_io_t* io, int layer, void* arg)
+{
+    return my->ber_sockbuf_add_io(sb, allocate_sockbuf_io_Struct(io), layer, arg);
+}
+
+EXPORT int my_ber_sockbuf_remove_io(x64emu_t* emu, void* sb, my_sockbuf_io_t* io, int layer)
+{
+    int ret = my->ber_sockbuf_remove_io(sb, find_sockbuf_io_Struct(io), layer);
+    reset_sockbuf_io_Struct(io);
```
ptitSeb: You are assuming that the `io` struct has been used only 1 time with this mecanism.

zengdage: I plan to add a reference counter to my_sockbuf_io_ref_. What do you think?

ptitSeb: Yes, that would be needed here, as there is a "remove" part.

ptitSeb: but beware to increment only in add_io and not in remove_io...

zengdage: Reference count has been added, and tests pass.

Test1: Pass
```
        rc = ber_sockbuf_add_io( &sb, &test_io, LBER_SBIOD_LEVEL_PROVIDER, &arg );
        rc = ber_sockbuf_add_io( &sb, &test_io, LBER_SBIOD_LEVEL_PROVIDER, &arg );
        rc = ber_sockbuf_add_io( &sb, &test_io, LBER_SBIOD_LEVEL_PROVIDER, &arg );
        ber_sockbuf_remove_io(&sb, &test_io, LBER_SBIOD_LEVEL_PROVIDER);
        ber_sockbuf_remove_io(&sb, &test_io, LBER_SBIOD_LEVEL_PROVIDER);
        ber_sockbuf_remove_io(&sb, &test_io, LBER_SBIOD_LEVEL_PROVIDER);
```

Test2: report `Warning, No available liblber sockbuf_io slot found`
```
        rc = ber_sockbuf_add_io( &sb, &test_io, LBER_SBIOD_LEVEL_PROVIDER, &arg );
        rc = ber_sockbuf_add_io( &sb, &test_io, LBER_SBIOD_LEVEL_PROVIDER, &arg );
        ber_sockbuf_remove_io(&sb, &test_io, LBER_SBIOD_LEVEL_PROVIDER);
        ber_sockbuf_remove_io(&sb, &test_io, LBER_SBIOD_LEVEL_PROVIDER);
        ber_sockbuf_remove_io(&sb, &test_io, LBER_SBIOD_LEVEL_PROVIDER);

```

src/wrapped/wrappedlber.c
```diff
@@ -178,21 +178,92 @@ static void* find_sbi_close_Fct(void* fct)
...
+        return &my_sockbuf_io_struct_##A;                   \
+    }
+    SUPER()
+    #undef GO
+    printf_log(LOG_NONE, "Warning, no more slot for liblber sockbuf_io\n");
+    return NULL;
 }
 
-#include "wrappedlib_init.h"
+static my_sockbuf_io_t* find_sockbuf_io_Struct(my_sockbuf_io_t* io)
```
ptitSeb: Why duplicating the code. just add a parameter to find_ to do the add or not.
Also, you can silence  the warning if it's not a ADD operation and return the struct itself in that case...

src/wrapped/wrappedlber.c
```diff
@@ -178,21 +178,92 @@ static void* find_sbi_close_Fct(void* fct)
...
+        my_sockbuf_io_ref_count_##A--;                      \
+        if (my_sockbuf_io_ref_count_##A == 0) {             \
+            my_sockbuf_io_ref_##A = NULL;                   \
+            memset(&my_sockbuf_io_struct_##A, 0, sizeof(*io));  \
+        }                                                   \
+        return;                                             \
+    }
+    SUPER()
+    #undef GO
+    printf_log(LOG_NONE, "Warning, No available liblber sockbuf_io slot found.\n");
```
ptitSeb: The warning is incorect. It's not a lack sockbuf_io slot free slot, but more an existing slot not found.

src/wrapped/wrappedlber.c
zengdage: Done.

src/wrapped/wrappedlber.c
zengdage: Done

src/wrapped/wrappedlber.c
```diff
@@ -178,21 +178,83 @@ static void* find_sbi_close_Fct(void* fct)
...
+        wrap_sockbuf_io(&my_sockbuf_io_struct_##A, io);     \
+        my_sockbuf_io_ref_##A = io;                         \
+        my_sockbuf_io_ref_count_##A = 1;                    \
+        return &my_sockbuf_io_struct_##A;                   \
     }
-    return my->ber_sockbuf_add_io(sb, io?&my_io:NULL, layer, arg);
+    SUPER()
+    #undef GO
+    printf_log(LOG_NONE, "Warning, no more slot for liblber sockbuf_io\n");
+    return NULL;
```
ptitSeb, line 227: I think it's better to return `io` instead of `NULL` when `!alloc`...

src/wrapped/wrappedlber.c
```diff
@@ -178,21 +178,83 @@ static void* find_sbi_close_Fct(void* fct)
...
+    #define GO(A) if(!my_sockbuf_io_ref_##A) {              \
+        wrap_sockbuf_io(&my_sockbuf_io_struct_##A, io);     \
+        my_sockbuf_io_ref_##A = io;                         \
+        my_sockbuf_io_ref_count_##A = 1;                    \
+        return &my_sockbuf_io_struct_##A;                   \
     }
-    return my->ber_sockbuf_add_io(sb, io?&my_io:NULL, layer, arg);
+    SUPER()
+    #undef GO
+    printf_log(LOG_NONE, "Warning, no more slot for liblber sockbuf_io\n");
```
ptitSeb, line 226: I don't think this warning is usefull when `!alloc`

src/wrapped/wrappedlber.c
zengdage: I have added i`f(!alloc) return io;` at line 217.

src/wrapped/wrappedlber.c
zengdage: I have added `if(!alloc) return io;` at line 217, and will not go to the printf_log i`f(!alloc)`.

ptitSeb:
Thanks you (and sorry) for all the changes!

PR #4229, opened 2026-08-11
[DYNAREC] Removed the gdb backtrace feature for trace build

ksco:
Feeling unfair to pay for the performance overhead just being a developer.


ksco:
Jokes aside, I think obtaining real performance data is far more important than a rarely used debugging feature. Took me quite some time to figure out why my program runs slower on CALLRET=2 than on CALLRET=0, because I forgot the gdb backtrace hack exists.

PR #4231, opened 2026-08-11
[RV64_DYNAREC] Disable flags fusion for all the shift/rotate on CL opcodes

ksco:


ptitSeb:
I understand why you disable it, but it's a shame it needs to be done

PR #4237, opened 2026-08-12
[NON4K] Disabled the 64KiB aligned check

ksco:


ptitSeb:
are you sure it's a good idea to disable that test? I don't see why it would not be needed on 16k but would be on 4k pagesize

ksco:
It breaks the prefix creation on Wine 11.14.

ptitSeb:
Well, what is strange is that the lack of this used to break wine.

But again, I find it strange that this test breaks anythng, because it should try again and gives a good mmap the second time!

(also, pretty sure this test missing will break something later)

ksco:
Oh yeah, it breaks Wine prefix creation because the "try again" part is so slow; rundll32.exe gets stuck forever.

ptitSeb:
> Oh yeah, it breaks Wine prefix creation because the "try again" part is so slow; rundll32.exe gets stuck forever.

Might be some mutex or something like that when memory map are refreshed with current mapping and make it stuck?

Anyway, I can merge this PR, but it's not a good thing to have a different behaviour in a place where it should behave the same.

ksco:
Yeah, I need to rework the fake 4k protection stuff later.

PR #4247, opened 2026-08-14
[PROFILE] Optimized Wukong profile

ksco:


ptitSeb:
looks like a less optimise verion to me.

PR #4271, opened 2026-08-19
[CORE] Fix some logic errors reported in #4214

zengdage:
1. Preserve `state` across spaces and `%m` modifiers.

2. Initialize GVariant parser state(`tmp`) for `a` and `@` formats.

3. Advance the `st` only when an argument is read from stack.

4. Ignore NULL in XCB display registrations.

5. Fix space handling in the 32-bit wide scanf parser.

### DIFF
--- a/src/libtools/myalign32.c
+++ b/src/libtools/myalign32.c
@@ -80,18 +80,11 @@ void myStackAlign32(const char* fmt, uint32_t* st, uint64_t* mystack)
                     case 's': state = 30; break; // pointers
                     case '$': ++p; break; // should issue a warning, it's not handled...
                     case '*': *(mystack++) = *(st++); ++p; break; // fetch an int in the stack....
-                    case ' ': state=(state==1)?6:0; ++p; break;
+                    case ' ': ++p; break;
                     default:
                         state=20; // other stuff, put an int...
                 }
                 break;
-            case 6: // %"space", wants number number now
-                    switch(*p) {
-                        case '0'...'9': state = 1; ++p; break;
-                        default:
-                            state = 0; break;   // something else
-                    }
-                    break;
             case 11:    //double
             case 12:    //%lg, still double
             case 13:    //%llg, still double
@@ -758,18 +751,11 @@ void myStackAlignW32(const char* fmt, uint32_t* st, uint64_t* mystack)
                     case 's': state = 30; break; // pointers
                     case '$': ++p; break; // should issue a warning, it's not handled...
                     case '*': *(mystack++) = *(st++); ++p; break; // fetch an int in the stack....
-                    case ' ': state=(state==1)?6:0; ++p; break;
+                    case ' ': ++p; break;
                     default:
                         state=20; // other stuff, put an int...
                 }
                 break;
-            case 6: // %"space", wants number number now
-                    switch(*p) {
-                        case '0'...'9': state = 1; ++p; break;
-                        default:
-                            state = 0; break;   // something else
-                    }
-                    break;
             case 11:    //double
             case 12:    //%lg, still double
             case 13:    //%llg, still double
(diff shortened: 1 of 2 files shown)


ptitSeb:
I'l validate this one next week, after my vacaccions, so next week.

zengdage:
Test case as follow, previously these programs failed and this pr makes them run properly.
```
//   gcc $(pkg-config --cflags glib-2.0) -O1 myalign_fix.c -o myalign_fix $(pkg-config --libs glib-2.0)

#include <stdio.h>
#include <string.h>
#include <wchar.h>
#include <glib.h>

static int failures;
static int total;

static void check(const char* name, const char* got, const char* expected)
{
    ++total;
    if (strcmp(got, expected) != 0) {
        printf("[FAIL] %s: got \"%s\", expected \"%s\"\n", name, got, expected);
        ++failures;
    } else {
        printf("[PASS] %s: \"%s\"\n", name, got);
    }
}

static void check_variant(const char* name, GVariant* v, const char* expected)
{
    if (v) {
        gchar* s = g_variant_print(v, TRUE);
        check(name, s, expected);
        g_free(s);
        g_variant_unref(v);
    } else {
        printf("[FAIL] %s: NULL\n", name);
        ++failures;
        ++total;
    }
}

int main(void)
{
    char buf[64];

    snprintf(buf, sizeof(buf), "[% d][% 6d]", 7, 8);
    check("printf % d", buf, "[ 7][     8]");

    /* @i as 6th arg (first from stack) */
    check_variant("overflow @i",
        g_variant_new("(iiiii@i)", 1, 2, 3, 4, 5, g_variant_new("i", 42)),
        "(1, 2, 3, 4, 5, 42)");

    /* two @i on stack — second relies on tmp reset */
    check_variant("overflow @i@i",
        g_variant_new("(iiiii@i@i)", 1, 2, 3, 4, 5,
            g_variant_new("i", 10), g_variant_new("i", 20)),
        "(1, 2, 3, 4, 5, 10, 20)");

    /* @(ii) on stack — exercises tmp nesting in state 3 */
    check_variant("overflow @(ii)",
        g_variant_new("(iiiii@(ii))", 1, 2, 3, 4, 5,
            g_variant_new("(ii)", 6, 7)),
        "(1, 2, 3, 4, 5, (6, 7))");

    /* ai as 6th arg — exercises state 1 in overflow */
    GVariantBuilder* b = g_variant_builder_new(G_VARIANT_TYPE("ai"));
    g_variant_builder_add(b, "i", 8);
    g_variant_builder_add(b, "i", 9);
    check_variant("overflow ai",
        g_variant_new("(iiiiiai)", 1, 2, 3, 4, 5, b),
        "(1, 2, 3, 4, 5, [8, 9])");
    g_variant_builder_unref(b);

    printf("\n%d passed, %d failed\n", total - failures, failures);
    return failures != 0;
}
```

src/libtools/myalign32.c
```diff
@@ -80,7 +80,7 @@ void myStackAlign32(const char* fmt, uint32_t* st, uint64_t* mystack)
                     case 's': state = 30; break; // pointers
                     case '$': ++p; break; // should issue a warning, it's not handled...
                     case '*': *(mystack++) = *(st++); ++p; break; // fetch an int in the stack....
-                    case ' ': state=(state==1)?6:0; ++p; break;
```
ptitSeb, line 83: you remove the case 6 here, and yet, let the code for case 6 in, as dead code.
1. Are you sure of that?
2. if you are, remove case 6 code then, don't let dead code in.

src/libtools/myalign32.c
```diff
@@ -758,7 +758,7 @@ void myStackAlignW32(const char* fmt, uint32_t* st, uint64_t* mystack)
...
```
ptitSeb, line 761: same issue with `case 6`

src/libtools/myalign32.c
zengdage: Done. Removed the state 6 dead code. 

state 6 is used for scanf, not for printf. Spaces behave differently in `printf` and `scanf`. In `scanf`, a space can resets the parsing state.

src/libtools/myalign32.c
zengdage: Same as above.

PR #4273, opened 2026-08-20
[TRACE] Added %name support for BOX64_TRACE_FILE

Mikura-Kyouka:
BOX64_TRACE_FILE can now use %name to include the target executable basename in the trace filename, in addition to the existing %pid.

This is useful when tracing is enabled globally through an rcfile or environment variable, including Wine and Wine preloader invocations. The name is sanitized before being used as a filename and falls back to "unknown" when it cannot be determined.

Updated the relevant documentation.

Tested by running Needy Girl Overdose via Steam.

屏幕截图_20260820_132515

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -760,6 +760,7 @@ Enable or disable EMM (i.e. MMX) trace output.
 Send all log and trace to a file instead of `stdout`.
 
  * XXXX: Send all log and trace to file XXXX.
+ * XXXX%name: Send all log and trace to file XXXX with executable name appended.
  * XXXX%pid: Send all log and trace to file XXXX with pid appended.
  * stderr: Send all log and trace to `stderr`. [Default]
  * stdout: Send all log and trace to `stdout`.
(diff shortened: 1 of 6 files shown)


ksco:
The idea is okay, but the code seems to be some copy-paste from the `initialize()` function. I understand that you may try not to break anything, but there is quite some duplicated code there.

Mikura-Kyouka:
> The idea is okay, but the code seems to be some copy-paste from the `initialize()` function. I understand that you may try not to break anything, but there is quite some duplicated code there.

To be honest, I can't think of a better way... Obtaining the executable filename in `initialize()` is more of a bonus than the main purpose, so these identification logics can't be eliminated here. While `openFTrace()` needs to execute before those identification logics in `initialize()` in order to prepare where to write the trace...

ksco:
Sure, but someone needs to clean this up later.

ptitSeb:
looks ok to me, but yeah, some cleaning to make the change smaller would be nice.

PR #4274, opened 2026-08-20
[ELFLOADER] Support versioned data symbols

zengdage:
This patch is to fix `sys_siglist` not found error reported by dingtalk app.


src/librarian/library.c
```diff
@@ -836,6 +836,12 @@ static int getSymbolInDataMaps(library_t*lib, const char* name, int noweak, uint
         symbol = (void*)(kh_value(lib->w.datamap, k).addr);
         #else
         symbol = dlsym(lib->w.lib, kh_key(lib->w.datamap, k));
+        // glibc keeps some data objects such as `sys_siglist` versioned, so
```
ptitSeb: I don't think this comment is really usefull

zengdage: Done.

PR #4280, opened 2026-08-22
[ARM64_DYNAREC] Fix memory corruption in unaligned LOCK CMPXCHG8B/16B

devarajabc:
Since x5 is guaranteed to be 0 after line CBNZw_MARK2(x5), this incorrectly zeroed the first byte.
Changed x5 to x2 to safely write back the original lowest byte fetched by the initial LDP.


ptitSeb:
Nice catch.

PR #4284, opened 2026-08-24
[DYNAREC] Added preliminary interblocks deferred flags elimination

ksco:
The first version is probably too conservative, but it gives good results in some benchmarks.

### DIFF
--- a/src/dynarec/dynarec_native.c
+++ b/src/dynarec/dynarec_native.c
@@ -70,6 +70,11 @@ void printf_x64_instruction(dynarec_native_t* dyn, zydis_dec_t* dec, instruction
 void add_next(dynarec_native_t *dyn, uintptr_t addr) {
     if (!BOX64DRENV(dynarec_bigblock))
         return;
+    if (dyn->peeking_flags) {
+        if (dyn->next_sz < dyn->next_cap)
+            dyn->next[dyn->next_sz++] = addr;
+        return;
+    }
     int ret;
     kh_put(nextset, khnextset, addr, &ret);
     if(!ret)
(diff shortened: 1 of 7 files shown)


ptitSeb:
So for each "peekeable" jump, a limited 16 instructions decoding using only pass0 is done at the landing address to check if flags are regenated and so useless to store a defered flag befor ethe jump. Clever reuse of existing code!

src/dynarec/dynarec_native.c
```diff
@@ -409,6 +414,91 @@ uintptr_t native_pass1(dynarec_native_t* dyn, uintptr_t addr, int alternate, int
...
+#define PEEK_FLAGS_INSTS 16
+
+typedef struct peek_flags_state_s {
+    dynarec_native_t dyn;
+    instruction_native_t insts[PEEK_FLAGS_INSTS + 2];
+    int jmps[PEEK_FLAGS_INSTS + 2];
+    uintptr_t next[PEEK_FLAGS_INSTS + 2];
+} peek_flags_state_t;
+
+static __thread peek_flags_state_t peek_flags_state;
```
ptitSeb: I think this break the Wow64 build:
```
ld.lld: error: undefined symbol: _tls_index
/usr/local/bin/cmake -E rm -f CMakeFiles/wowbox64.dir/objects.a
>>> referenced by objects.a(dynarec_native.c.obj):(interblock_flags_needed)
>>> referenced by objects.a(dynarec_native.c.obj):(interblock_flags_needed)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

On the other hand, I'm not sure it's necessary to have this thread local has dynarec is not supposed to be multithreaded: everything is static and dynablock building is behind a mutex.

PR #4285, opened 2026-08-24
[WOWBOX64] Initialize memory before reading RC file

airidosas252:
This fixes hangs when .box64rc file is provided.

### DIFF
--- a/wine/wow64/wowbox64.c
+++ b/wine/wow64/wowbox64.c
@@ -245,8 +245,9 @@ NTSTATUS WINAPI BTCpuProcessInit(void)
 #undef STATIC_ASSERT
 
     LoadEnvVariables();
-    InitializeEnvFiles();
     InitializeSystemInfo();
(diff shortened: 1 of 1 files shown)


wine/wow64/wowbox64.c
```diff
@@ -245,8 +245,9 @@ NTSTATUS WINAPI BTCpuProcessInit(void)
 #undef STATIC_ASSERT
 
     LoadEnvVariables();
-    InitializeEnvFiles();
     InitializeSystemInfo();
```
ptitSeb, line 248: The issue with this ordering is that `InitialiazeSystemInfo()` might depends on some of the env. var. in `InitializeEnvFiles()` (for cpu cores counts & rdtsc options)

airidosas252: I think I just overthinked this too much...

The order of inits was correct, custommem init wasn't. I'll fix it up later.

PR #4290, opened 2026-08-25
[CMAKE] Optionally build without box64-configurator

### LINKED ISSUE #3957: Add cmake config to disable compiling box64-configurator
It would reduce compiling time

LukeShortCloud:
to speed up build times.

Resolves #3957

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ option(LARCH64_ABI_1 "Set to ON if building for Loongarch64 ABI 1.0 system" ${LA
 option(GDBJIT "Set to ON to enable GDB JIT support" ${GDBJIT})
 option(WOW64 "Set to ON if you want a WOW64 PE build in addition")
 option(BOX64_KDE_PROFILE_PLUGIN "Build the Dolphin/KIO Box64 profile properties plugin when KF5 or KF6 is available" OFF)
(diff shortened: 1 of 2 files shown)


ksco:
box64-configurator is a Python project; it takes no time to build.

LukeShortCloud:
I was skeptical about that, too. I figured maybe the related KDE profile plugin is what took so long.

I have now profiled building with and without `box64-configurator`. With it enabled, it ironically compiled 6 seconds faster lol.

I updated the the commit message to remove the mention of "to speed up build times." The configurable option could still be useful. Morale of the story: do not trust everything you read on the Internet.

LukeShortCloud:
Updated the compilation documentation now, too.

ksco:
But if it does not affect the build time, why disable it?

LukeShortCloud:
My only thought is to avoid a Python runtime dependency for Box64 packages. Otherwise, I can close this out.

ksco:
Python is already a hard dependency for building Box64. For end users, we also ship box64-python, which requires a native Python installation too. Also, btw, the KDE plugin is disabled by default for obvious reasons.

LukeShortCloud:
Sounds good. Thanks for the information. I'll close this out and the related GitHub issue.

PR #4297, opened 2026-08-27
[WRAPPER] Fix some errors reported in #4214

zengdage:
1. Correct variadic argument offsets for `my_g_fprintf` and `my_dbus_message_new_error_printf`.

2. Fix callback bridge reverse lookups in glib2 and libx11.

3. Correct `gdk_window_set_modal_hint` signature. `void gdk_window_set_modal_hint (GdkWindow *window, gboolean modal)`

4. Add `mq_notify` callback translation for `SIGEV_THREAD` notifications.


ptitSeb:
ANDROID (and TERMUX) build are Red: mqueue is not present there, so the `my_mq_notify(...)` needs to follow the same scheme as `my_aio_cancel(...)` for example.

PR #4301, opened 2026-08-28
[WRAPPER] Enforce RLIMIT_AS set by the guest instead of ignoring it

ksco:


ptitSeb:
I don't remember why I did that, but I'm sure there was a reason, with a program that limited something that box64 was already using (like dynablock in very high space). Why do you need this?

ksco:
Yeah, a low limit may break Box64's DynaRec allocations, but the raw `setrlimit` syscall wrapper as well as the `prlimit64` wrapper do not have this hack, so there are inconsistencies.

By the way, the kernel will keep existing mappings when a process lowers RLIMIT_AS below its current usage; only new     
 mappings fail with ENOMEM.

PR #4303, opened 2026-08-28
[MMAP] Don't allocate bridge bricks in guest low memory for 64bits guests

ksco:
The guest can MAP_FIXED over low addresses, destroying the bridges living there, and even munmap them. Only allocate the bridges low for 32bits guests, which needs the guest bitness to be detected before the context is created.


ptitSeb:
You mean, there are 64bits program that blindily allocate 32bits memory with the assuption that 32bits memory is just all free to use?

ksco:
Basically, yes. But for 64-bit programs, there is a choice for us not to put the bridge page there.

PR #991, opened 2023-09-25
Add SDL_RenderWindowToLogical and SDL_SetSurfaceColorMod wrapped functions for #939

seeschloss:
I'm far from a box64 or SDL expert, but this seems sufficient for me to run Dwarf Fortress 0.50.10 on a M1 machine, see bug #939

PR #4308, opened 2026-08-28
Add SDL_RenderLogicalToWindow function wrapper

seeschloss:
A minor addition needed to run dfhack 53.16-r1.1.

I'm not sure whether I should add the same to "wrapped32", as the reverse function (SDL_RenderWindowToLogical) is there but commented out. I can't test it so I just left it out for now.


ksco (changes requested):
```
void SDL_RenderLogicalToWindow(SDL_Renderer * renderer, 
                                    float logicalX, float logicalY,
                                    int *windowX, int *windowY);
```

seeschloss:
I'm not sure I understand which change is requested? I just did the same as for SDL_RenderWindowToLogical (https://github.com/ptitSeb/box64/pull/991) except with the arguments swapped around.

seeschloss:
WTF sorry, I understand now, what I don't understand is how it even works at all (I've actually been playing dwarf fortess/dfhack with that patch for a while now).

seeschloss:
Here's a less stupid version, hopefully

ksco (approved):

ptitSeb:
I think your commit is now missing some generated files, as this is probably a new signature.

seeschloss:
Thanks and sorry, I had missed this thing about generated files. I'll keep it in mind next time

PR #4312, opened 2026-08-28
[DYNAREC] Don't truncate registers to 32bits when switching to 32-bit mode

ksco:


ptitSeb:
Really? Aren't you affraid of some nasty side effect in dynarec?

ksco:
It's not about being afraid or not, real hardwares does not have this behaviour.

ptitSeb:
Ok, but that still might bring some issue on 32bits code, because the upper 32bits are often considered 0 and not enforced...

ksco:
> Ok, but that still might bring some issue on 32bits code, because the upper 32bits are often considered 0 and not enforced...

We'll see.

PR #4314, opened 2026-08-28
Space Box 86 and Box32 emulator details to README

Mosesbanda775:
Updated README to include Box32 emulator information and related links.


ksco:
I'm so confused.

PR #4316, opened 2026-08-29
[SYSCALL] Fixed io_getevents and io_setup wrapping

ksco:


src/emu/x64syscall.c
```diff
@@ -565,7 +565,8 @@ void EXPORT x64Syscall_linux(x64emu_t *emu)
     }
     // check wrapper first
     uint32_t cnt = sizeof(syscallwrap) / sizeof(scwrap_t);
-    if(s<cnt && syscallwrap[s].nats) {
+    // __NR_io_setup is 0, so syscallwrap[s].nats alone is not enough
+    if(s<cnt && !(syscallwrap[s].nats == 0 && syscallwrap[s].nbpars == 0)) {
```
ptitSeb, line 569: Mmmm, why entrering the if only for `syscallwrap[s].nbpars == 0` and leave the `switch(syscallwrap[s].nbpars)` in the if, when only the `0` case will be run, and what will happens to the others cases?

ptitSeb: Oh, it's just to exclude an all 0 case, I see

PR #4321, opened 2026-08-29
[ARM64_DYNAREC] Raised SIGFPE on div0

ksco:
This is Arm64 DynaRec only for now; other architectures will come later.


ptitSeb:
Are you sure about the AC flags and it's effect? Is it really modifiable in user space? and triggering stuffs on a simple MOV???

ksco:
Oh, it's indeed a system flag.....

ksco:
F_AC stuff removed.

PR #4323, opened 2026-08-29
[STEAM] Added a hack to fix the stair-stepping logging issue

ksco:
I believe this is not a Box64 issue, so all we can do is a hack..


ptitSeb:
Ooooh, that thing!

PR #4324, opened 2026-08-30
[RV64_DYNAREC] Added preliminary RVV1.0 (vlen>=256) AVX/AVX2 implementation

ksco:
Fully tested. Not optimized for performance (yet).


src/dynarec/rv64/dynarec_rv64_660f_vector.c
```diff
@@ -1708,45 +1717,7 @@ uintptr_t dynarec64_660F_vector(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t i
             VMERGE_VIM(q0, q0, 0b11111); // implies vmask and widened it
             break;
         case 0x7C:
-            INST_NAME("HADDPD Gx, Ex");
```
ptitSeb, line 1711: Why removing this one? is it wrong somehow?

ksco: Yeah, scratch registers are short here. I’ll add it back later.

ptitSeb: Sounds good

PR #4335, opened 2026-09-01
[SIGNAL] Refined handling of invalid sigsetsize in syscall_rt_sigaction

ksco:


ptitSeb:
Why removing the old mechanism for one that is more restrictive? is that the expected behaviour to just refuse the function call if sigsetsize is too small (or too big)? Also, this is expecting sigset_t to be exactly the same size on x64 and native arch (which I assume is not always true, especially on machine that have like 128 signals instead of 64)

ksco:
> is that the expected behaviour to just refuse the function call if sigsetsize is too small (or too big)? 

Yes, it is how the kernel handles it

ksco:
> Also, this is expecting sigset_t to be exactly the same size on x64 and native arch (which I assume is not always true, especially on machine that have like 128 signals instead of 64)

Fair point! Fixed.

PR #4340, opened 2026-09-01
[LA64_DYNAREC] Added host_call to arch state

ksco:
This fixed  box64-bash's unstablility on LoongArch.

### DIFF
--- a/src/libtools/signals.c
+++ b/src/libtools/signals.c
@@ -346,7 +346,7 @@ void leave_critical_section()
     emu->deferred_signal_processing = 0;
 }
 
-int my_sigactionhandler_oldcode_64(x64emu_t* emu, int32_t sig, int simple, siginfo_t* info, void * ucntx, int* old_code, void* cur_db)
+int my_sigactionhandler_oldcode_64(x64emu_t* emu, int32_t sig, int simple, siginfo_t* info, void * ucntx, int* old_code, void* cur_db, uintptr_t x64pc)
 {
     int Locks = unlockMutex();
     int log_minimum = (BOX64ENV(showsegv))?LOG_NONE:LOG_DEBUG;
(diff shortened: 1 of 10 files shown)


src/libtools/signals.c
```diff
@@ -388,8 +388,12 @@ int my_sigactionhandler_oldcode_64(x64emu_t* emu, int32_t sig, int simple, sigin
     void* pc = NULL;
     if(p) {
         pc = (void*)CONTEXT_PC(p);
-        if(db)
-            frame = (uintptr_t)CONTEXT_REG(p, xRSP);    //this should not be needed, as emu has been "adjusted" to dynablock value already in the caller
+        if(db) {
+            if(ARCH_HOST_CALL(db, x64pc))
+                frame = R_RSP;
```
ptitSeb, line 393: I'm not sure what difference does that does? the emu-->RSP and signal->RSP should be identical at this stage, as we are inside a "call_c" basicaly?

ksco: RSP is `a7`, which is clobbered by the native function.

ptitSeb: Ok, but only after RSP is actualy saved, wich happens quite far in the opcode. I think you need some kind of marker (at least the number of native opcode since the start of the opcode until this is a valid assumption)

ksco: Added a call_window offset so that we know RSP is valid.

ptitSeb:
LGTM now. I suppose ARM64 doesn't needs that for now, as RSP is not clobbered during native call. But RV64 (and PPC64LE) might need something similar at some point.

xiangzhai:
Hi,

> But RV64 (and PPC64LE) might need something similar at some point.

RV64 might doesn't need. Because RV64 `RSP` mapping to `x9` (`s1`), Saved register, not Function argument.
There is `uintptr_t xSPSave` in the `x64emu_t` for RV64 and PPC64LE. But LA64 `xSavedSP` mapping to `$r22` (`$fp`), otherwise `RSP` mapped to `$r22` instead.
But LA64 mapping to `$r11` (`a7`), Function argument.
Do I need to refactor the la64 mapping?

Thanks,
Leslie Zhai

ksco:
It's not just RSP.

PR #4346, opened 2026-09-03
[WRAPPER] Add missing 32bit libc/libresolv symbols

Ericky14:
Disclaimer: Claude code has been used to diagnose the issue.

When using `medic` steam runtime, it seems to leak some library paths onto the system, at least on my setup.. causing steam client to use glib 2.84 instead of 2.58. It's probably a bug in my setup, but I thought this would be nice to have anyways in case of someone using newer glib.

Reproducer:
```
    BOX64_LD_LIBRARY_PATH=$RT/lib/i386-linux-gnu:$RT/lib \
    BOX64_LD_PRELOAD=$RT/lib/i386-linux-gnu/libgio-2.0.so.0 \
      box64 ~/.steam/steam/ubuntu12_32/gldriverquery
```

Before:
```
    exit=134 (Aborted)
    Symbol copy_file_range not found
    Symbol dn_expand not found
    Symbol getservbyname_r not found
    Symbol lchmod not found
    Symbol res_nquery not found
    Unhandled signal caught
```

After:
```
    exit=0, no unresolved symbols
```


ptitSeb:
Looks good, thanks Ericky.

Issue #4347, opened 2026-09-03
Error: Symbol getaddrinfo_a not found, cannot apply R_X86_64_JUMP_SLOT


mountainlord:
running with unity, first loading dll, cannot load libnanomsg.so, error msg below, can someone tell me why?
[[31mBOX64[0m] Error: Symbol getaddrinfo_a not found, cannot apply R_X86_64_JUMP_SLOT @0x3f1f045dc0 (0xeb80) in /opt/apps/eastsim/projectmanager/VOSPRun/1.0/esu3d_Data/Plugins/libnanomsg.so (optver=3 / GLIBC_2.2.5)
[[31mBOX64[0m] Error: relocating Plt symbols in elf libnanomsg.so

PR #4352, opened 2026-09-03
[WRAPPER] Wrapped getaddrinfo_a for libc

ksco:
For #4347


ptitSeb:
Nope, this needs to be a GOM because the structure can hide a function callback
image

ksco:
Oh

PR #4357, opened 2026-09-04
[LA64_DYNAREC] Added fastpath for native nofpu calls

ksco:


ptitSeb:
Oh, so nofpu is based on a few function name, not generic signature analysis because it's for function that doesn't even use fpu internaly?

ksco:
> Oh, so nofpu is based on a few function name, not generic signature analysis because it's for function that doesn't even use fpu internaly?

Yeah, that's the idea.

PR #4366, opened 2026-09-06
[RV64_DYNAREC] Added AES and SHA opcodes

ksco:


ptitSeb:
Oooh, crypto extensions, nice!

PR #4375, opened 2026-09-10
[ELFLOADER] Suppport R_386_IRELATIVE

zengdage:
1. Support `R_386_IRELATIVE `for ifunc.

2. Initialize GOT[1] and GOT[2] before processing i386 dynamic relocations, so PLT0 has valid elf header and resolver function during IFUNC and IRELATIVE handling.

3. Add `!bindnow` check for `SeedPltResolverxx`.


ptitSeb:
I see there is some fixes from last PR also, mixed into this one...

PR #4376, opened 2026-09-10
[LA64_DYNAREC] Added leaf call inline optimization

ksco:


ptitSeb:
LGTM. Just instead of all the `#ifdef LA64` in the dynarec code, why not something more generic like `HAS_LEAF` or something like that, defined in `dynarec_arch.h`?

src/dynarec/la64/dynarec_la64_pass0.h
```diff
@@ -70,6 +70,14 @@
     } else                                         \
         dyn->insts[ninst].barrier_maybe = 1
 #define SET_HASCALLRET() dyn->insts[ninst].x64.has_callret = 1
+#define MARK_LEAF_CALL() dyn->insts[ninst].x64.leaf_kind = LEAF_KIND_CALL
```
ptitSeb, line 73: does kind needs to be a bitfield instead of a single value (i.e. is there a risk of an opcode having ultiple kind, like "push" & "call" for example)?

ksco: I don't think so. A single value seems enough.

ptitSeb: ok

PR #4379, opened 2026-09-11
[RV64_DYNAREC] Fix clobbering scratch register for VDPPS

zengdage:
Hoist scratch register allocations (d0/q2/d1/d2) outside the loop to prevent exhaustion when vex.l=1.

The original code allocated 3 vector registers per lane iteration, exceeding the vector registers number limit and causing some vector register aliasing.

Also merge 4 per-lane output masks into one combined mask.


zengdage:
Reported by https://github.com/zengdage/x86_64-inst-test/blob/main/avx256/test_vdppd_dpps.c

ptitSeb (approved):

ksco (approved):

PR #4389, opened 2026-09-14
[RV64_DYNAREC] Introduced a dynarec version of the UpdateFlags helper function

ksco:


src/dynarec/rv64/dynarec_rv64_consts.c
```diff
@@ -106,6 +108,7 @@ uintptr_t getConst(rv64_consts_t which)
         case const_cpuid: return (uintptr_t)my_cpuid;
         case const_getsegmentbase: return (uintptr_t)GetSegmentBaseEmu;
         case const_updateflags: return (uintptr_t)UpdateFlags;
```
ptitSeb, line 110: The legacy "updateflags" can be removed I suppose.

ksco: Indeed, I'll submit another PR to remove it for all 3 DynaRecs.

ptitSeb: Thanks

PR #4394, opened 2026-09-15
[CUSTOMMEM] Fixed race between concurrent mmaps silently overwriting guest mappings

ksco:
This fixes the unstability of steamwebhelper.


ptitSeb:
WowBox64 build is broken. Some pthread macro issue it seems, or something like that

src/include/custommem.h
```diff
@@ -3,10 +3,12 @@
 #define __CUSTOM_MEM__H_
 #include <unistd.h>
 #include <stdint.h>
-
+#include <pthread.h>
```
ptitSeb: This break WowBox64 build

src/include/custommem.h
```diff
@@ -3,10 +3,12 @@
...
 
 typedef struct box64context_s box64context_t;
 
+extern pthread_mutex_t mutex_mmap;
```
ptitSeb: also, this is not correct when `CUSTOM_MUTEX` is used (it's an ̀`uint32_t` instead)

src/custommem.c
```diff
@@ -3461,6 +3489,12 @@ EXPORT void* box_mmap(void *addr, size_t length, int prot, int flags, int fd, ss
         }
     }
     #endif
+    if(ret!=MAP_FAILED) {
+        LOCK_PROT();
+        rb_set(mapallmem, (uintptr_t)ret, ALIGN((uintptr_t)ret + length), MEM_RESERVED);
+        UNLOCK_PROT();
+    }
+    mutex_unlock(&mutex_mmap);
```
ptitSeb, line 3497: Isn't that a risk of deadlock? can the unlock of mutex_mmap be before the LOCK_PROT instead?

ksco: I don't think so--no code path holding mutex_prot ever asks for mutex_mmap. But good point, I added a comment on the usage of these mutexes.

ptitSeb: But `LOCK_PROT()` is used for some rbtree modification, wich will use some memory at some point, that might then use an `mmap`. Not sure it would trigger the `mutex_mmap` tho.

ptitSeb: Is this case is already handled with the deferred data stuffs already in place?

ptitSeb: I guess we can check later if there is deadlock, and refer to the comment to not forget this

ksco: Nah, it will not. I checked this very carefully.

PR #4396, opened 2026-09-16
[RV64_DYNAREC] Share fast hash between RV64 and PPC64LE backends

zengdage:
Move the fast hash function from the PPC64LE backend into dynablock.c as a shared helper, `normal_fast_hash`.

RV64 previously had no `ARCH_CRC` defined, so `X31_hash_code` fell back to the slow byte-by-byte loop on every dynablock validation.

PPC64LE keeps the same algorithm, just routed through the shared helper.


zengdage:
I previously attempted to accelerate CRC32 computation using the risc-v clmul family of instructions, but the performance was worse than ppc64le_hash. Therefore, using ppc64le_hash directly is more better.

Test data:
```
len=200, iters=2000000

offset         x31_hash    rv64_clmulh    rv64_clmulr     rv64_linux   ppc64le_hash
------         --------       --------       --------       --------       --------
off=0          353.1 ns       91.3 ns      119.7 ns      128.5 ns       50.0 ns
off=1          349.1 ns      102.7 ns      132.1 ns      145.4 ns       51.5 ns
off=2          358.0 ns       99.2 ns      127.1 ns      146.7 ns       51.4 ns
off=3          352.5 ns      104.2 ns      131.5 ns      143.6 ns       51.5 ns
off=4          352.5 ns       94.4 ns      123.3 ns      141.5 ns       51.4 ns
off=5          352.0 ns      104.1 ns      132.9 ns      141.2 ns       51.5 ns
off=6          349.0 ns      101.6 ns      128.3 ns      142.0 ns       51.4 ns
off=7          352.1 ns      105.9 ns      132.8 ns      141.5 ns       51.4 ns

x31_hash      (off=0) = 0xFE50F118
x31_hash      (off=3) = 0x3956DA18
ppc64le_hash  (off=0) = 0xF6EC1308
ppc64le_hash  (off=3) = 0xB5BD86F4
rv64_clmulh   (off=0) = 0xC4D0A04D
rv64_clmulh   (off=3) = 0x72763394
rv64_clmulr   (off=0) = 0xC4D0A04D
rv64_clmulr   (off=3) = 0x72763394
rv64_linux    (off=0) = 0xC4D0A04D
rv64_linux    (off=3) = 0x72763394
```

ptitSeb (approved):

ksco:
A "fastpath" is meant to be specific to the architecture. If it works so well, I think we should make it the fallback, i.e., to replace the fallback path of X31_hash_code.

ptitSeb:
Note that the result hash is different from X31, that's fine, but we need to be sure there is no C vs ASM result comparison used (like with auto-CRC for example)

*EDIT*: We should be fine, becasue of `ARCH_CRC` that will used an optimised version if available.

zengdage:
I checked the locations where hash codes are used in arm64 and la64. I found no code paths that compare the C and assembly implementations of X31 hash.

- On arm64, either `arm64_crc` or `arm64_x31_hash` is used, the C implementation of X31 hash is never reached.
- On la64, `la64_crc` and `la64_crc_autocrc` are used, and likewise the C X31 hash implementation is unreachable.

Therefore, I think we can use `normal_fast_path` as the fallback hash path. 

Either approach works for me. I'll implement whichever you decide on.

ptitSeb:
PR is good to go for me. Waiting for @ksco approuval.

PR #4406, opened 2026-09-17
[RV64_DYNAREC] Reduce FLAGS_ADJUST_TO11 from 6 to 4 insts

zengdage:
we can use ~(1UL<<5) directly for and operation, no need to construct 0xFFFFFFFFFFFFF7DF.


ptitSeb:
But `~(1UL<<5)` (so `0xffffffffffffffdf`) and `0xFFFFFFFFFFFFF7DF`are 2 different value, you are missing 0x800 wich is not doable directly.

zengdage:
> But `~(1UL<<5)` (so `0xffffffffffffffdf`) and `0xFFFFFFFFFFFFF7DF`are 2 different value, you are missing 0x800 wich is not doable directly.

Sorry, I got it wrong. Close this PR.

PR #4407, opened 2026-09-17
[RV64_DYNAREC] Optimize aligned REP STOSB with DF=0

zengdage:
Keep DF=1, unaligned path and RCX<8 unchanged.

Use 64-bit stores for the aligned path, reducing the loop overhead for large byte fills.


zengdage:
Trade minor regression for unaligned paths in exchange for a substantial speedup on aligned paths.

ptitSeb (approved):

ptitSeb (approved):

PR #4408, opened 2026-09-17
[INTERP] Honor MXCSR rounding mode in SSE/AVX operations

yzewei:
The interpreter implements many SSE and AVX instructions using host floating-point operations without synchronizing with the guest's MXCSR.RC. `mxcsr_setround()` has been added to save the host rounding mode and temporarily apply the rounding mode selected by the guest's MXCSR (similar to `x87_setround()`). Regression tests for the interpreter have also been added, covering all four MXCSR rounding modes and a representative set of scalar, packed, conversion, AVX, and FMA instructions.


ptitSeb (approved):
I don't really like the coding style (with the break inside the braces of the case...

yzewei:
> I don't really like the coding style (with the break inside the braces of the case...

I can modify it.= .  =

Issue #4382, opened 2026-09-13
[Bug] In Termux glibc‑runner, attempting to run Wine using box64 caused box64 to crash


ocnedkf:
In the glibc‑runner environment of Termux, box64 installed via the package‑manager triggers a "buffer overflow detected" error and SIGABRT when running Wine, causing the process to abort. I tried running an Ubuntu container with proot to recompile the latest box64 and modified the ELF interpreter binary, after which the problem disappeared. The only difference in the compilation parameters that could have an effect was that the box32 feature was not enabled.

Would it be possible to add a patch to fix this issue? Thank you！

[Other related issues](https://github.com/termux-pacman/glibc-packages/issues/386)

ptitSeb:
I have no issue adding a fix to this issue, but I have no idea what to fix for now? The logs in the other ticket is not realy detailled enough and just gives a buffer overflow. At least runs with `BOX64_LOG=2` maybe?

Also, having latest box64 v0.4.5 built with box32 would be interesting (I have seen some symbols issue preventing the build it seems, but there is no details, so I cannot really help either)

ocnedkf:
Alright, I've attached the box64 log collected with `BOX64_LOG=2`. I will also test and report its behavior when the Box32 feature is enabled at a later time.

ocnedkf:
[box64_log.txt](https://github.com/user-attachments/files/32158785/box64_log.txt)

ocnedkf:
@ptitSeb I just tried compiling the latest Box64 with Box32 support using proot in an Ubuntu container and in a Termux glibc environment. The build and compilation processes went very smoothly without any errors. The only issue occurred when attempting to install to a specified directory in the Termux glibc environment, where some installation paths were hardcoded under /usr, causing the installation to fail. I resolved this problem by modifying cmake_install.cmake. Additionally, I also tried running the Box64 compiled in Ubuntu within the Termux glibc environment. It did not encounter any issues when running my Wine. I believe this might be because my Wine operates in wow64 mode, which prevented it from exhibiting any issues related to Box32.

ptitSeb:
[...]

PR #4426, opened 2026-09-18
Fix buffer overflow and stack smashing in FileIs* functions during execve

libx264:
### Summary
Fixes an issue where `execve` (inside `my_execve`) crashes with `*** buffer overflow detected ***` or `*** stack smashing detected ***: terminated` (`SIGABRT`) when calling `FileIsX64ELF`, `FileIsX86ELF`, `FileIsShell`, or `FileIsPython`.

### Problem Description
In `src/utils/fileutils.c`, the file-checking functions used tightly sized stack buffers:
- `char head[20]` in `FileIsX64ELF`, `FileIsX86ELF`, and `FileIsShell`
- `char head[25]` in `FileIsPython`

This caused two issues:
1. **Fortify Source false-positives / crashes:** When compiled against modern glibc or toolchains with `_FORTIFY_SOURCE` / `__fread_chk` enabled, `fread(head, size, 1, f)` triggered:
   ```text
   *** buffer overflow detected ***: terminated
   ```
2. **Stack smashing:** When fortify checks were bypassed, reading close to the buffer boundary without safety padding corrupted the stack canary / alignment on ARM64, resulting in:
   ```text
   *** stack smashing detected ***: terminated
   ```
3. In `FileIsShell` and `FileIsPython`, strings were read directly to the end of the buffer without leaving room for a trailing `\0` null-terminator for subsequent `strncmp` calls.

#### Backtrace:
```text
*** buffer overflow detected ***: terminated
[BOX64] NativeBT: libc.so.6(abort+0x28)
[BOX64] NativeBT: libc.so.6(__fread_chk+0x194)
[BOX64] NativeBT: box64() [my_execve helper]
[BOX64] NativeBT: box64(my_execve+0xc8)
[BOX64] EmulatedBT: box64(execve+0)
[BOX64] EmulatedBT: box64-bash(shell_execve+32)
```

### Changes
- Increased `head` buffer size to `64` bytes in `FileIsX64ELF`, `FileIsX86ELF`, `FileIsX64X86ELF`, `FileIsShell`, and `FileIsPython`.
- Changed `fread` call convention from `fread(head, size, 1, f)` to `fread(head, 1, count, f)` to prevent buffer overflows and ensure safe bounds.
- In `FileIsShell` and `FileIsPython`, limit reading to `sizeof(head) - 1` to guarantee the buffer is always null-terminated.
- Updated return checks to verify that enough bytes were read (`sz < expected_length`) rather than strict `sz != 1`.

### Testing
- Tested on ARM64 with glibc and `-fstack-protector-strong`.
- Verified nested `execve` calls within `box64-bash` (running commands and subshells) now execute cleanly without triggering fortify or stack corruption.


libx264:
Should fix #4382

ptitSeb:
I'm sorry, but I don't understand were is the fix? the fread change you do is ... strange. The form used in box64 is perfectly legal and correct, so please expand (and please, don't use AI to explain, it's your PR, not an AI one)

src/tools/fileutils.c
```diff
@@ -67,18 +70,19 @@ int FileIsX64ELF(const char* filename)
...
     fclose(f);
-    if(sz < 20) {
+    if(sz != 1) {
         return 0;
     }
-    head[7] = x64lib[7];   // this one changes
-    head[8] = x64lib[8];   // AppImage customized this
-    head[9] = x64lib[9];   // and this one too
-    head[10] = x64lib[10]; // and that last one too
+    head[7] = x64lib[7];
```
ptitSeb: why did you removed those comments?

src/tools/fileutils.c
```diff
@@ -67,18 +70,19 @@ int FileIsX64ELF(const char* filename)
     FILE *f = fopen(filename, "rb");
     if(!f)
         return 0;
-    char head[64] = {0};
-    size_t sz = fread(head, 1, 20, f);
+    
+    char head[sizeof(x64lib)] = {0};
```
ptitSeb, line 74: what was wrong with the 64-bytes size of the array before, because I think that was the onyl fix there

src/tools/fileutils.c
```diff
@@ -67,18 +70,19 @@ int FileIsX64ELF(const char* filename)
     FILE *f = fopen(filename, "rb");
     if(!f)
         return 0;
-    char head[20] = {0};
-    int sz = fread(head, 20, 1, f);
+    
+    char head[sizeof(x64lib)] = {0};
+    int sz = fread(head, sizeof(x64lib) - 1, 1, f);
```
ptitSeb, line 75: x64lib contains a trailling 0 and should be kept in (it's data, not a C string)

src/tools/fileutils.c
```diff
@@ -106,8 +111,9 @@ int FileIsX64X86ELF(const char* filename)
...
```
ptitSeb, line 116: same

src/tools/fileutils.c
```diff
@@ -88,8 +92,9 @@ int FileIsX86ELF(const char* filename)
     FILE *f = fopen(filename, "rb");
     if(!f)
         return 0;
-    char head[20] = {0};
-    int sz = fread(head, 20, 1, f);
+    
+    char head[sizeof(x86lib)] = {0};
+    int sz = fread(head, sizeof(x86lib) - 1, 1, f);
```
ptitSeb, line 97: -1 is wrong

src/tools/fileutils.c
```diff
@@ -88,8 +92,9 @@ int FileIsX86ELF(const char* filename)
     FILE *f = fopen(filename, "rb");
     if(!f)
         return 0;
-    char head[20] = {0};
-    int sz = fread(head, 20, 1, f);
+    
+    char head[sizeof(x86lib)] = {0};
```
ptitSeb, line 96: either use 64, it's fine

src/tools/fileutils.c
```diff
@@ -106,19 +111,20 @@ int FileIsX64X86ELF(const char* filename)
     FILE *f = fopen(filename, "rb");
     if(!f)
         return 0;
-    char head[64] = {0};
-    size_t sz = fread(head, 1, 20, f);
+    
+    char head[sizeof(x64lib)] = {0};
```
ptitSeb, line 115: 64 was fine before, why changing?

src/tools/fileutils.c
```diff
@@ -129,31 +135,39 @@ int FileIsShell(const char* filename)
...
-    char head[64] = {0};
-    size_t sz = fread(head, 1, sizeof(head) - 1, f);
+
+    #define SHELL_MAX_LEN MAX3(sizeof(bashsign), sizeof(bashsign2), sizeof(shsign))
+    char head[SHELL_MAX_LEN] = {0};
+
+    int sz = fread(head, SHELL_MAX_LEN - 1, 1, f);
     fclose(f);
-    if(sz < strlen(shsign))
+    if(sz != 1)
```
ptitSeb, line 144: this is not the same code. sz has a differnt meaning here than in the original code.

src/tools/fileutils.c
```diff
@@ -129,31 +135,39 @@ int FileIsShell(const char* filename)
...
         return 0;
-    if(!strncmp(head, bashsign2, strlen(bashsign2)))
+
+    if(!strncmp(head, bashsign2, sizeof(bashsign2) - 1))
         return 1;
-    if(!strncmp(head, bashsign, strlen(bashsign)))
+    if(!strncmp(head, bashsign, sizeof(bashsign) - 1))
```
ptitSeb, line 149: what is the difference?

src/tools/fileutils.c
```diff
@@ -129,31 +135,39 @@ int FileIsShell(const char* filename)
...
         return 1;
-    if(!strncmp(head, shsign, strlen(shsign)))
+    if(!strncmp(head, shsign, sizeof(shsign) - 1))
```
ptitSeb, line 151: what is the difference?

src/tools/fileutils.c
```diff
@@ -129,31 +135,39 @@ int FileIsShell(const char* filename)
...
 
 int FileIsPython(const char* filename)
 {
     FILE *f = fopen(filename, "rb");
     if(!f)
         return 0;
-    char head[64] = {0};
-    size_t sz = fread(head, 1, sizeof(head) - 1, f);
+
+    char head[sizeof(pythonsign)] = {0};
```
ptitSeb, line 163: why not 64?

src/tools/fileutils.c
```diff
@@ -129,31 +135,39 @@ int FileIsShell(const char* filename)
...
+
+    int sz = fread(head, sizeof(pythonsign) - 1, 1, f);
```
ptitSeb, line 165: what was wrong with previous code?

libx264:
I must apologize, at least for ai description and unnecessary changes. The issue persist on termux, because of patches that change system prefix, but they forgot to modify head size. I should have edited their patch, but I thought it would be good to implement dynamic head buffer size.

src/tools/fileutils.c
libx264: On accident, will return on next commit

src/tools/fileutils.c
libx264: Compiler adds an extra 0 on string literals (without -1 it will be 21 instead of 20). Can leave as it or change to unsigned char
ex:
static const unsigned char x64lib[] = {
    0x7f, 'E', 'L', 'F', 0x02, 0x01, 0x01, 0x03, 
    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 
    0x02, 0x00, 0x3e, 0x00
};
in that case sizeof is 20

src/tools/fileutils.c
libx264: Compile time instead of runtime, why not?
-1 because of trailing zero

src/tools/fileutils.c
libx264: Why?

src/tools/fileutils.c
libx264: > I thought it would be good to implement dynamic head buffer size

To reduce the number of magic numbers

ptitSeb:
That looks ok for now, but I still don't see any "Fix buffer overflow and stack smashing". Where is the buffer overflow in the original code?

libx264:
The description and title were generated by an AI, so they are incorrect and outdated. Initially, I simply increased the buffer size and saw that it fixed the issue. After further testing, I realized the bug was actually on Termux's side, as they were blindly replacing shebangs at the beginning of the file with patch in pkg repo. However, I still think that dynamically determining the buffer size at compile time is a good idea. I could close this and open a new pull request, but I can also just edit the current title and description instead.

ptitSeb:
Well, I would appreciate some cleanup yes. Title and main description are really heavy and confusing. Either edit or create a new ticket is up to you.

PR #4429, opened 2026-09-19
[LIBRARIAN] Fixed dlopen/dlsym/dlclose races

ksco:


ptitSeb:
What happens if a dlopenned library use dlopen or dlsym itself (during library `init`)?

ksco:
> What happens if a dlopenned library use dlopen or dlsym itself (during library `init`)?

the lib_loader_mutex has type PTHREAD_MUTEX_RECURSIVE.

PR #4431, opened 2026-09-19
[BOX32][X11] Fixed refreshDisplay() on an unregistered Display

ksco:


src/libtools/my_x11_conv.c
```diff
@@ -376,11 +384,9 @@ void refreshDisplay(void* dpy)
     }
     // functions
     bridge_t* system = my_context->libx11->w.bridge;
-    int N = -1;
-    for(int i=0; i<N_DISPLAY && (N==-1); ++i) {
-        if(my32_Displays_64[i]==dpy)
-            N = i;
-    }
+    int N = (int)(dst - my32_Displays_32);
```
ptitSeb, line 387: don't you need to divide by `sizeof(dst)` to get `N`?

ksco: No, it's pointer math in C.

PR #4434, opened 2026-09-20
[RV64_DYNAREC] Fix aliased operands for scalar-version VPACKUSDW

zengdage:
Handle in three cases:

1. When `ed == vex.v` and `gd == ed == vex.v`, the computed result of `ed` matches `vex.v`. Thus overwriting `ed` has no effect, we can get it's high 64-bits data from it's low 64 bits data.

2. When `gd == ed`, retrieve the high 64-bit data by processing `ed` first, to prevent overwriting the low 64-bit source data of `ed` when processing `vex.v` beforehand.

3. For `gd == vex.v` and all other cases, keep the original implementation.


zengdage:
Reported by https://github.com/zengdage/x86_64-inst-test/blob/main/avx256/test_vpackusdw.c with:

BOX64_DYNARC_NOHOSTEXT=vector

src/dynarec/rv64/dynarec_rv64_avx_66_0f38.c
```diff
@@ -934,30 +934,54 @@ uintptr_t dynarec64_AVX_66_0F38(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t i
...
-            for (int i = 0; i < 4; ++i) {
-                LW(x3, vback, vxoffset + i * 4);
-                SATUw(x3, x5);
-                SH(x3, gback, gdoffset + i * 2);
-            }
-            for (int i = 0; i < 4; ++i) {
-                LW(x3, wback, fixedaddress + i * 4);
-                SATUw(x3, x5);
-                SH(x3, gback, gdoffset + 8 + i * 2);
+#define PACKUSDW_FILL_ELEMENT(SRC, SOFF, DST, DOFF, I) \
```
ptitSeb: Why not moving this macro inside `dynarec_rv64_helper.h` instead? Or have a ginirc one where the midle instuction (`SAT` here) is a macro parameter?

ptitSeb (approved):

src/dynarec/rv64/dynarec_rv64_avx_66_0f38.c
zengdage: Move it to rv64_helper.h now, and also apply it for vpackssdw opcode. Please help to review again. Thanks.

Test it by

https://github.com/zengdage/x86_64-inst-test/blob/main/avx256/test_vpackssdw.c https://github.com/zengdage/x86_64-inst-test/blob/main/avx256/test_vpackusdw.c

ptitSeb (approved):

src/dynarec/rv64/dynarec_rv64_avx_66_0f38.c
ptitSeb: That's better now, thanks.

src/dynarec/rv64/dynarec_rv64_avx_66_0f.c
```diff
@@ -1165,20 +1165,14 @@ uintptr_t dynarec64_AVX_66_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip,
                 wback = x5;
                 fixedaddress = 0;
             }
-            for (int i = 0; i < 4; ++i) {
-                LW(x3, vback, vxoffset + i * 4);
-                SATw(x3, x6, x7);
-                SH(x3, gback, gdoffset + i * 2);
-            }
+            for (int i = 0; i < 4; ++i)
+                PACKSDW_FILL_ELEMENT(x3, SATw(x3, x6, x7), vback, vxoffset, i, gback, gdoffset, i);
```
ksco: So the macro saves 2 lines and takes 8 parameters, that must be very hard to use and to understand. I really don't see the necessity of making this a macro.

src/dynarec/rv64/dynarec_rv64_avx_66_0f38.c
```diff
@@ -934,27 +934,44 @@ uintptr_t dynarec64_AVX_66_0F38(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t i
...
-                SATUw(x3, x5);
-                SH(x3, gback, gdoffset + i * 2);
-            }
-            for (int i = 0; i < 4; ++i) {
-                LW(x3, wback, fixedaddress + i * 4);
-                SATUw(x3, x5);
-                SH(x3, gback, gdoffset + 8 + i * 2);
+            if (MODREG && ed == vex.v) {
+                // If gd == ed == vex.v, the high 64 bits and low 64 bits of
+                // the final gd value are identical.
```
ksco: I don't appreciate comments that literally explain the code. Comments are only necessary when the information is not obvious from the code.

ksco (changes requested):

src/dynarec/rv64/dynarec_rv64_avx_66_0f38.c
zengdage: Done. Remove the comments.

src/dynarec/rv64/dynarec_rv64_avx_66_0f.c
zengdage: This macro reduces 3 lines per use: 10 instances in VPACKUSDW save 30 lines, and 4 in VPACKSSDW save 12 lines, for a total of 42 lines removed.

ksco: Yeah, but it makes the code harder to understand. I prefer to keep them "inlined" as-is.

zengdage: @ptitSeb what’s your take? Either is fine with me. I’ll change it based on the final decision you two make.

ptitSeb: Well, macro are here for 2 reason: mask complexity (like the SAT one) or allow future improvements. I don't think this one really covers either point, as it is only 3 lines. But I don't have a strong opinion on this subject. It's only 3 lines, but it can be used in many places I guess (not just with SAT). I leave the choice to @ksco as he is the one spending the most time on this backend.

zengdage: Remove the macro now. Please help to review again. thanks.

PR #4435, opened 2026-09-20
[RV64_DYNAREC] Add MXCSR rounding translation in VCVTPD2PS

zengdage:
VCVTPD2PS lacks round mode translation, add it now.


zengdage:
I am not sure whether I should add a `fastround` check before calling `setround`. I noticed `CVTPD2PS` does not have a `fastround` check, so I omitted it.

ptitSeb (approved):

ptitSeb:
`double` to `flloat` doesn't have a fastround option indeed, because there was x86 specific case in the rounding

src/dynarec/rv64/dynarec_rv64_avx_66_0f.c
```diff
@@ -566,12 +567,12 @@ uintptr_t dynarec64_AVX_66_0F(dynarec_rv64_t* dyn, uintptr_t addr, uintptr_t ip,
                 FCVTSD(d0, d0);
                 if (!BOX64ENV(dynarec_fastnan)) {
                     BNEZ(x4, 4 + 9 * 4);
-                    SRLI(x6, x3, 63);
-                    SLLI(x6, x6, 31);
+                    SRLI(x4, x3, 63);
```
ksco, line 570: Why change this from x6 to x4?

zengdage: Only x1 is left available, but I want to avoid it since it may be consumed by base routines. x4’s live range ends immediately after BNEZ, so I use x4 to replace x6, and use x6 for rounding mode.

Sure, if you have concerns, I can just use x1 directly and skip this change.

ksco: Ohh, I see!

PR #4447, opened 2026-09-22
[DYNAREC] Fix FNSAVE/FRSTOR data direction

zengdage:
fnsave should use D2LD when storing x87 double values into the 80-bit memory area.

frstor should use LD2D when restoring values from the 80-bit memory area.


zengdage:
Reported by https://github.com/zengdage/x86_64-inst-test/blob/main/float/test_fmisc2.c

